PR #4, opened 2017-07-26
draft - contributing guidelines

helgridly:
IMO the most obvious criticism here is that I haven't implemented [SemVer](http://semver.org/), which would necessitate `major.minor.patch-commit` instead of the proposed `major.minor-commit`.

If we want to follow the SemVer path, I see a couple of options:

- Make devs bump `patch` manually in the same way they bump `major` and `minor`. I think this is going to be error-prone.
- Make Travis generate the `patch` based on its internal build number (easy). This would:
    - be monotonically increasing regardless of major or minor version, e.g. two releases might be numbered `1.1.241` and `1.2.242`. This is a break from the [SemVer spec](http://semver.org/#spec-item-8).
    - generate a bunch of identical packages in Artifactory with different patch numbers.
- Do some funky shenanigans where we get the last build from Artifactory to figure out what the next build should be called.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,16 @@
+Before you start:
+- [ ] For each library you've modified here, decide whether it requires a major, minor, or no version bump. (Click [here](/CONTRIBUTING.md) for guidance)
+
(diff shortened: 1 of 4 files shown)


helgridly:
tagging @kshakir, resident SemVer champion

PR #6, opened 2017-07-26
Added FutureWithTimeout from Rawls

helgridly:


util/src/main/scala/org/broadinstitute/dsde/workbench/util/FutureSupport.scala
```diff
@@ -1,7 +1,13 @@
 package org.broadinstitute.dsde.workbench.util
 
+import java.util.concurrent.TimeoutException
+
+import akka.actor.ActorContext
+
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
+import akka.pattern.after
```
rtitle: Organize imports?

rtitle:
Someday we'll replace `toFutureTry: Future[Try[T]]` with `EitherT[IO, Error, A]` :)

rtitle (approved):

PR #8, opened 2017-07-27
Increased coverage

helgridly:
I tried deprecating the old version of `withTimeout` but Scala can't resolve overloaded functions using their implicits. :(

Before you start:

In all cases:
- [x] Verify all tests go green, including CI tests

- [x] Update `README.md` and the `CHANGELOG.md` for any libs you changed with the new dependency string. The git hash is the same short version you see in GitHub (i.e. seven characters). You can commit these changes straight to develop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ jdk:
 - oraclejdk8
 language: scala
 scala:
+- 2.11.8
 - 2.12.2
 sudo: required
 env:
--- a/model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
+++ b/model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
@@ -25,7 +25,7 @@ class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers 
     ErrorReport.message(msgExc) shouldBe "boom"
   }
 
-  "message function" should "return exception class if no message exists" in {
+  it should "return exception class if no message exists" in {
     val noMsgExc = new RuntimeException()
     ErrorReport.message(noMsgExc) shouldBe "RuntimeException"
   }
@@ -35,7 +35,7 @@ class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers 
     ErrorReport.causes(exc) shouldBe empty
   }
 
-  "causes function" should "return an ErrorReport with suppressed exceptions if there are any" in {
+  it should "return an ErrorReport with suppressed exceptions if there are any" in {
     val excSuppressed = new RuntimeException("suppressed")
     val exc = new RuntimeException("boom")
     exc.addSuppressed(excSuppressed)
@@ -43,7 +43,7 @@ class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers 
     ErrorReport.causes(exc) shouldBe Seq(ErrorReport(excSuppressed))
   }
 
-  "causes function" should "return an ErrorReport with suppressed exceptions even if there are other causes" in {
+  it should "return an ErrorReport with suppressed exceptions even if there are other causes" in {
     val excSuppressed = new RuntimeException("suppressed")
     val excCause = new RuntimeException("cause")
     val exc = new RuntimeException("boom", excCause)
@@ -52,7 +52,7 @@ class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers 
     ErrorReport.causes(exc) shouldBe Seq(ErrorReport(excSuppressed))
   }
 
-  "causes function" should "return an ErrorReport with causes if there are no suppressed exceptions" in {
+  it should "return an ErrorReport with causes if there are no suppressed exceptions" in {
     val excCause = new RuntimeException("cause")
     val exc = new RuntimeException("boom", excCause)
 
--- a/model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
+++ b/model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
@@ -7,8 +7,9 @@ import spray.json._
 import ErrorReportJsonSupport._
 
 class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers {
-  "ErrorReport" should "roundtrip json" in {
-    implicit val errorReportSource = ErrorReportSource("test")
+  implicit val errorReportSource = ErrorReportSource("test")
+
+  "json serialization" should "roundtrip" in {
     val internalEr = ErrorReport("internalERMessage")
     val stacktrace = Seq(new StackTraceElement("testClass", "testMethod", "fileName", 42))
     val er = ErrorReport("testMessage", Some(StatusCodes.OK), Seq(internalEr), stacktrace, Some(classOf[RuntimeException]))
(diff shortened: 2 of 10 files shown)


helgridly:
77.778% coverage achieved!!!

.travis.yml
```diff
@@ -2,6 +2,7 @@ jdk:
 - oraclejdk8
 language: scala
 scala:
+- 2.11.8
```
helgridly, line 5: travis wasn't compiling for scala 2.11 until publishing, which meant that scala 2.11-only errors weren't being reported as failures. oops!

scripts/publish.sh
```diff
@@ -2,8 +2,10 @@
 
 set -e
 
+# Note we only do sbt publish here instead of sbt +publish.
```
helgridly: for the casual reader: `sbt +foo` does `foo` for all versions listed in `crossScalaVersions`.

.travis.yml
rtitle: good catch

scripts/publish.sh
rtitle: TIL

project/Settings.scala
```diff
@@ -53,7 +53,7 @@ object Settings {
   val utilSettings = commonSettings ++ List(
     name := "workbench-util",
     libraryDependencies ++= utilDependencies,
-    version := createVersion("0.1")
+    version := createVersion("0.2")
```
rtitle, line 56: Is it practice to bump the version for every change to this repo? Next question: can it be automated?

util/src/main/scala/org/broadinstitute/dsde/workbench/util/FutureSupport.scala
```diff
@@ -28,7 +28,11 @@ trait FutureSupport {
    */
   def assertSuccessfulTries[K, T](tries: Map[K, Try[T]])(implicit executionContext: ExecutionContext): Future[Map[K, T]] = {
     val failures = tries.values.collect{ case Failure(t) => t }
-    if (failures.isEmpty) Future.successful(tries.map { case (k, v) => k -> v.get}) else Future.failed(failures.head)
+    if (failures.isEmpty) {
+      Future.successful(tries.map { case (k, v) => k -> v.get})
```
rtitle: `tries.mapValues(_.get)`
?

project/Settings.scala
helgridly: 1. no, see https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md
2. see 1. travis does jam the commit hash on the end of the thing it publishes to artifactory

util/src/main/scala/org/broadinstitute/dsde/workbench/util/FutureSupport.scala
```diff
@@ -41,8 +45,8 @@ trait FutureSupport {
     * }}}
     */
   implicit class FutureWithTimeout[A](f: Future[A]) {
-    def withTimeout(duration: FiniteDuration, errMsg: String)(implicit context: ActorContext, ec: ExecutionContext): Future[A] =
-      Future.firstCompletedOf(Seq(f, after(duration, context.system.scheduler)(Future.failed(new TimeoutException(errMsg)))))
+    def withTimeout(duration: FiniteDuration, errMsg: String)(implicit scheduler: Scheduler, ec: ExecutionContext): Future[A] =
```
rtitle, line 48: OK, makes sense to make the implicit the specific type we need

rtitle (approved):
LGTM

project/Settings.scala
rtitle: ok 👍

scripts/publish.sh
jmthibault79: Let's add that in the comment.

helgridly: done

util/src/test/scala/org/broadinstitute/dsde/workbench/util/PackageSpec.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.dsde.workbench.util
+
+import org.scalatest.{FlatSpecLike, Matchers}
+import scala.concurrent.duration._
+
+class PackageSpec extends FlatSpecLike with Matchers {
+
+  "addJitter" should "not add more than 10% jitter" in {
```
jmthibault79, line 8: Could you add a check against `addJitter` being a no-op?  maybe `shouldNotBe 0.5`?

model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
```diff
@@ -18,4 +19,43 @@ class ErrorReportSpec extends FlatSpecLike with BeforeAndAfterAll with Matchers
...
+
+  it should "return an ErrorReport with suppressed exceptions if there are any" in {
+    val excSuppressed = new RuntimeException("suppressed")
+    val exc = new RuntimeException("boom")
+    exc.addSuppressed(excSuppressed)
+
+    ErrorReport.causes(exc) shouldBe Seq(ErrorReport(excSuppressed))
+  }
+
+  it should "return an ErrorReport with suppressed exceptions even if there are other causes" in {
```
jmthibault79, line 46: This is counter-intuitive

util/src/test/scala/org/broadinstitute/dsde/workbench/util/PackageSpec.scala
helgridly: apparently this would fail very very occasionally: https://stackoverflow.com/questions/3065554/can-javas-random-function-be-zero

but i could add `addJitter(0.5) shouldNotBe addJitter(0.5)`, which would only fail if the Random implementation doesn't do consecutive repeats.

jmthibault79: ok

model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
jmthibault79: So we're gonna fix it here: https://broadinstitute.atlassian.net/browse/GAWB-2376

util/src/main/scala/org/broadinstitute/dsde/workbench/util/package.scala
```diff
@@ -8,15 +8,13 @@ import scala.concurrent.duration._
...
     baseTime + ((scala.util.Random.nextFloat * maxJitterToAdd.toNanos) nanoseconds)
   }
 
   def addJitter(baseTime: FiniteDuration): FiniteDuration = {
-    if(baseTime < (1 second)) {
-      addJitter(baseTime, 100 milliseconds)
-    } else if (baseTime < (10 seconds)) {
-      addJitter(baseTime, 500 milliseconds)
+    if(baseTime <= (10 seconds)) {
+      addJitter(baseTime, baseTime * 0.1)
```
jmthibault79, line 17: This is what I like, but as I recall some other folks (Doug?) had Opinions.

helgridly: IIRC I wrote this and pulled the implementation out of my ass. This version isn't wildly different; I checked rawls and all the intervals that get jittered with this function are >10s anyway, so no behaviour change there.

model/src/test/scala/org/broadinstitute/dsde/workbench/model/ErrorReportSpec.scala
helgridly: fwiw the commit message for this is in 2015, by ted, and says "chain sub-service exceptions, clean up error reporting"

PR #9, opened 2017-07-28
Moved workbench-metrics from rawls

helgridly:
**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ In this repo:
 ## workbench-utils
 
 Contains utility functions and classes.
+
 Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.2-f87e766"`
 
 [Changelog](util/CHANGELOG.md)
@@ -33,10 +34,14 @@ NOTE: This library uses akka-http's implementation of spray-json and is therefor
 
 - `ErrorReport`
 
-## workbench-google
-
-Coming soon!
-
 ## workbench-metrics
+
+Contains utilities for instrumenting Scala code and reporting to StatsD using [metrics-scala](https://github.com/erikvanoosten/metrics-scala) and [metrics-statsd](https://github.com/ReadyTalk/metrics-statsd).
+
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metrics" % "0.1-xxxxxxx"`
+
+[Changelog](metrics/CHANGELOG.md)
+
+## workbench-google
 
 Coming soon!
--- a/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Expansion.scala
+++ b/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Expansion.scala
@@ -1,0 +1,60 @@
+package org.broadinstitute.dsde.workbench.metrics
+
+import java.util.UUID
+import akka.http.scaladsl.model._
+import scala.annotation.implicitNotFound
+
+/**
+  * Typeclass for something that can be converted into a metric name fragment.
+  * Metric name fragments can be combined via ExpandedMetricBuilder to generate an "expanded" metric name.
+  * By default this just calls toString on the object of type A, but this can be overridden.
+  */
+@implicitNotFound(msg = "Cannot expand instances of type ${A}")
+trait Expansion[A] {
+  def makeName(a: A): String = a.toString
+
+  final def makeNameWithKey(key: String, a: A) =
+    s"$key.${makeName(a)}"
+}
+
+object Expansion {
+
+  // Typeclass instances:
+
+  /**
+    * Implicit expansion for UUID using the default makeName.
+    */
+  implicit object UUIDExpansion extends Expansion[UUID]
+
+  /**
+    * Implicit expansion for HttpMethod.
+    */
+  implicit object HttpMethodExpansion extends Expansion[HttpMethod] {
+    override def makeName(m: HttpMethod): String = m.value.toLowerCase
+  }
+
+  /**
+    * Implicit expansion for Uri.
+    * Statsd doesn't allow slashes in metric names, so we override makeName to override
+    * the default toString based implementation.
+    */
+  implicit object UriExpansion extends Expansion[Uri] {
+    override def makeName(uri: Uri): String = {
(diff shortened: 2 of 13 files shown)


metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Expansion.scala
```diff
@@ -0,0 +1,61 @@
...
+    * Implicit expansion for UUID using the default makeName.
+    */
+  implicit object UUIDExpansion extends Expansion[UUID]
+
+  /**
+    * Implicit expansion for HttpMethod.
+    */
+  implicit object HttpMethodExpansion extends Expansion[HttpMethod] {
+    override def makeName(m: HttpMethod): String =
+      m.value.toLowerCase
```
helgridly: note this used to be `m.toString.toLowerCase` but that results in `httpMethod(get)` in akka-http, which is ugly

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Expansion.scala
```diff
@@ -0,0 +1,60 @@
...
+  /**
+    * Implicit expansion for UUID using the default makeName.
+    */
+  implicit object UUIDExpansion extends Expansion[UUID]
+
+  /**
+    * Implicit expansion for HttpMethod.
+    */
+  implicit object HttpMethodExpansion extends Expansion[HttpMethod] {
+    override def makeName(m: HttpMethod): String = m.value.toLowerCase
```
helgridly, line 33: note this used to be `m.toString.toLowerCase` but that results in `httpMethod(get)` in akka-http, which is ugly

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/InstrumentationDirectives.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import scala.concurrent.duration._
+
+import akka.http.scaladsl.server.Directive0
+import akka.http.scaladsl.server.Directives.{extractRequest, mapResponse}
+
+trait InstrumentationDirectives extends WorkbenchInstrumented {
+  def instrumentRequest: Directive0 = extractRequest flatMap { request =>
+    val timeStamp = System.currentTimeMillis
+    mapResponse { response =>
```
helgridly, line 11: `requestInstance` -> `extractRequest` and `mapHttpResponse` -> `mapResponse` for spray -> akka-http

it seems to work, according to tests.

README.md
```diff
@@ -33,10 +34,14 @@ NOTE: This library uses akka-http's implementation of spray-json and is therefor
 
 - `ErrorReport`
 
-## workbench-google
+## workbench-metrics
 
-Coming soon!
+Contains utilities for instrumenting Scala code and reporting to StatsD using [metrics-scala](https://github.com/erikvanoosten/metrics-scala) and [metrics-statsd](https://github.com/ReadyTalk/metrics-statsd).
```
rtitle, line 39: Actually the code in this module is agnostic to the backend reporter -- our code is just concerned with writing to the in-memory `MetricRegistry` (provided by metrics-scala/dropwizard). Then we happen to use `metrics-scala` to pull off the MetricsRegistry and write to statsd, but we could conceivably use something else besides statsd.

As I'm thinking about this -- the `workbench-metrics` includes the `metrics-statsd` dependency but it's only used in Rawls [here](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/Boot.scala#L280-L287). So _technically_ we could remove the `metrics-statsd` dependency from `workbench-metrics`. But since we're standardizing on statsd, it might be better to leave it in this module for convenience sake.

Sorry for the long-winded way of saying "this is fine".

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Expansion.scala
rtitle: 👍

README.md
helgridly: Funnily enough I was wondering if that block of code in Rawls should be moved _up_ into this library. We should decide one way or another if this library should be "drop in and you get statsd reporting" or "drop in and you get metrics but wire up your own backend", and do whichever the right thing is based on that.

Opinions? I lean toward the former.

rtitle: Good point - I also like the idea of moving 
```
def startStatsDReporter(host: String, port: Int, period: java.time.Duration): Unit = {
    logger.info(s"Starting statsd reporter writing to [$host:$port] with period [${period.getSeconds} seconds]")
    val reporter = StatsDReporter.forRegistry(SharedMetricRegistries.getOrCreate("default"))
      .convertRatesTo(TimeUnit.SECONDS)
      .convertDurationsTo(TimeUnit.MILLISECONDS)
      .build(host, port)
    reporter.start(period.getSeconds, period.getSeconds, TimeUnit.SECONDS)
  }
```
into this library somewhere. It would be a convenience for other modules depending on this library (they wouldn't need to wire up statsd themselves).

rtitle (approved):
All looks good to me -- cool that the spray->akka-http conversion didn't require too many changes besides renaming stuff.

README.md
helgridly: Anything else that needs to be yanked over?

rtitle: I think that's it in terms of existing code.

One other thought, Rawls has this config:
```
metrics {
  prefix = "{{$environment}}.firecloud.rawls"
  includeHostname = false
  reporters {
    # statsd commented out until it's available in the docker container
    statsd {
      host = "statsd"
      port = 8126
      period = 1m
    }
  }
}
```
https://github.com/broadinstitute/firecloud-develop/blob/dev/configs/rawls/rawls.conf.ctmpl#L134-L145

Currently it's all parsed in `Boot.scala`, but maybe we could make a `MetricsConfig` case class and put it in `workbench-metrics`. Maybe even add logic for parsing it from a .conf file, though I'm not sure if we should assume all modules are using typesafe-config.

helgridly: I think it's reasonable to let users fill in `startStatsDReporter` themselves. Similarly providing prefix etc.

jmthibault79 (approved):

PR #11, opened 2017-07-31
Add API key to metrics

rtitle:
Duplicated in https://github.com/broadinstitute/rawls/pull/747

**PR checklist**

In all cases:

### DIFF
--- a/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
+++ b/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
@@ -9,10 +9,10 @@ import com.typesafe.scalalogging.LazyLogging
 import scala.concurrent.duration.Duration
 
 package object metrics extends LazyLogging {
-  def startStatsDReporter(host: String, port: Int, apiKey: String, period: Duration, registryName: String = "default"): Unit = {
+  def startStatsDReporter(host: String, port: Int, period: Duration, registryName: String = "default", apiKey: Option[String] = None): Unit = {
     logger.info(s"Starting statsd reporter writing to [$host:$port] with period [${period.toMillis} ms]")
     val reporter = StatsDReporter.forRegistry(SharedMetricRegistries.getOrCreate(registryName))
-      .prefixedWith(apiKey)
+      .prefixedWith(apiKey.orNull)
       .convertRatesTo(TimeUnit.SECONDS)
       .convertDurationsTo(TimeUnit.MILLISECONDS)
       .build(host, port)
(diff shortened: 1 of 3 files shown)


metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
```diff
@@ -9,12 +9,13 @@ import com.typesafe.scalalogging.LazyLogging
 import scala.concurrent.duration.Duration
 
 package object metrics extends LazyLogging {
-  def startStatsDReporter(host: String, port: Int, period: Duration, registryName: String = "default"): Unit = {
+  def startStatsDReporter(host: String, port: Int, apiKey: String, period: Duration, registryName: String = "default"): Unit = {
```
helgridly: can we keep the old implementation, deprecate it, and point it to the new one with `apiKey=""` ?

rtitle: Yeah

helgridly: i know it doesn't _really_ matter in this case but i want us to get in the habit of "try to deprecate things rather than update them"

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
```diff
@@ -9,12 +9,13 @@ import com.typesafe.scalalogging.LazyLogging
 import scala.concurrent.duration.Duration
 
 package object metrics extends LazyLogging {
-  def startStatsDReporter(host: String, port: Int, period: Duration, registryName: String = "default"): Unit = {
+  def startStatsDReporter(host: String, port: Int, period: Duration, registryName: String = "default", apiKey: Option[String] = None): Unit = {
```
rtitle, line 12: @helgridly I did this instead of making a deprecated method. Thinking about it, `apiKey` should really be optional -- we need it because we're using hosted graphite, but we wouldn't need to pass it if running our own Graphite/statsd.

helgridly: 👍

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
```diff
@@ -9,12 +9,13 @@ import com.typesafe.scalalogging.LazyLogging
...
     logger.info(s"Starting statsd reporter writing to [$host:$port] with period [${period.toMillis} ms]")
     val reporter = StatsDReporter.forRegistry(SharedMetricRegistries.getOrCreate(registryName))
+      .prefixedWith(apiKey.orNull)
```
rtitle, line 15: confirmed the field is `@Nullable` in this Java library

helgridly (approved):

PR #14, opened 2017-08-14
[GAWB-2146] Instrument Google API calls

rtitle:
Pulling over common changes from https://github.com/broadinstitute/rawls/pull/748/.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ In this repo:
 
 Contains utility functions and classes.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.2-f87e766"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.2-xxxxxxx"`
 
 [Changelog](util/CHANGELOG.md)
 
@@ -17,7 +17,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" 
 
 Contains generic, externally-facing model classes used across Workbench.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.1-f87e766"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.1-xxxxxxx"`
 
 [Changelog](model/CHANGELOG.md)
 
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -9,15 +9,15 @@ In this repo:
 
 Contains utility functions and classes.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.2-f87e766"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.2-xxxxxxx"`
```
rtitle, line 12: These were all out of date. Would be cool if there was a way to automate this.

Also: how can I possibly know the latest hash before I merge to dev? :philosoraptor:

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
```diff
@@ -17,6 +17,8 @@ import scala.concurrent.duration._
 
 class GoogleUtilitiesSpec extends TestKit(ActorSystem("MySpec")) with GoogleUtilities with FlatSpecLike with BeforeAndAfterAll with Matchers with ScalaFutures {
```
rtitle: `GoogleUtilitiesSpec` doesn't exist in Rawls? Seems like I could actually add some instrumentation test cases here.

README.md
helgridly: I agree it would be cool. But not worth dev effort :) Similarly not much we can do about the manual process of updating the readme after merging the PR. Which then bumps the hash of develop anyway. Infinite loop!

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
helgridly: Yeah, I haven't really been backporting.

google/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.2
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.2-xxxxxxx"`
+
+### Added
+
+Instrumentation of Google Cloud HTTP requests.
```
helgridly: `InstrumentedRetry` lives in metrics, so you should update that changelog too.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
helgridly: And I agree. You've dropped below the coverage threshold so you're gonna have to write some tests somewhere :)

helgridly (changes requested):
haven't verified that everything's been pulled from rawls correctly but i trust you :) need to update the metrics changelog and make coveralls happy, but otherwise this is a go.

google/CHANGELOG.md
rtitle: Ah right, ok

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
rtitle: Yep, working on adding tests :)

rtitle:
Made requested changes and got coverage % up!

jmthibault79 (approved):
woo

PR #21, opened 2017-09-08
Added fn to ExpandedMetricBuilder to remove a metric from the registry

helgridly:
**PR checklist**

In all cases:

### DIFF
--- a/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
+++ b/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
@@ -1,0 +1,18 @@
+package org.broadinstitute.dsde.workbench.metrics
+
+import nl.grons.metrics.scala.{Gauge => GronsGauge}
+import nl.grons.metrics.scala.{Counter => GronsCounter}
+import nl.grons.metrics.scala.{Timer => GronsTimer}
+import nl.grons.metrics.scala.{Histogram => GronsHisto}
+
+//Wrapper classes around scala-metrics Metrics so they don't immediately forget their own names.
+
+sealed trait Metric[M] {
+  val name: String
+  val metric: M
+}
+
+class Gauge[T](val name: String, val metric: GronsGauge[T]) extends Metric[GronsGauge[T]]
+class Counter(val name: String, val metric: GronsCounter) extends Metric[GronsCounter]
+class Timer(val name: String, val metric: GronsTimer) extends Metric[GronsTimer]
+class Histogram(val name: String, val metric: GronsHisto) extends Metric[GronsHisto]
--- a/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
+++ b/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
@@ -10,6 +10,7 @@ import nl.grons.metrics.scala.{Histogram => GronsHisto}
 sealed trait Metric[M] {
   val name: String
   val metric: M
+  override def toString: String = name
 }
 
 class Gauge[T](val name: String, val metric: GronsGauge[T]) extends Metric[GronsGauge[T]]
--- a/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/WorkbenchInstrumented.scala
+++ b/metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/WorkbenchInstrumented.scala
@@ -41,6 +41,9 @@ trait WorkbenchInstrumented extends DefaultInstrumented {
         (if (m == "") m else m + ".") + implicitly[Expansion[A]].makeNameWithKey(key, a))
     }
 
+    def getFullName(name: String): String =
+      metricBaseName.append(makeName(name)).name
+
     def asCounter(name: String): Counter =
       metrics.counter(makeName(name))
 
@@ -49,7 +52,7 @@ trait WorkbenchInstrumented extends DefaultInstrumented {
 
     def asGaugeIfAbsent[T](name: String)(fn: => T): Gauge[T] = {
       // Get the fully qualified metric name for inspecting the registry.
-      val gaugeName = metricBaseName.append(makeName(name)).name
+      val gaugeName = getFullName(name)
       metricRegistry.getGauges().asScala.get(gaugeName) match {
         case None =>
           // If the gauge does not exist in the registry, create it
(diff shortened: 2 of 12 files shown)


metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/WorkbenchInstrumented.scala
```diff
@@ -41,6 +41,9 @@ trait WorkbenchInstrumented extends DefaultInstrumented {
         (if (m == "") m else m + ".") + implicitly[Expansion[A]].makeNameWithKey(key, a))
     }
 
+    def getFullName(name: String): String =
+      metricBaseName.append(makeName(name)).name
```
helgridly: wasn't sure why `makeName()` is private; added this as public because we need it for testing. thoughts?

metrics/src/test/scala/org/broadinstitute/dsde/workbench/metrics/MetricsSpec.scala
```diff
@@ -156,6 +156,20 @@ class MetricsSpec extends FlatSpec with Matchers with BeforeAndAfter with Eventu
...
   }
 
+  it should "remove metrics from the registry" in {
+    class RemovedInstrumented extends WorkbenchInstrumented {
+      override val workbenchMetricBaseName = "removeme"
+    }
+    val ri = new RemovedInstrumented
+
+    val ctrBuilder = ri.ExpandedMetricBuilder.expand("a", "counter")
+    val counter = ctrBuilder.asCounter("count")
```
helgridly, line 165: TOL: this habit of passing in a string called "count" or "current" to the metrics we build seems... strange. it's also weird to me that the resulting counter doesn't know its own name, so i have to reconstruct it (including the magic "count" suffix) from the builder.

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/WorkbenchInstrumented.scala
rtitle: I like `getFullName`. I made `makeName` private because it sort of depends on state; it doesn't return the full name of the metric, it returns a name fragment which gets appended to the `metricBaseName`.

metrics/src/test/scala/org/broadinstitute/dsde/workbench/metrics/MetricsSpec.scala
rtitle: 1. Originally I made it optional, so you could do: `builder.asCounter()`. I made it non-optional because I thought it would be useful to have a string defining the _name_ of the metric, as opposed to the _dimensions_ of the metric (i.e. the stuff that gets passed to `ExpandedMetricBuilder`). Also Stackdriver was a lot more picky with the naming structure -- with Graphite it doesn't matter so much.

2. Yeah it is weird that once you have an instance of `Counter` or (`Gauge`, `Timer`, whatever) that you can't get back its name. I've been annoyed by that too.

rtitle (approved):
LGTM. Probably needs a changelog entry

metrics/src/test/scala/org/broadinstitute/dsde/workbench/metrics/MetricsSpec.scala
helgridly: The Derpwizard metric classes don't even implement a common interface or trait! I guess I could wrap _them_ too, but adding another layer to the onion seems a little silly...

helgridly: The more I think about this the more I'm tempted to do it. Encapsulating stupid things to make them more usable is what libraries are for, right?

rtitle: :)
Could wrap the DW classes to add name and add an implicit conversion so API users don't need to change.
```
class CounterWrapper(val counter: Counter, val fullName: String)
implicit def wrapperToCounter(wrapper: CounterWrapper): Counter = wrapper.counter
```

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsde.workbench.metrics
+
+import nl.grons.metrics.scala.{Gauge => GronsGauge}
```
rtitle, line 3: These could all be on one line, but maybe more readable this way?

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/Metrics.scala
```diff
@@ -0,0 +1,18 @@
...
+import nl.grons.metrics.scala.{Counter => GronsCounter}
+import nl.grons.metrics.scala.{Timer => GronsTimer}
+import nl.grons.metrics.scala.{Histogram => GronsHisto}
+
+//Wrapper classes around scala-metrics Metrics so they don't immediately forget their own names.
+
+sealed trait Metric[M] {
+  val name: String
```
rtitle, line 11: Maybe as a convenience:
```
override def toString: String = name
```
?

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
```diff
@@ -18,4 +18,6 @@ package object metrics extends LazyLogging {
       .build(host, port)
     reporter.start(period.toMillis, period.toMillis, TimeUnit.MILLISECONDS)
   }
+
+  implicit def metricUnwrapper[M](metric: Metric[M]): M = metric.metric
```
rtitle: Wondering: should this live in `WorkbenchInstrumented`? Then it wouldn't need to be explicitly imported (subclasses automatically get parent traits' implicits).

metrics/src/test/scala/org/broadinstitute/dsde/workbench/metrics/MetricsSpec.scala
```diff
@@ -19,6 +18,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.Try
+import nl.grons.metrics.scala
```
rtitle: Unused?

rtitle:
Nitpicks aside, the approach looks good to me

metrics/src/main/scala/org/broadinstitute/dsde/workbench/metrics/package.scala
helgridly: good idea, thanks

metrics/src/test/scala/org/broadinstitute/dsde/workbench/metrics/MetricsSpec.scala
helgridly: yup

PR #30, opened 2017-10-07
ported HealthMonitor from rawls

dvoet:
port HealthMonitor from rawls for use in sam and elsewhere

**PR checklist**

In all cases:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,7 @@ val testAndCompile = "test->test;compile->compile"
 
 lazy val workbenchUtil = project.in(file("util"))
   .settings(utilSettings:_*)
+  .dependsOn(workbenchModel)
   .withTestSettings
 
 lazy val workbenchModel = project.in(file("model"))
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
@@ -41,7 +41,11 @@ class MockGoogleDirectoryDAO( implicit val executionContext: ExecutionContext ) 
   }
 
   override def getGoogleGroup(groupEmail: WorkbenchGroupEmail): Future[Option[Group]] = {
-    Future.successful(Some(new Group()))
+    Future.successful(groups.get(groupEmail).map { _ =>
+      val googleGroup = new Group()
+      googleGroup.setEmail(groupEmail.value)
+      googleGroup
+    })
   }
 
   override def isGroupMember(groupEmail: WorkbenchGroupEmail, memberEmail: WorkbenchEmail): Future[Boolean] = {
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/health/HealthMonitor.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/health/HealthMonitor.scala
@@ -1,0 +1,159 @@
+package org.broadinstitute.dsde.workbench.health
+
+import java.util.concurrent.TimeoutException
+
+import akka.actor.{Actor, Props}
+import akka.pattern.{after, pipe}
+import cats._
+import cats.implicits._
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.workbench.health.Subsystems.Subsystem
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+
+import HealthMonitor._
+
+/**
+  * Created by rtitle on 5/17/17.
+  */
+object HealthMonitor {
+  val DefaultFutureTimeout = 1 minute
+  val DefaultStaleThreshold = 15 minutes
+
+  val OkStatus = SubsystemStatus(true, None)
+  val UnknownStatus = SubsystemStatus(false, Some(List("Unknown status")))
+  def failedStatus(message: String) = SubsystemStatus(false, Some(List(message)))
+
+  // Actor API:
+
+  sealed trait HealthMonitorMessage
+  /** Triggers subsystem checking */
+  case object CheckAll extends HealthMonitorMessage
+  /** Stores status for a particular subsystem */
+  case class Store(subsystem: Subsystem, status: SubsystemStatus) extends HealthMonitorMessage
+  /** Retrieves current status and sends back to caller */
+  case object GetCurrentStatus extends HealthMonitorMessage
+
+  def props(subsystems: Set[Subsystem], futureTimeout: FiniteDuration = DefaultFutureTimeout, staleThreshold: FiniteDuration = DefaultStaleThreshold)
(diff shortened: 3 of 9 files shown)


util/src/main/scala/org/broadinstitute/dsde/workbench/health/StatusModel.scala
```diff
@@ -0,0 +1,61 @@
...
+  case object Cromwell extends Subsystem
+  case object Database extends Subsystem
+  case object GoogleBilling extends Subsystem
+  case object GoogleBuckets extends Subsystem
+  case object GoogleGenomics extends Subsystem
+  case object GoogleGroups extends Subsystem
+  case object GooglePubSub extends Subsystem
+  case object Sam extends Subsystem
+  case object Thurloe extends Subsystem
+  case object Mongo extends Subsystem
```
dvoet: happy for suggestions of how not to have this list statically defined here

google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
```diff
@@ -41,7 +41,11 @@ class MockGoogleDirectoryDAO( implicit val executionContext: ExecutionContext )
   }
```
helgridly, line 41: I know you didn't put this file here, but mocks should live in `src/test`, not `src/main`

dvoet:
I think it is here so it makes it into the jar and others can use it. A
test package maybe?

helgridly:
We've managed to pull in classes from test packages before. You can do `% test` or something.

helgridly:
We do publish our test packages too: https://github.com/broadinstitute/workbench-libs/blob/develop/project/Publishing.scala#L31

build.sbt
```diff
@@ -5,6 +5,7 @@ val testAndCompile = "test->test;compile->compile"
 
 lazy val workbenchUtil = project.in(file("util"))
   .settings(utilSettings:_*)
+  .dependsOn(workbenchModel)
```
rtitle, line 8: I always kind of pictured that `workbenchUtil` would be the root of the dependency tree -- i.e. containing common workbench utilities free of other workbench dependencies.

If it depends on model then we should make sure that model doesn't depend on util in the future (because we don't want circular dependencies).

Alternatively: should HealthMonitor actually live in `workbenchModel` (because it does carry with it a `StatusModel`)?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
rtitle: +1
We can publish the test jar as an artifact (I think we do that already with util and metrics).

util/src/main/scala/org/broadinstitute/dsde/workbench/health/HealthMonitor.scala
```diff
@@ -0,0 +1,159 @@
...
+  /**
+    * Adds non-blocking timeout support to futures.
+    * Example usage:
+    * {{{
+    *   val future = Future(Thread.sleep(1000*60*60*24*365)) // 1 year
+    *   Await.result(future.withTimeout(5 seconds, "Timed out"), 365 days)
+    *   // returns in 5 seconds
+    * }}}
+    */
+  private implicit class FutureWithTimeout[A](f: Future[A]) {
```
rtitle: I think this is already in `util/FutureSupport.scala` so this can be removed

util/src/main/scala/org/broadinstitute/dsde/workbench/health/HealthMonitor.scala
```diff
@@ -0,0 +1,159 @@
...
+  * @param subsystems the list of all the subsystems this instance should care about
+  * @param futureTimeout amount of time after which subsystem check futures will time out (default 1 minute)
+  * @param staleThreshold amount of time after which statuses are considered "stale". If a status is stale
+  *                       then it won't be returned to the caller; instead a failing status with an "Unknown"
+  *                       message will be returned. This shouldn't normally happen in practice if we have
+  *                       reasonable future timeouts; however it is still a defensive check in case something
+  *                       unexpected goes wrong. Default 15 minutes.
+  */
+class HealthMonitor private (val subsystems: Set[Subsystem], val futureTimeout: FiniteDuration, val staleThreshold: FiniteDuration)
+                            (checkHealth: () => Traversable[(Subsystem, Future[SubsystemStatus])]) extends Actor with LazyLogging {
```
rtitle: Should `checkHealth()` return a `Map[Subsystem, Future[SubsytemStatus]]` since presumably you want the keys to be unique?

util/src/main/scala/org/broadinstitute/dsde/workbench/health/HealthMonitor.scala
```diff
@@ -0,0 +1,159 @@
...
+  private def checkAll: Unit = {
+    checkHealth().foreach(processSubsystemResult)
+  }
+
+  /**
+    * A monoid used for combining SubsystemStatuses.
+    * Zero is an ok status with no messages.
+    * Append uses && on the ok flag, and ++ on the messages.
+    */
+  implicit val SubsystemStatusMonoid = new Monoid[SubsystemStatus] {
```
rtitle: I don't think this is actually used here. Remove it?

It was used in Rawls for when there are multiple items to check which need to be folded up into a single `SubsystemStatus`. Example: https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitor.scala#L194-L203

But since the checking is now delegated to `checkHealth()` then the implementer of `checkHealth` should be responsible for using this monoid.

build.sbt
dvoet: I was thinking of it the other way, model is the root. Also not that utils is free from other workbench dependencies but it is a place for utils shared across workbench.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
dvoet: I don't have time to deal with this now so will JIRA it

build.sbt
rtitle: ok, that's fine

rtitle (approved):

build.sbt
helgridly: My vision was that utils is totally separate and model is the root for all the others.

If something utilitish depends on model it should probably go in model, no?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleDirectoryDAO.scala
dvoet: https://broadinstitute.atlassian.net/browse/GAWB-2705

MatthewBemis (approved):

PR #31, opened 2017-10-11
[GAWB-2646] Sam: implement GET /api/user/petServiceAccount endpoint

rtitle:
Adds models for pet service accounts, and `GoogleIamDAO` for creating service accounts and manipulating IAM roles.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" 
 
 Contains generic, externally-facing model classes used across Workbench.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.2-1b977d7"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.3-???????"`
 
 [Changelog](model/CHANGELOG.md)
 
@@ -35,6 +35,6 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and a DAO for Google PubSub as well as Google Directory. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-1b977d7"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.4-???????"`
 
 [Changelog](google/CHANGELOG.md)
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.4
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-???????"`
+
+### Added
+
+- GoogleIamDAO for creating service accounts and modifying IAM roles
+
 ## 0.3
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-0085f3f"`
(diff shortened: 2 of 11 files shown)


model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -26,15 +26,26 @@ object WorkbenchIdentityJsonSupport {
...
 }
 
 sealed trait WorkbenchSubject extends ValueObject
 sealed trait WorkbenchEmail extends ValueObject
 
 case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchUserEmail)
 case class WorkbenchUserId(value: String) extends WorkbenchSubject
-case class WorkbenchUserEmail(value: String) extends WorkbenchEmail
+case class WorkbenchUserEmail(value: String) extends WorkbenchEmail {
+  def isServiceAccount: Boolean = value.endsWith("gserviceaccount.com")
```
rtitle: This will come in handy for https://broadinstitute.atlassian.net/browse/GAWB-2647

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -0,0 +1,181 @@
...
+   * Google has different model classes for policy manipulation depending on the type of resource.
+   *
+   * For project-level policies we have:
+   *   com.google.api.services.cloudresourcemanager.model.{Policy, Binding}
+   *
+   * For service account-level policies we have:
+   *   com.google.api.services.iam.v1.model.{Policy, Binding}
+   *
+   * These classes are for all intents and purposes identical. To deal with this we create our own
+   * {Policy, Binding} case classes in Scala, with utilties for converting to/from the above Google classes.
```
rtitle: If anyone has a better way to deal with this, let me know. I tried something with Scala structural types, but it was messy.

dvoet (approved):

README.md
```diff
@@ -35,6 +35,6 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and a DAO for Google PubSub as well as Google Directory. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-0085f3f"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-???????"`
```
helgridly: 0.4-???? (for when you do actually update this)

google/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.4
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.3-???????"`
```
helgridly: 0.4

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
helgridly: Suggest making this `endsWith(".gserviceaccount.com")` to stop me registering a parallel TLD

helgridly (commented):
Code to make service accounts in workbench-libs: fine, I guess
Code to make _pet_ service accounts (named as such) in workbench-libs: seems off. Other services aren't going to be doing this, right? Only sam?

rtitle:
@helgridly I agree that `createServiceAccount()` in workbench-libs shouldn't be pet-specific. 
```
def createServiceAccount(googleProject: String, serviceAccountId: WorkbenchUserPetServiceAccountId, displayName: WorkbenchUserPetServiceAccountDisplayName): Future[WorkbenchUserPetServiceAccount]
```
I can look into removing the pet-specific types from the method signature. Pet service accounts are just service accounts with special permissions, but this method should probably be agnostic to that.

MatthewBemis (approved):

PR #33, opened 2017-10-21
[GAWB-2648] Leo call sam to create pet service account at cluster creation time

rtitle:
Supports [Leo](https://github.com/broadinstitute/sam/pull/39) and [Sam](https://github.com/broadinstitute/sam/pull/39) changes.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -30,7 +30,15 @@ trait GoogleIamDAO {
     * @param userEmail the user email address
     * @param rolesToAdd Set of roles to add (example: roles/storage.admin)
     */
-  def addIamRolesForUser(googleProject: String, userEmail: WorkbenchUserEmail, rolesToAdd: Set[String]): Future[Unit]
+  def addIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit]
+
+  /**
+    * Removes project-level IAM roles for the given user.
+    * @param googleProject the google project in which to remove the roles
+    * @param userEmail the user email address
+    * @param rolesToRemove Set of roles to remove (example: roles/dataproc.worker)
+    */
+  def removeIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToRemove: Set[String]): Future[Unit]
 
   /**
     * Adds the Service Account Actor role for the given users on the given service account.
@@ -40,5 +48,5 @@ trait GoogleIamDAO {
     *                               (i.e. the IAM resource).
     * @param userEmail the user email address for which to add Service Account Actor
     */
-  def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchUserEmail): Future[Unit]
+  def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchEmail): Future[Unit]
 }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.workbench.google
 
+import org.broadinstitute.dsde.workbench.google.model.GoogleProject
 import org.broadinstitute.dsde.workbench.model._
 
 import scala.concurrent.Future
@@ -15,30 +16,30 @@ trait GoogleIamDAO {
     * @param displayName the service account display name
     * @return newly created service account
     */
-  def createServiceAccount(googleProject: String, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
+  def createServiceAccount(googleProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
 
   /**
     * Removes a service account in the given project.
     * @param googleProject the project in which to remove the service account
     * @param serviceAccountId the service account id
     */
-  def removeServiceAccount(googleProject: String, serviceAccountId: WorkbenchUserServiceAccountId): Future[Unit]
+  def removeServiceAccount(googleProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId): Future[Unit]
 
   /**
     * Adds project-level IAM roles for the given user.
     * @param googleProject the project in which to add the roles
-    * @param userEmail the user email address
+    * @param email the user email address
     * @param rolesToAdd Set of roles to add (example: roles/storage.admin)
     */
-  def addIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit]
+  def addIamRolesForUser(googleProject: GoogleProject, email: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit]
 
   /**
     * Removes project-level IAM roles for the given user.
     * @param googleProject the google project in which to remove the roles
-    * @param userEmail the user email address
+    * @param email the user email address
     * @param rolesToRemove Set of roles to remove (example: roles/dataproc.worker)
     */
-  def removeIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToRemove: Set[String]): Future[Unit]
+  def removeIamRolesForUser(googleProject: GoogleProject, email: WorkbenchEmail, rolesToRemove: Set[String]): Future[Unit]
 
   /**
     * Adds the Service Account Actor role for the given users on the given service account.
@@ -46,7 +47,7 @@ trait GoogleIamDAO {
     * @param googleProject the project in which to add the roles
     * @param serviceAccountEmail the service account on which to add the Service Account Actor role
     *                               (i.e. the IAM resource).
-    * @param userEmail the user email address for which to add Service Account Actor
+    * @param email the user email address for which to add Service Account Actor
     */
-  def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchEmail): Future[Unit]
+  def addServiceAccountActorRoleForUser(googleProject: GoogleProject, serviceAccountEmail: WorkbenchUserServiceAccountEmail, email: WorkbenchEmail): Future[Unit]
 }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -11,43 +11,43 @@ import scala.concurrent.Future
 trait GoogleIamDAO {
   /**
     * Creates a service account in the given project.
-    * @param googleProject the project in which to create the service account
+    * @param serviceAccountProject the project in which to create the service account
     * @param serviceAccountId the service account id
     * @param displayName the service account display name
     * @return newly created service account
     */
-  def createServiceAccount(googleProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
+  def createServiceAccount(serviceAccountProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
 
   /**
     * Removes a service account in the given project.
-    * @param googleProject the project in which to remove the service account
+    * @param serviceAccountProject the project in which to remove the service account
     * @param serviceAccountId the service account id
     */
-  def removeServiceAccount(googleProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId): Future[Unit]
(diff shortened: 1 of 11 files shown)


model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -33,19 +33,23 @@ object WorkbenchIdentityJsonSupport {
...
 sealed trait WorkbenchSubject extends ValueObject
-sealed trait WorkbenchEmail extends ValueObject
-
-case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchUserEmail)
-case class WorkbenchUserId(value: String) extends WorkbenchSubject
-case class WorkbenchUserEmail(value: String) extends WorkbenchEmail {
+sealed trait WorkbenchEmail extends ValueObject {
   def isServiceAccount: Boolean = value.endsWith(".gserviceaccount.com")
 }
+sealed trait WorkbenchPerson {
```
rtitle: `WorkbenchPerson` is a supertrait of `WorkbenchUser` and `WorkbenchUserServiceAccount`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -30,7 +30,15 @@ trait GoogleIamDAO {
     * @param userEmail the user email address
     * @param rolesToAdd Set of roles to add (example: roles/storage.admin)
     */
-  def addIamRolesForUser(googleProject: String, userEmail: WorkbenchUserEmail, rolesToAdd: Set[String]): Future[Unit]
+  def addIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit]
```
helgridly: We don't have value classes for `GoogleProject`? :o

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -77,20 +83,31 @@ class HttpGoogleIamDAO(clientSecrets: GoogleClientSecrets,
     }
   }
 
-  override def addIamRolesForUser(googleProject: String, userEmail: WorkbenchUserEmail, rolesToAdd: Set[String]): Future[Unit] = {
+  override def addIamRolesForUser(googleProject: String, userEmail: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit] = {
```
helgridly: nit: could do with a comment to indicate that the project is the one in which you want to add the IAM roles

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -77,20 +83,31 @@ class HttpGoogleIamDAO(clientSecrets: GoogleClientSecrets,
...
-      val request = cloudResourceManager.projects().setIamPolicy(s"projects/$googleProject", policyRequest)
+      val request = cloudResourceManager.projects().setIamPolicy(googleProject, policyRequest)
       retryWhen500orGoogleError { () =>
         executeGoogleRequest(request)
       }.void
     }
   }
 
-  override def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchUserEmail): Future[Unit] = {
+  override def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchEmail): Future[Unit] = {
```
helgridly: this could do with a comment too imo

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -114,22 +131,42 @@ class HttpGoogleIamDAO(clientSecrets: GoogleClientSecrets,
   }
 
   /**
-    * Read-modify-write a Policy to insert new bindings for the given member and roles.
+    * Read-modify-write a Policy to insert or remove new bindings for the given member and roles.
```
helgridly, line 142: Could this get racy?

helgridly: Actually, nothing we could do about it here; the lock would need to be global on the IAM end. Let's hope Goog has some good error handling.

helgridly: I take it there's no way to simply add a policy instead of replacing the entire list?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -114,22 +131,42 @@ class HttpGoogleIamDAO(clientSecrets: GoogleClientSecrets,
...
+    // Apply additions
+    val withAdditions = if (rolesToAdd.nonEmpty) {
+      val rolesToAddMap = rolesToAdd.map { _ -> List(email) }.toMap
+      curMembersByRole |+| rolesToAddMap
+    } else {
+      curMembersByRole
+    }
+
+    // Apply deletions
+    // Note that if the same role is in both rolesToAdd and rolesToRemove, the deletion takes precedence.
```
helgridly: Seems like an important point. You should probably roll this up into at least the top-level comment for this function; if there are others that call this with both add and remove, add the note there too

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
helgridly: nit: maybe `WorkbenchIdentity` is a better name, given that service accounts aren't people

(cue my house being firebombed by People for the Ethical Treatment of Service Accounts)

helgridly (approved):
nits

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
rtitle: We do in Leo, not in workbench-libs. It seems generic enough to move to workbench-libs.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
rtitle: The docs suggest doing a read-modify-write:
https://cloud.google.com/iam/docs/granting-changing-revoking-access

> Read-Modify-Write: A common pattern for updating a resource's metadata, such as the Policy, is to read its current state, update the data locally, and then send the modified data for writing. This pattern could cause a conflict if two or more independent processes attempt the sequence simultaneously. For example, if two owners for a project try to make conflicting policy changes at the same time, some changes could fail. Cloud IAM solves this problem using an etag property in Cloud IAM policies. This property is used to verify if the policy has changed since the last request. When you make a request to Cloud IAM with an etag value, Cloud IAM compares the etag value in the request with the existing etag value associated with the policy. It writes the policy only if the etag values match.

> When you update a policy, first get the policy using getIamPolicy(), update the policy, and then write the updated policy using setIamPolicy(). Use the etag value when setting the policy only if the corresponding policy in GetPolicyResponse contains an etag value.

Huh, just reading this I wonder if I should be doing something with etags..

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -33,19 +33,23 @@ object WorkbenchIdentityJsonSupport {
...
+  val id: WorkbenchSubject
```
dvoet: id being a WorkbenchSubject is not right. This would mean that it is legal for a person to have an id that is a group name.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -33,19 +33,23 @@ object WorkbenchIdentityJsonSupport {
 }
 
 sealed trait WorkbenchSubject extends ValueObject
-sealed trait WorkbenchEmail extends ValueObject
-
-case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchUserEmail)
-case class WorkbenchUserId(value: String) extends WorkbenchSubject
-case class WorkbenchUserEmail(value: String) extends WorkbenchEmail {
+sealed trait WorkbenchEmail extends ValueObject {
   def isServiceAccount: Boolean = value.endsWith(".gserviceaccount.com")
```
dvoet: I missed this isServiceAccount function go in but don't really like it here. It feels like it should be closer to where we actually care. It feels dirty having googley stuff here.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -40,5 +48,5 @@ trait GoogleIamDAO {
     *                               (i.e. the IAM resource).
     * @param userEmail the user email address for which to add Service Account Actor
     */
-  def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchUserEmail): Future[Unit]
+  def addServiceAccountActorRoleForUser(googleProject: String, serviceAccountEmail: WorkbenchUserServiceAccountEmail, userEmail: WorkbenchEmail): Future[Unit]
```
dvoet: all of these in here use WorkbenchEmail instead of WorkbenchUserEmail but the parameter name is userEmail. If it must be a user email function signature should be specific.

dvoet (changes requested):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
rtitle: All of these methods (`addIamRolesForUser` / `removeIamRolesForUser` / `addServiceAccountActorRoleForUser`) can act on users or service accounts. In my case I'm calling them on service accounts (e.g. adding Dataproc Worker role to the pet service account). I can change the parameter name from `userEmail` to `email`.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
rtitle: We care in `HttpGoogleIamDAO` (some calls are slightly different depending whether it's a user or service account) and in Sam `StandardUserInfoDirectives` when we process the incoming user email header. I could put this function in workbench-libs/google - it feels less dirty there.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
rtitle: Hm, ok. I'll look at this some more. Maybe we need a trait that represents an id that's _not_ a group name.

rtitle:
@dvoet @helgridly this is ready for re-review, please

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -15,30 +16,38 @@ trait GoogleIamDAO {
     * @param displayName the service account display name
     * @return newly created service account
     */
-  def createServiceAccount(googleProject: String, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
+  def createServiceAccount(googleProject: GoogleProject, serviceAccountId: WorkbenchUserServiceAccountId, displayName: WorkbenchUserServiceAccountDisplayName): Future[WorkbenchUserServiceAccount]
```
helgridly: changing function signatures = you should rev the module version

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -77,20 +84,38 @@ class HttpGoogleIamDAO(clientSecrets: GoogleClientSecrets,
     }
   }
 
-  override def addIamRolesForUser(googleProject: String, userEmail: WorkbenchUserEmail, rolesToAdd: Set[String]): Future[Unit] = {
+  override def addIamRolesForUser(googleProject: GoogleProject, userEmail: WorkbenchEmail, rolesToAdd: Set[String]): Future[Unit] = {
+    // Note the project here is the one in which we're adding the IAM roles
```
helgridly: for all these functions i think it'd be helpful to rename the `googleProject` variable to say _which_ project we mean... might help remove some confusion if they're instead called `serviceAccountProject`, `iamTargetProject`, etc. (that last one is terribly named because i'm not awake yet)

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -37,9 +37,7 @@ sealed trait WorkbenchEmail extends ValueObject
 
 case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchUserEmail)
 case class WorkbenchUserId(value: String) extends WorkbenchSubject
-case class WorkbenchUserEmail(value: String) extends WorkbenchEmail {
-  def isServiceAccount: Boolean = value.endsWith(".gserviceaccount.com")
-}
+case class WorkbenchUserEmail(value: String) extends WorkbenchEmail
```
helgridly, line 40: function go away = bump version

helgridly (approved):
couple of nits but 👍

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
rtitle: yep, will do

PR #42, opened 2017-11-15
List group members

dvoet:
**PR checklist**

In all cases:


MatthewBemis (approved):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
```diff
@@ -11,11 +11,13 @@ import scala.concurrent.Future
 
 trait GoogleDirectoryDAO {
 
+  @deprecated(message = "use createGroup(String, WorkbenchGroupEmail) instead", since = "0.9")
   def createGroup(groupName: WorkbenchGroupName, groupEmail: WorkbenchGroupEmail): Future[Unit]
+  def createGroup(displayName: String, groupEmail: WorkbenchGroupEmail): Future[Unit]
```
breilly2, line 16: Does this mean that we want to move away from wrapper types like `WorkbenchGroupName` in favor of bare `String`s?

breilly2 (approved):
👍  Curious about the change in function signature but otherwise looks good. The recursive code for handling paginated results from Google is a good read!

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
dvoet: no, this particular one does not need a WorkbenchGroupName - it just needs some human readable text. This was fine when we had google groups only for WorkbenchGroups but now we have google groups for policies too which don't have WorkbenchGroupNames.

PR #45, opened 2017-11-27
Reorg service account model

dvoet:
I know, this breaks backwards compatibility. Sorry.

- the various different types of Workbench*Email was too confining and not adding much value. An email is an email is an email and we don't always know which type to create. There is now only 1 WorkbenchEmail case class.
- made a google section in the the model lib because GoogleProject is part of the new PetServiceAccountId.
- moved all WorkbenchUserServiceAccount* model classes to google section of model and removed the WorkbenchUser prefix because they have nothing to do with WorkbenchUsers

**PR checklist**

In all cases:

### DIFF
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.10
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.10-xxxxxxx"`
+
+### Changed
+
+- Moved GoogleProject to model lib
+- Wpdated for 0.8 version of model 
+
 ## 0.9
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.9-dcca21f"`
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -9,7 +9,7 @@ SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.
 ### Changed
 
 - Moved GoogleProject to model lib
-- Wpdated for 0.8 version of model 
+- Updated for 0.8 version of model 
 
 ## 0.9
 
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -75,25 +75,25 @@ trait GoogleIamDAO {
     * Adds the Service Account User role for the given users on the given service account.
     * This allows the users to impersonate as the service account.
     * @param serviceAccountProject the project in which to add the roles
-    * @param WorkbenchEmail the service account on which to add the Service Account User role
+    * @param serviceAccountEmail the service account on which to add the Service Account User role
     *                               (i.e. the IAM resource).
     * @param email the user email address for which to add Service Account User
     */
-  def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject, WorkbenchEmail: WorkbenchEmail, email: WorkbenchEmail): Future[Unit]
+  def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchEmail, email: WorkbenchEmail): Future[Unit]
 
   /**
     * Creates a user-managed key for the given service account.
     * @param serviceAccountProject the google project the service account resides in
-    * @param WorkbenchEmail the service account email
+    * @param serviceAccountEmail the service account email
     * @return instance of ServiceAccountKey
     */
-  def createServiceAccountKey(serviceAccountProject: GoogleProject, WorkbenchEmail: WorkbenchEmail): Future[ServiceAccountKey]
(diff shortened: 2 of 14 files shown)


MatthewBemis (approved):
really like the changes to the emails

helgridly (approved):
i approve of this

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -75,25 +75,25 @@ trait GoogleIamDAO {
     * Adds the Service Account User role for the given users on the given service account.
     * This allows the users to impersonate as the service account.
     * @param serviceAccountProject the project in which to add the roles
-    * @param serviceAccountEmail the service account on which to add the Service Account User role
+    * @param WorkbenchEmail the service account on which to add the Service Account User role
     *                               (i.e. the IAM resource).
     * @param email the user email address for which to add Service Account User
     */
-  def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchUserServiceAccountEmail, email: WorkbenchEmail): Future[Unit]
+  def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject, WorkbenchEmail: WorkbenchEmail, email: WorkbenchEmail): Future[Unit]
```
breilly2: Here and on lines 90 and 98 and all related scaladoc: property name shouldn't be `WorkbenchEmail`. At first, this was going to be a nit that it's not camelCase but there are actually 2 parameters of type `WorkbenchEmail` now. Implementations use `serviceAccountEmail` and `userEmail`. Whatever parameter names you choose, they should be descriptive and consistent here and in the concrete implementations.

breilly2 (changes requested):
Minor change request.

breilly2 (approved):
👍

google/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.10
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.10-xxxxxxx"`
+
+### Changed
+
+- Moved GoogleProject to model lib
+- Wpdated for 0.8 version of model
```
rtitle: typo

rtitle (approved):
Looks good. This will break Leo but that's why we have versioning! This looks much cleaner than the way it was before.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -1,82 +1,31 @@
...
 
-sealed trait WorkbenchSubject {
-  def emailOf(emailString: String): WorkbenchEmail = this match {
-    case _: WorkbenchUserId => WorkbenchUserEmail(emailString)
-    case _: WorkbenchGroupIdentity => WorkbenchGroupEmail(emailString)
-    case _: WorkbenchUserServiceAccountSubjectId => WorkbenchUserServiceAccountEmail(emailString)
-  }
-}
-sealed trait WorkbenchEmail extends ValueObject
+sealed trait WorkbenchSubject
```
jmthibault79, line 19: I'm curious why this is not a `ValueObject` - there's a Swat test that would benefit from that

dvoet: Not all subjects have a single value

breilly2: So really `ValueObject` should be renamed to `SingleValueObject`. That would also help avoid the conflict with what I generally think of as a "value object".

rtitle: In scala “value class” (or object) typically implies a single value: https://docs.scala-lang.org/overviews/core/value-classes.html

breilly2: Ah, okay. Good to know. Thanks! I still wonder if this could have a name that better describes its purpose instead of the mechanism, but at least now I understand what it's referring to.

jmthibault79: I was interpreting this type as "a thing that has a SubjectID" - is that not correct?  That is what I want: a generic accessor for a user's or service account's Subject ID.

dvoet:
Those are not the only kind of subjects

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
breilly2: @dvoet said,

> Those are not the only kind of subjects

Is there an opportunity to clarify this, maybe by choosing a name other than `WorkbenchSubject`? I can understand using "subject" can cause confusion as it did for Joel. Does "subject" refer to something in Google's vocabulary? LDAP's? Something else? I worry that we may be borrowing vocabulary from one context that conflicts with another. We should be clear about which meaning it has in the context of our code.

PR #49, opened 2017-12-07
allow for custom health checks

davidangb:
I had a need to add a new healthcheck for "TrialIndex", which is I think the 6th new subsystem in 10 days? Reviewers: how do you feel about allowing custom health checks like I've implemented below? I haven't really tested this, mostly looking for feedback.

In all cases:

### DIFF
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/util/health/StatusModel.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/util/health/StatusModel.scala
@@ -15,7 +15,13 @@ case class StatusCheckResponse(
 
 object Subsystems {
   sealed trait Subsystem extends ValueObject {
-    override val value: String = getClass.getSimpleName.stripSuffix("$")
+    override val value: String = {
+      if (getClass == Custom.getClass) {
+        this.asInstanceOf[Custom].customName
+      } else {
+        getClass.getSimpleName.stripSuffix("$")
+      }
+    }
   }
 
   def withName(name: String): Subsystem = {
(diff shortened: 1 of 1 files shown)


util/src/main/scala/org/broadinstitute/dsde/workbench/util/health/StatusModel.scala
```diff
@@ -15,7 +15,13 @@ case class StatusCheckResponse(
 
 object Subsystems {
   sealed trait Subsystem extends ValueObject {
-    override val value: String = getClass.getSimpleName.stripSuffix("$")
+    override val value: String = {
+      if (getClass == Custom.getClass) {
```
rtitle, line 19: I'd probably write this as:
```
sealed trait Subsystem extends ValueObject with Product with Serializable {
  override val value: String = this match {
    case Custom(name) => name
    case _ => this.productPrefix
  }
}
```

rtitle: Test:
```
scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed trait Base extends Product with Serializable {
  def value: String = this match {
    case Custom(name) => name
    case _ => this.productPrefix
  }
}
case object Foo extends Base
case object Bar extends Base
case class Custom(name: String) extends Base

// Exiting paste mode, now interpreting.

defined trait Base
defined object Foo
defined object Bar
defined class Custom

scala> Foo.value
res0: String = Foo

scala> Bar.value
res1: String = Bar

scala> Custom("foobar").value
res2: String = foobar
```

rtitle:
Seems ok to me to add custom subsystems (at least I can't think of a downside..). Just a Scala style comment above.

PR #57, opened 2018-01-12
Add HttpGoogleStorageDAO

MatthewBemis:
TODO:

1) ~Implement MockDAO~
2) ~Increment version number~

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -104,4 +104,12 @@ trait GoogleIamDAO {
     * @return list of service account keys
     */
   def listServiceAccountKeys(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchEmail): Future[Seq[ServiceAccountKey]]
+
+  /**
+    * Lists user managed keys associated with a given service account.
+    * @param serviceAccountProject the google project the service account resides in
+    * @param serviceAccountEmail the service account email
+    * @return list of service account keys
+    */
+  def listUserManagedServiceAccountKeys(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchEmail): Future[Seq[ServiceAccountKey]]
 }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
@@ -1,0 +1,21 @@
+package org.broadinstitute.dsde.workbench.google
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+
+import com.google.api.services.storage.model.StorageObject
+
+import scala.concurrent.Future
+
+/**
+  * Created by mbemis on 1/8/18.
+  */
+trait GoogleStorageDAO {
+
+  def createBucket(billingProjectName: String, bucketName: String): Future[String]
+  def storeObject(bucketName: String, objectName: String, objectContents: ByteArrayInputStream, objectType: String = "text/plain"): Future[Unit]
+  def removeObject(bucketName: String, objectName: String): Future[Unit]
+  def getObject(bucketName: String, objectName: String): Future[Option[ByteArrayOutputStream]]
+  def setBucketLifecycle(bucketName: String, lifecycleAge: Int, lifecycleType: String = "Delete"): Future[Unit]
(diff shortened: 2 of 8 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -0,0 +1,21 @@
...
+  * Created by mbemis on 1/8/18.
+  */
+trait GoogleStorageDAO {
+
+  def createBucket(billingProjectName: String, bucketName: String): Future[String]
+  def storeObject(bucketName: String, objectName: String, objectContents: ByteArrayInputStream, objectType: String = "text/plain"): Future[Unit]
+  def removeObject(bucketName: String, objectName: String): Future[Unit]
+  def getObject(bucketName: String, objectName: String): Future[Option[ByteArrayOutputStream]]
+  def setBucketLifecycle(bucketName: String, lifecycleAge: Int, lifecycleType: String = "Delete"): Future[Unit]
+  def listObjectsWithPrefix(bucketName: String, objectNamePrefix: String): Future[Seq[StorageObject]]
```
MatthewBemis, line 19: im not sure i like leaking Google types out, so I'll come up with a better solution for this

google/src/test/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleStorageDAO.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.google.api.services.storage.model.StorageObject
+import org.broadinstitute.dsde.workbench.google.GoogleStorageDAO
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * Created by mbemis on 1/12/18.
+  */
+class MockGoogleStorageDAO(  implicit val executionContext: ExecutionContext ) extends GoogleStorageDAO {
```
MatthewBemis, line 16: note to reviewers: this mock is still a bit rough

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -0,0 +1,120 @@
...
+/**
+  * Created by mbemis on 1/8/18.
+  */
+
+class HttpGoogleStorageDAO(serviceAccountClientId: String,
+                           pemFile: String,
+                           appName: String,
+                           override val workbenchMetricBaseName: String)( implicit val system: ActorSystem, implicit val executionContext: ExecutionContext ) extends GoogleStorageDAO with FutureSupport with GoogleUtilities {
+
+  val storageScopes = Seq(StorageScopes.DEVSTORAGE_FULL_CONTROL, ComputeScopes.COMPUTE, PlusScopes.USERINFO_EMAIL, PlusScopes.USERINFO_PROFILE)
```
dvoet, line 33: do we need compute scope?

google/src/test/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleStorageDAO.scala
dvoet: do we need a mock? can we use mockito?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
MatthewBemis: rawls uses it for storage so that's why i have it here, but i'll test it without compute and see what happens. it's not obvious that anything here should require it

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -101,7 +101,8 @@ trait GoogleIamDAO {
     * Lists keys associated with a given service account.
     * @param serviceAccountProject the google project the service account resides in
     * @param serviceAccountEmail the service account email
+    * @param userManagedOnly set to false to return all keys for the service account, including Google system managed keys
```
MatthewBemis: there are a bunch of extra keys that google returns that they manage and the user has no control over. we can filter them out in the API call.

in order to be backwards compatible, i default this parameter to false

rtitle:
Leo needs this! We currently have a bunch of storage and dataproc stuff mixed together in the same DAO class, and even have a ticket to create a `GoogleStorageDAO`. But I'm happy to use this version.

Glancing at the changes they look good -- we might want to add some additional methods to set bucket ACLs. I might explore with making Leo use `GoogleStorageDAO` in a branch.

rtitle (approved):

PR #59, opened 2018-01-25
GAWB-3018 (part 2): Limit key life using Pub Sub triggers

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
@@ -5,13 +5,12 @@ import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers.{BasicHttpCredentials, OAuth2BearerToken, RawHeader}
+import akka.http.scaladsl.model.headers.OAuth2BearerToken
+import akka.stream.ActorMaterializer
 import com.google.api.client.auth.oauth2.Credential
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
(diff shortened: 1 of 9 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -74,6 +79,43 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
     })
   }
 
+  //This functionality doesn't exist in the com.google.apis Java library.
+  //When we migrate to the com.google.cloud library, we will be able to re-write this to use their implementation
```
MatthewBemis: would also need to deal with migrating to using json credentials instead of pem files. unfortunately this seems like the less messy alternative at the moment.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -74,6 +78,45 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+      Http().singleRequest(request).map { response =>
+        val endTime = System.currentTimeMillis()
+        logger.debug(GoogleRequest(HttpMethods.POST.value, url, Option(entity), endTime - startTime, Option(response.status.intValue), None).toJson(GoogleRequestFormat).compactPrint)
+        ()
+      }
+    }
+  }
+
   override def listObjectsWithPrefix(bucketName: String, objectNamePrefix: String): Future[Seq[StorageObject]] = {
     val getter = getStorage(getBucketServiceAccountCredential).objects().list(bucketName).setPrefix(objectNamePrefix)
```
dvoet: the objects.list api call is paginated, need to make sure we get all the pages or provide a way to get the next page

dvoet (changes requested):
do the paginated thing

dvoet (approved):

PR #60, opened 2018-01-26
preserve and use etags when modifying google iam policies

davidangb:
no jira; found while researching GAWB-2127.

When setting iam policy, we should always be getting the current policy, and round-tripping at least the etag value when we post back to Google. The etag value ensures that the policy has not been modified between the time we read it and the time we (attempted to) write it.

See https://cloud.google.com/logging/docs/audit/configure-data-access#updatemask, or the code comments on com.google.api.services.cloudresourcemanager.model.Policy.getEtag() ... "Removing etag disables the checking for concurrent changes to your policy, and might result in your changes accidentally overwriting someone else's changes."

Furthermore, even though the default behavior of setIamPolicy is to only consider bindings and etag, I've added a `setUpdateMask("bindings,etag")` to ensure those are the only things this code updates. If the default ever changed (and we didn't have the update mask), this code would be overwriting/erasing things that are outside of bindings and etag, such as audit logging config.

**PR checklist**

In all cases:

### DIFF
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -10,6 +10,7 @@ SBT dependency: `TBD`
 
 - Added GoogleStorageDAO
 - Added `listUserManagedServiceAccountKeys` to `GoogleIamDAO`
+- `HttpGoogleIamDAO` now properly preserves etags to protect against concurrent policy changes
 
 ## 0.11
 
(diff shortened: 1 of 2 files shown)


rtitle (approved):
Cool, these changes look good. Thanks.

aednichols (approved):
👍 LGTM.

(I've never reviewed workbench-libs before; what does it take to merit a minor-version bump?)

davidangb:
@aednichols explanation for version bumps is in [CONTRIBUTING.md](https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md). I think this can get away with being a non-versioned PR: no client code changes are necessary.

PR #62, opened 2018-01-28
GoogleStorageDAO enhancements

rtitle:
Updated `GoogleStorageDAO` so it can be used by Leo:

1. Added methods for adding/removing bucket/object/defaultObject ACLs
2. Added `storeObject` variants that take `String` or `File`
3. Replaced `String` with model objects where possible
4. Moved `org.broadinstitute.dsde.workbench.google.gcs` package to workbench-model

**PR checklist**

In all cases:


jmthibault79:
What is the motivation for part 4?

rtitle:
No real technical requirement, I just felt like `GcsPathParser` should be close to the Google model objects in workbench-model. It’s more of a model than a service (i.e. it’s a way of constructing Google models from strings), and we’ve been putting our google models in workbench-model and our google services in workbench-google.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -37,29 +40,67 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+    new Storage.Builder(httpTransport, jsonFactory, bucketServiceAccountCredential).setApplicationName(appName).build()
+  }
+
+  private lazy val bucketServiceAccountCredential: Credential = {
+    new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(serviceAccountClientId)
+      .setServiceAccountScopes(storageScopes.asJava) // grant bucket-creation powers
+      .setServiceAccountPrivateKeyFromPemFile(new java.io.File(pemFile))
```
rtitle, line 56: I'm going to make this DAO accept either a JSON key or a PEM key, since there is some desire for both (and the Google API supports it).

rtitle: (in a separate PR)

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -37,29 +40,84 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+      retryWithRecoverWhen500orGoogleError { () =>
+        executeGoogleRequest(deleteBucketRequest)
+        ()
+      } {
+        case e: HttpResponseException if e.getStatusCode == StatusCodes.NotFound.intValue => ()
+      }
+    }
+  }
+
+  override def bucketExists(bucketName: GcsBucketName): Future[Boolean] = {
```
rtitle, line 94: @vkumra-broad I added this method

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -74,45 +132,96 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+  override def removeObject(bucketName: GcsBucketName, objectName: GcsObjectName): Future[Unit] = {
+    val remover = storage.objects().delete(bucketName.value, objectName.value)
 
-    retryWhen500orGoogleError(() => {
+    retryWithRecoverWhen500orGoogleError { () =>
       executeGoogleRequest(remover)
-    })
+      ()
+    } {
+      case e: HttpResponseException if e.getStatusCode == StatusCodes.NotFound.intValue => ()
```
rtitle, line 220: Trying to use the pattern of recovering from 404s in `delete*` methods

project/Dependencies.scala
```diff
@@ -60,7 +60,8 @@ object Dependencies {
   )
 
   val modelDependencies = commonDependencies ++ Seq(
-    akkaHttpSprayJson
+    akkaHttpSprayJson,
+    googleGuava
```
rtitle, line 64: Needed for `com.google.common.net.UrlEscapers` used by `GcsPathParser`

rtitle:
I think this is ready for review and merge, to unblock other Leo work which depends on it. 

Note this _just_ contains changes to `GoogleStorageDAO` and related classes, no `GoogleCredential` changes (doing that in a separate PR). 

I have tested these methods in sbt console and in Leo (an upcoming Leo PR will use these methods extensively).

google/src/main/scala/org/broadinstitute/dsde/workbench/google/gcs/package.scala
```diff
@@ -1,40 +0,0 @@
-package org.broadinstitute.dsde.workbench.google
-
-import java.util.UUID
-
-/**
-  * Created by rtitle on 9/28/17.
-  */
-package object gcs {
```
rtitle: Moved to workbench-model

jmthibault79:
Comment for part 4 makes sense 👍

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -11,11 +12,28 @@ import scala.concurrent.Future
   */
 trait GoogleStorageDAO {
 
-  def createBucket(billingProjectName: String, bucketName: String): Future[String]
-  def storeObject(bucketName: String, objectName: String, objectContents: ByteArrayInputStream, objectType: String = "text/plain"): Future[Unit]
-  def removeObject(bucketName: String, objectName: String): Future[Unit]
-  def getObject(bucketName: String, objectName: String): Future[Option[ByteArrayOutputStream]]
-  def setBucketLifecycle(bucketName: String, lifecycleAge: Int, lifecycleType: String = "Delete"): Future[Unit]
-  def listObjectsWithPrefix(bucketName: String, objectNamePrefix: String): Future[Seq[StorageObject]]
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName]
```
jmthibault79: yay organization

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -37,29 +40,84 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
       bucketName
     })
   }
 
-  override def storeObject(bucketName: String, objectName: String, objectContents: ByteArrayInputStream, objectType: String = "text/plain"): Future[Unit] = {
-    val storageObject = new StorageObject().setName(objectName)
-    val media = new InputStreamContent(objectType, objectContents)
-    val inserter = getStorage(getBucketServiceAccountCredential).objects().insert(bucketName, storageObject, media)
+  override def deleteBucket(bucketName: GcsBucketName, recurse: Boolean): Future[Unit] = {
+    // If `recurse` is true, first delete all objects in the bucket
```
rtitle, line 71: Leo uses this to delete the init bucket

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -74,45 +132,96 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+    retryWithRecoverWhen500orGoogleError { () =>
       executeGoogleRequest(remover)
-    })
+      ()
+    } {
+      case e: HttpResponseException if e.getStatusCode == StatusCodes.NotFound.intValue => ()
+    }
   }
 
   //"Delete" is the only lifecycle type currently supported, so we'll default to it
```
jmthibault79: Is this worth enforcing on our end, or will the Google error for non-`Delete` be obvious?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -11,11 +12,28 @@ import scala.concurrent.Future
...
+  def deleteBucket(bucketName: GcsBucketName, recurse: Boolean): Future[Unit]
+  def bucketExists(bucketName: GcsBucketName): Future[Boolean]
+
+  def storeObject(bucketName: GcsBucketName, objectName: GcsObjectName, objectContents: ByteArrayInputStream, objectType: String): Future[Unit]
```
jmthibault79, line 17: 👍 for de-stringification.  Could we go one more with a `ContentType` here?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
jmthibault79: Leaving a note here: that's #66

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
rtitle: I'm not sure this comment is accurate. From Google javadoc:

> Type of the action. Currently, only Delete and SetStorageClass are supported.

Of course, we only use Delete. Maybe we could de-stringify this.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -37,29 +40,84 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+          removeObject(bucketName, GcsObjectName(item.getName))
+        }
+      }.void
+    } else Future.successful(())
+
+    deleteObjectsFuture.flatMap { _ =>
+      val deleteBucketRequest = storage.buckets().delete(bucketName.value)
+      retryWithRecoverWhen500orGoogleError { () =>
+        executeGoogleRequest(deleteBucketRequest)
+        ()
```
jmthibault79, line 87: not `.void` ?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -74,45 +132,96 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
...
+
+    retryWithRecoverWhen500orGoogleError { () =>
+      executeGoogleRequest(deleter)
+      ()
+    } {
+      case e: HttpResponseException if e.getStatusCode == StatusCodes.NotFound.intValue => ()
+    }
+  }
+
+  override def setObjectAccessControl(bucketName: GcsBucketName, objetName: GcsObjectName, accessControl: GcsAccessControl): Future[Unit] = {
```
jmthibault79: typo: object

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
rtitle: Well `ContentType` isn't really a Google-specific concept (they just use standard media types like `text/plain`, `application/json`, etc). I sort of want to avoid modeling it when other libraries do a better job (akka-http for example). I feel like this is one place where a `String` is ok.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleStorageDAO.scala
```diff
@@ -14,14 +15,23 @@ import scala.concurrent.{ExecutionContext, Future}
   * Created by mbemis on 1/12/18.
   */
 class MockGoogleStorageDAO(  implicit val executionContext: ExecutionContext ) extends GoogleStorageDAO {
-  val buckets: TrieMap[String, Set[(String, ByteArrayInputStream)]] = TrieMap()
+  val buckets: TrieMap[GcsBucketName, Set[(GcsObjectName, ByteArrayInputStream)]] = TrieMap()
 
-  override def createBucket(billingProjectName: String, bucketName: String): Future[String] = {
-    buckets.putIfAbsent(s"$billingProjectName/$bucketName", Set.empty)
+  override def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName] = {
+    buckets.putIfAbsent(bucketName, Set.empty)
```
jmthibault79: We're losing the project info here.  Is that right?  I think so - buckets are global.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
rtitle: Ha I struggled with `.void` when using `retryWithRecoverWhen500orGoogleError`. (That method IMO is a little wonky)

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
rtitle: thanks

google/src/test/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleStorageDAO.scala
rtitle: Google "create bucket" takes a project (but "delete bucket" does not). The mock doesn't keep track of projects, so this is unused.

jmthibault79: WFM

jmthibault79 (approved):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -132,10 +188,12 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
         }
       }.getOrElse(List.empty)
     }
+
+    // Convert Google StorageObjects to Workbench GcsObjectNames
+    objects.map(_.map(obj => GcsObjectName(obj.getName)))
```
MatthewBemis: this change removes some functionality that sam relies on. for the case where we're checking if pet service account keys are expired, we need to also know the date that the object was created.


for reference, here's the relevant code in sam: https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala#L78

rtitle: Ah, thanks. So maybe we can store the time created in `GcsObjectName` as well.

rtitle: Pushed a commit to track time created in `GcsObjectName`.

MatthewBemis (approved):

PR #66, opened 2018-01-30
Pluggable Google credential mode (pem, json, token) and update all DAOs

rtitle:
Implements pluggable Google credential modes in all Google DAOs. Currently supported:
1. PEM key
2. JSON key
3. Access token

This breaks backwards compatibility so requires a major version bump.

@jmthibault79 FYI

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
@@ -6,7 +6,7 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.client.util.Charsets
-import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail}
 
 import scala.collection.JavaConverters._
 
@@ -50,12 +50,20 @@ object GoogleCredentialModes {
   }
 
   /**
-    * Gets a GoogleCredential from an access token.
-    * Note scopes are not used in this case since we are using a pre-existing token.
+    * Gets a GoogleCredential from an access token. A function is passed in this case so
+    * the token can be refreshed. For example:
+    *
+    * val dao = new HttpFooDAO("my-app", Token(() => obtainAccessToken()), ...)
+    *
+    * Note scopes are not used in this case since we are using pre-existing tokens.
+    *
+    * This implementation does not cache tokens or keep track of expiry. It invokes
+    * the tokenProvider every time a token is needed. It is the caller's responsibility
+    * to handle token expiration and refreshes.
     */
-  case class Token(token: String) extends GoogleCredentialMode {
+  case class Token(tokenProvider: () => String) extends GoogleCredentialMode {
     override def toGoogleCredential(scopes: Seq[String]): GoogleCredential = {
-      new GoogleCredential().setAccessToken(token)
+      new GoogleCredential().setAccessToken(tokenProvider())
     }
   }
 }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
@@ -6,6 +6,7 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.client.util.Charsets
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 
 import scala.collection.JavaConverters._
 
@@ -26,13 +27,13 @@ object GoogleCredentialModes {
   /**
     * Gets a GoogleCredential from a pem file.
     */
-  case class Pem(serviceAccountClientId: String, pemFile: File, serviceAccountUser: Option[String] = None) extends GoogleCredentialMode {
+  case class Pem(serviceAccountClientId: WorkbenchEmail, pemFile: File, serviceAccountUser: Option[WorkbenchEmail] = None) extends GoogleCredentialMode {
     def toGoogleCredential(scopes: Seq[String]) = {
       new GoogleCredential.Builder()
         .setTransport(httpTransport)
         .setJsonFactory(jsonFactory)
-        .setServiceAccountId(serviceAccountClientId)
-        .setServiceAccountUser(serviceAccountUser.orNull)
+        .setServiceAccountId(serviceAccountClientId.value)
+        .setServiceAccountUser(serviceAccountUser.map(_.value).orNull)
         .setServiceAccountScopes(scopes.asJava)
         .setServiceAccountPrivateKeyFromPemFile(pemFile)
         .build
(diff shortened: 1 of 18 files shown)


rtitle:
Gonna look into 2 changes based on comments so far (mostly reminding myself before I go to bed):
1. Make the token mode take an optional callback so it can refresh tokens if they expire. The token mode is really only used by automation tests which _shouldn't_ take more than 1 hour, but still nice to have.
2. Leave the old constructor signatures for backwards compatibility (they should be able to trivially call the new constructor).

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleBigQueryDAO.scala
```diff
@@ -1,38 +1,36 @@
...
-
-  private def bigquery(accessToken: String): Bigquery = {
-    val credential = new GoogleCredential().setAccessToken(accessToken)
-    new Bigquery.Builder(httpTransport, jsonFactory, credential).setApplicationName(appName).build()
+  private lazy val bigquery: Bigquery = {
+    new Bigquery.Builder(httpTransport, jsonFactory, googleCredential).setApplicationName(appName).build()
   }
 
-  override def startQuery(accessToken: String, project: GoogleProject, querySql: String): Future[JobReference] = {
+  override def scopes = Seq(BigqueryScopes.BIGQUERY)
```
jmthibault79: Nit: this is a `val` everywhere else.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
```diff
@@ -0,0 +1,60 @@
...
+    * Represents a way of obtaining a GoogleCredential.
+    */
+  sealed trait GoogleCredentialMode {
+    def toGoogleCredential(scopes: Seq[String]): GoogleCredential
+  }
+
+  /**
+    * Gets a GoogleCredential from a pem file.
+    */
+  case class Pem(serviceAccountClientId: String, pemFile: File, serviceAccountUser: Option[String] = None) extends GoogleCredentialMode {
```
jmthibault79: I would like to see `WorkbenchEmail` used for both the SA ID and User here

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleBigQueryDAO.scala
```diff
@@ -1,38 +1,36 @@
...
-  implicit val service = GoogleInstrumentedService.BigQuery
+  override implicit val service = GoogleInstrumentedService.BigQuery
 
-  lazy val httpTransport = GoogleNetHttpTransport.newTrustedTransport
-  lazy val jsonFactory = JacksonFactory.getDefaultInstance
-
-  private def bigquery(accessToken: String): Bigquery = {
-    val credential = new GoogleCredential().setAccessToken(accessToken)
-    new Bigquery.Builder(httpTransport, jsonFactory, credential).setApplicationName(appName).build()
+  private lazy val bigquery: Bigquery = {
```
jmthibault79: May be out of scope, but can we standardize the pattern we use for Google service creation inside the HTTP DAOs?  No real preference except for including them at the top like this, rather than at the bottom like PubSub.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -160,18 +141,7 @@ class HttpGoogleDirectoryDAO(serviceAccountClientId: String,
   }
 
   private def getGroupDirectory = {
-    new Directory.Builder(httpTransport, jsonFactory, getGroupServiceAccountCredential).setApplicationName(appName).build()
-  }
-
-  private def getGroupServiceAccountCredential: Credential = {
```
jmthibault79: exciting to see this go away

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleCredentialModes.scala
rtitle: 👍  good call

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleBigQueryDAO.scala
rtitle: Yeah agreed - I like them at the top too and `lazy val`s. I can update the DAOs to follow this pattern.

jmthibault79 (approved):
Love it!  with minor changes

rtitle:
@dvoet @helgridly maybe you could take a look at this? Thanks

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -1,32 +1,41 @@
...
-                             override val workbenchMetricBaseName: String,
-                             maxPageSize: Int = 200)( implicit val system: ActorSystem, implicit val executionContext: ExecutionContext ) extends GoogleDirectoryDAO with FutureSupport with GoogleUtilities {
+class HttpGoogleDirectoryDAO(appName: String,
+                             googleCredentialMode: GoogleCredentialMode,
+                             workbenchMetricBaseName: String,
+                             maxPageSize: Int = 200)
+                            (implicit system: ActorSystem, executionContext: ExecutionContext)
+  extends AbstractHttpGoogleDAO(appName, googleCredentialMode, workbenchMetricBaseName) with GoogleDirectoryDAO {
+
+  @deprecated(message = "This way of instantiating HttpGoogleDirectoryDAO has been deprecated. Please upgrade your configs appropriately.", since = "1.0")
```
helgridly: it'll still work though, right? might want to say that to avoid enormous alarm

helgridly: also, is it correct to say "upgrade your configs"? is it configs that need updating, or the code that crates the dao?

project/Settings.scala
```diff
@@ -72,7 +72,7 @@ object Settings {
   val googleSettings = commonSettings ++ List(
     name := "workbench-google",
     libraryDependencies ++= googleDependencies,
-    version := createVersion("0.13"),
+    version := createVersion("1.0"),
```
helgridly: FYI if bumping to this version of wb-libs will cause deprecation warnings in downstream libraries but all code will still compile and work without changes, you can get away with just a minor version bump :)

helgridly (approved):
looks good! i still think a token-refreshing callback might be useful, though. but perhaps that's for another time

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
jmthibault79: Deprecated implies "still works" to me

rtitle: Yeah it should still work (and deprecated implies "still works" to me as well).

I copied this from elsewhere - agreed that "upgrade your configs" are not the right words. Will fix.

project/Settings.scala
rtitle: Yeah was thinking that too. 👍

rtitle:
Thanks @helgridly ! A token-refreshing callback seemed a little tricky to me - the DAOs would need to catch all 401 (I think?) errors and invoke the token-refreshing callback. It might be a nice feature but should maybe be done separately.

helgridly:
If you accept something that looks more like a `UserInfo` -- i.e. something that defines its own expiry time -- then you can use that to decide when to refresh.

rtitle:
Ah true - that would be simpler.

PR #76, opened 2018-02-05
no jira: Ah fix register cleanup

ahaessly:
the register cleanup code was not actually invoking the cleanup functions

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
@@ -100,7 +100,7 @@ trait CleanUp extends TestSuiteMixin with ExceptionHandling with LazyLogging { s
 
   private def runCleanUpFunctions() = {
     println("number of cleanup functions: " + cleanUpFunctions.size())
-    cleanUpFunctions.asScala.foreach { f => try f catch nonFatalAndLog("Error in clean-up function")}
+    cleanUpFunctions.asScala.foreach { f => try f() catch nonFatalAndLog("Error in clean-up function")}
     cleanUpFunctions.clear()
   }
 }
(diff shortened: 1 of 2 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
```diff
@@ -4,12 +4,15 @@ import com.typesafe.scalalogging.LazyLogging
 import java.util.concurrent.ConcurrentLinkedDeque
 import org.broadinstitute.dsde.workbench.service.util.ExceptionHandling
 import org.scalatest.{Outcome, TestSuite, TestSuiteMixin}
+//import scala.collection.concurrent.
```
rtitle: Remove commented line

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
```diff
@@ -95,7 +97,7 @@ trait CleanUp extends TestSuiteMixin with ExceptionHandling with LazyLogging { s
   }
 
   private def runCleanUpFunctions() = {
-    cleanUpFunctions.toArray().foreach { f  => try f catch nonFatalAndLog("Error in clean-up function")}
+    cleanUpFunctions.asScala.foreach { f => try f() catch nonFatalAndLog("Error in clean-up function")}
```
rtitle, line 99: 💯 

Note that previously, `try f` was returning the `f` object (`f` is of type `() => Any`) but not actually running it! Now it actually runs the function.

rtitle (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
breilly2: Wow, I know that cleanup worked originally so the code must have changed at some point. Thanks for fixing!

breilly2 (approved):
👍

PR #77, opened 2018-02-05
Add GoogleStorageDAO.copyObject()

jmthibault79:
For a test for https://github.com/DataBiosphere/leonardo/issues/127

In all cases:

- [x] Update `README.md` and the `CHANGELOG.md` for any libs you changed with the new dependency string. The git hash is the same short version you see in GitHub (i.e. seven characters). You can commit these changes straight to develop.

### DIFF
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -15,7 +15,9 @@ SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.
       - via a json file (`GoogleCredentialModes.Json`)
       - via an access token (`GoogleCredentialModes.Token`)
    - This allows for more flexibility in how Google DAOs can be used.
-
+- Updated methods in GoogleStorageDAO:
+   - Added `copyObject`
+ 
 ## 0.14
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.14-6800f3a"`
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -4,7 +4,7 @@ This file documents changes to the `workbench-google` library, including notes o
 
 ## 0.15
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.15-304e9d0"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.15-TBD"`
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


google/CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.
    - Added methods to create/remove bucket/object ACLs
    - Added `bucketExists(GcsBucketName): Future[Boolean]`
    - Added `storeObject` variants which take an `ByteArrayInputStream` or a `File`
-   
+   - Added `copyObject()`
```
jmthibault79: Just an add - no version bump needed.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -25,6 +25,7 @@ trait GoogleStorageDAO {
   def setBucketLifecycle(bucketName: GcsBucketName, lifecycleAge: Int, lifecycleType: GcsLifecycleType = Delete): Future[Unit]
   def setObjectChangePubSubTrigger(bucketName: GcsBucketName, topicName: String, eventTypes: List[String]): Future[Unit]
   def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String): Future[List[GcsObjectName]]
+  def copyObject(srcBucketName: GcsBucketName, srcObjectName: GcsObjectName, destBucketName: GcsBucketName, destObjectName: GcsObjectName): Future[Unit]
```
rtitle, line 29: Could this API be:
```
(src: GcsPath, dest: GcsPath)
```
since `GcsPath` is defined as:
```
case class GcsPath(bucketName: GcsBucketName, objectName: GcsObjectName)
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -231,6 +231,13 @@ class HttpGoogleStorageDAO(serviceAccountClientId: String,
     })
   }
 
+  override def copyObject(srcBucketName: GcsBucketName, srcObjectName: GcsObjectName, destBucketName: GcsBucketName, destObjectName: GcsObjectName): Future[Unit] = {
+    val copier = storage.objects().copy(srcBucketName.value, srcObjectName.value, destBucketName.value, destObjectName.value, new StorageObject())
```
rtitle, line 239: What's that `new StorageObject()` parameter for?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
jmthibault79: sure

jmthibault79: Actually no - keeping it for consistency w/ others

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
jmthibault79: The purpose of the parameter is to optionally override object resource metadata.  We're not doing that, so it's empty.  it's possible `null` will also work here.  I haven't tried it.

rtitle: OK, I see. I was wondering if `new StorageObject()` would cause it to override resource metadata with default values or something, but it seems like it wouldn't. `null` seems like it would also work. So either way is fine with me.

rtitle (approved):

aweng98 (approved):
thumb up!

PR #78, opened 2018-02-07
add support for GPAlloc

helgridly:
Documentation on what's going on is in this PR: https://github.com/broadinstitute/gpalloc/pull/36/

**PR checklist**

In all cases:

### DIFF
--- a/project/Publishing.scala
+++ b/project/Publishing.scala
@@ -28,6 +28,7 @@ object Publishing {
   //priority over the local package cache. see here: https://github.com/sbt/sbt/issues/2687#issuecomment-236586241
     Seq(
       publishTo := Option(artifactoryResolver(false)),
+      publishArtifact in Compile := true,
       publishArtifact in Test := true,
       credentials += artifactoryCredentials
     )
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -2,15 +2,16 @@ package org.broadinstitute.dsde.workbench.fixture
 
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.{Config, UserPool}
+import org.broadinstitute.dsde.workbench.service.{GPAlloc, Orchestration, Rawls}
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole.BillingProjectRole
 import org.broadinstitute.dsde.workbench.service.test.CleanUp
-import org.broadinstitute.dsde.workbench.service.{Orchestration, Rawls}
 import org.scalatest.TestSuite
 
 import scala.util.Random
 
-trait BillingFixtures extends CleanUp { self: TestSuite =>
+trait BillingFixtures extends CleanUp {
+  self: TestSuite =>
 
   // copied from WebBrowserSpec so we don't have to self-type it
   // TODO make it common to API and browser tests
@@ -18,21 +19,35 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
     Random.alphanumeric.take(length).mkString.toLowerCase
   }
 
+  protected def addMembersToBillingProject(projectName: String, memberEmails: List[String]): Unit = {
+    memberEmails foreach { email =>
+      Orchestration.billing.addUserToBillingProject(projectName, email, BillingProjectRole.Owner)
+    }
+  }
+
+  protected def removeMembersFromBillingProject(projectName: String, memberEmails: List[String]): Unit = {
+    memberEmails foreach { email =>
+      try {
+        Orchestration.billing.removeUserFromBillingProject(projectName, email, BillingProjectRole.Owner)
+      } catch nonFatalAndLog(s"Error removing user $email from billing project $projectName in removeMembersFromBillingProject clean-up")
+    }
+  }
+
+  @deprecated("withBillingProject is deprecated. Use withCleanBillingProject if you want a billing project to isolate your tests, or withBrandNewBillingProject if you want to create a brand new one")
   def withBillingProject(namePrefix: String, memberEmails: List[String] = List())
                         (testCode: (String) => Any)(implicit token: AuthToken): Unit = {
+    withBrandNewBillingProject(namePrefix, memberEmails)(testCode)(token)
+  }
+
+  def withBrandNewBillingProject(namePrefix: String, memberEmails: List[String] = List())
+                                (testCode: (String) => Any)(implicit token: AuthToken): Unit = {
     val billingProjectName = namePrefix + "-" + makeRandomId()
     Orchestration.billing.createBillingProject(billingProjectName, Config.Projects.billingAccountId)
-    memberEmails foreach { email =>
-      Orchestration.billing.addUserToBillingProject(billingProjectName, email, BillingProjectRole.Owner)
-    }
+    addMembersToBillingProject(billingProjectName, memberEmails)
     try {
       testCode(billingProjectName)
     } finally {
-      memberEmails foreach { email =>
-        try {
-          Orchestration.billing.removeUserFromBillingProject(billingProjectName, email, BillingProjectRole.Owner)
-        } catch nonFatalAndLog(s"Error removing user $email from billing project $billingProjectName in withBillingProject clean-up")
-      }
+      removeMembersFromBillingProject(billingProjectName, memberEmails)
       try {
         Rawls.admin.deleteBillingProject(billingProjectName)(UserPool.chooseAdmin.makeAuthToken())
       } catch nonFatalAndLog(s"Error deleting billing project in withBillingProject clean-up: $billingProjectName")
@@ -41,9 +56,8 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
 
   def addUserInBillingProject(billingProjectName: String, email: String, role: BillingProjectRole)
                              (implicit token: AuthToken): Unit = {
-      Orchestration.billing.addUserToBillingProject(billingProjectName, email, role)
-      register cleanUp Orchestration.billing.removeUserFromBillingProject(billingProjectName, email, role)
+    Orchestration.billing.addUserToBillingProject(billingProjectName, email, role)
+    register cleanUp Orchestration.billing.removeUserFromBillingProject(billingProjectName, email, role)
   }
-
 }
 
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -19,13 +19,13 @@ trait BillingFixtures extends CleanUp {
     Random.alphanumeric.take(length).mkString.toLowerCase
   }
 
-  protected def addMembersToBillingProject(projectName: String, memberEmails: List[String]): Unit = {
+  protected def addMembersToBillingProject(projectName: String, memberEmails: List[String])(implicit token: AuthToken): Unit = {
     memberEmails foreach { email =>
       Orchestration.billing.addUserToBillingProject(projectName, email, BillingProjectRole.Owner)
     }
   }
 
-  protected def removeMembersFromBillingProject(projectName: String, memberEmails: List[String]): Unit = {
+  protected def removeMembersFromBillingProject(projectName: String, memberEmails: List[String])(implicit token: AuthToken): Unit = {
     memberEmails foreach { email =>
       try {
         Orchestration.billing.removeUserFromBillingProject(projectName, email, BillingProjectRole.Owner)
@@ -33,7 +33,8 @@ trait BillingFixtures extends CleanUp {
     }
   }
 
-  @deprecated("withBillingProject is deprecated. Use withCleanBillingProject if you want a billing project to isolate your tests, or withBrandNewBillingProject if you want to create a brand new one")
+  @deprecated(message = "withBillingProject is deprecated. Use withCleanBillingProject if you want a billing project to isolate your tests, or withBrandNewBillingProject if you want to create a brand new one",
+              since="workbench-service-test-0.4")
   def withBillingProject(namePrefix: String, memberEmails: List[String] = List())
                         (testCode: (String) => Any)(implicit token: AuthToken): Unit = {
     withBrandNewBillingProject(namePrefix, memberEmails)(testCode)(token)
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
@@ -1,0 +1,146 @@
+package org.broadinstitute.dsde.workbench.fixture
+
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{ConcurrentHashMap, CopyOnWriteArrayList}
+
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchException}
+import org.broadinstitute.dsde.workbench.service.{GPAlloc, GPAllocProject, Rawls}
+import org.scalatest.{BeforeAndAfterAll, Suite, TestSuite}
+
+import scala.collection.JavaConverters._
+
+
+object GPAllocFixtures {
+  //potentially, if people are using ParallelTestExecution, suites may run in parallel.
+  //so these are synchronized java collections so we don't run into trouble.
+  private var gpAllocedProjects: ConcurrentHashMap[String, AuthToken] = new ConcurrentHashMap[String, AuthToken]()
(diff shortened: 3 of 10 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -2,12 +2,75 @@ package org.broadinstitute.dsde.workbench.fixture
...
+import org.broadinstitute.dsde.workbench.service.{GPAlloc, Orchestration, Rawls}
 import org.broadinstitute.dsde.workbench.service.test.CleanUp
 import org.scalatest.TestSuite
 
+import scala.collection.mutable
 import scala.util.Random
 
+object BillingFixtures {
+  private var gpAllocedProjects: mutable.Seq[String] = mutable.Seq.empty[String]
+  private var suiteStack: mutable.Seq[String] = mutable.Seq.empty[String]
```
kyuksel: I fear that simultaneous calls to methods that update these `var`s may result in unexpected behaviour that may be difficult to troubleshoot.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -2,12 +2,75 @@ package org.broadinstitute.dsde.workbench.fixture
...
+  }
+
+  def requestGPAllocedProject(testSuite: TestSuite): Option[(String, String)] = {
+    if( ! suiteStack.contains(testSuite.suiteName) ) {
+      throw new WorkbenchException(
+        "GPAlloc handler doesn't know who you are. Mix in BeforeAndAfterAll to your Suite and call BillingFixtures.registerGPAllocSuite in beforeAll()")
+    }
+
+    val opt = GPAlloc.projects.requestProject
+    opt foreach { gpAllocedProjects += _._1 }
```
kyuksel: Consider making this a bit more self-documenting. One way is to create something like

```scala
case class GPAllocProject(name: String, bucket: String)
```

and use it in the output types of `GPAlloc.projects.{request,release}Project()` and in

```scala
def requestGPAllocedProject(testSuite: TestSuite): Option[GPAllocProject] = {
    if( ! suiteStack.contains(testSuite.suiteName) ) {
      throw new WorkbenchException(
        "GPAlloc handler doesn't know who you are. Mix in BeforeAndAfterAll to your Suite and call BillingFixtures.registerGPAllocSuite in beforeAll()")
    }

    val project = GPAlloc.projects.requestProject
    project foreach { gpAllocedProjects += _.name }
    project
  }
```

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -2,12 +2,75 @@ package org.broadinstitute.dsde.workbench.fixture
...
+    their suite with this handler before requesting a GPAlloc project. This allows us to keep a "stack" of suites
+    that may individually attempt to call releaseAllGPAllocedProjects; we ignore such calls unless they come from the
+    bottom-most suite in the stack. If the tests are configured correctly, this bottom-most suite will either be the
+    outermost super-Suite in which the others are nested, or then we're running a single suite on its own and all is well.
+
+    Once releaseAllGPAllocedProjects is called successfully, we flip a switch that tells us to barf if someone attempts
+    to GPAlloc further projects; that would indicate that tests are set up wrong and are thus at risk of upsetting Rawls+Sam
+    with duplicate projects.
+     */
+  private var bNeverCleanUpAgain = false
```
kyuksel: `b` stands for boolean? It's a bit confusing to me.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -2,12 +2,75 @@ package org.broadinstitute.dsde.workbench.fixture
...
+
+    Rawls+Sam currently have no way to delete projects, so releasing the project back to GPAlloc risks getting
+    the same project back in a later call to requestGPAllocedProject. That would make Rawls+Sam upset at having
+    to set up a project they already know about.
+
+    Without the ability to delete projects, we only know for sure that the project is unused when the test run
+    is complete (at which point the Rawls + Sam databases will be wiped entirely). This means we have
+    to register a hook that runs after ALL the tests are complete, and only ONCE. Since ScalaTest provides no
+    "before and after the entire test run" mechanism, we are forced to mandate that all tests are nested inside
+    a single outermost ScalaTest Suite that mixes in BeforeAndAfterAll and calls releaseAllGPAllocedProjects in
```
kyuksel: What happens if we have a crash before `releaseAllGPAllocedProjects` is called or it completes?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -16,8 +79,14 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
...
   }
 
+  @deprecated("withBillingProject is deprecated. Use withCleanBillingProject if you want a billing project to isolate your tests, or withBrandNewBillingProject if you want to create a brand new one")
   def withBillingProject(namePrefix: String)
                         (testCode: (String) => Any)(implicit token: AuthToken): Unit = {
+    withBrandNewBillingProject(namePrefix)(testCode)(token)
+  }
+
+  def withBrandNewBillingProject(namePrefix: String)
+                        (testCode: (String) => Any)(implicit token: AuthToken): Unit = {
```
kyuksel: align the constructor argument groups for readability

kyuksel (changes requested):
Looks good in general. A few comments inline.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -2,12 +2,75 @@ package org.broadinstitute.dsde.workbench.fixture
...
+  /* A crude way of enforcing that tests are set up correctly.
+
+    Rawls+Sam currently have no way to delete projects, so releasing the project back to GPAlloc risks getting
+    the same project back in a later call to requestGPAllocedProject. That would make Rawls+Sam upset at having
+    to set up a project they already know about.
+
+    Without the ability to delete projects, we only know for sure that the project is unused when the test run
+    is complete (at which point the Rawls + Sam databases will be wiped entirely). This means we have
+    to register a hook that runs after ALL the tests are complete, and only ONCE. Since ScalaTest provides no
+    "before and after the entire test run" mechanism, we are forced to mandate that all tests are nested inside
```
aednichols: [`ParallelTestExecution`](http://doc.scalatest.org/3.0.1/index.html#org.scalatest.ParallelTestExecution):

> ScalaTest's normal approach for running suites of tests in parallel is to run different suites in parallel, but the tests of any one suite sequentially.

Given the above, I am worried about the `var`s on the `BillingFixtures` singleton causing concurrency issues.

Just throwing it out there, but we have been playing with the trait below in a different context.

[`SequentialNestedSuiteExecution`](http://doc.scalatest.org/3.0.1/index.html#org.scalatest.SequentialNestedSuiteExecution):

> Mix in this trait into any suite whose nested suites need to be run sequentially even with the rest of the run is being executed concurrently.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
helgridly: this: https://github.com/broadinstitute/gpalloc/issues/8

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
helgridly: Hmm, OK. I can make these vars synchronized / more parallel / something.

What are your thoughts around `SequentialNested`? Doing it at the top-level super-Suite is a good way to make tests take forever, I'd imagine.

aednichols: `ParallelTestExecution` is probably more appropriate, shouldn't slow things down (may even speed them up)

> To make it easier for users to write tests that run in parallel, this trait runs each test in its own instance of the class. Running each test in its own instance enables tests to use the same instance vars and mutable objects referenced from instance variables without needing to synchronize.

kyuksel: > I can make these vars synchronized / more parallel / something.

I have a mini library for lock synchronization from a previous project. If that'd help, let me know.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
aednichols: Pretty common pattern in dynamically-typed languages, used to see it a lot in PHP

kyuksel:
What's the testing strategy around this code? Should we create a UI test that goes through the nesting scenarios and also benchmark performance if necessary?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
kyuksel: I see

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
helgridly: Standard synchronization primitives should work fine here. Will change.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
helgridly: This is a hangover from my previous life as a C++ developer, where we did hungarian notation for _everything_. I've mostly kicked the habit but b-for-bool sticks with me.

helgridly:
@kyuksel @aednichols re-review plx

helgridly:
re testing strategy -- once we get an allocator instance stood up we'll try it out, probably manually, before merging this PR

kyuksel (approved):
👍ing pending tests

aednichols:
Can we use this to pre-allocate projects for our actual users, not just Selenium?

Should we do that first?

helgridly:
Facetalked with Adam, the question was "can we do this in prod" and the answer is "no, it's more complicated in prod / in tests we rely on the fiab db being taken down to clean up the projects / we don't have code to switcharoo the billing account". We could but it's not an immediate prio and certainly not for this service.

project/Publishing.scala
```diff
@@ -28,6 +28,7 @@ object Publishing {
   //priority over the local package cache. see here: https://github.com/sbt/sbt/issues/2687#issuecomment-236586241
     Seq(
       publishTo := Option(artifactoryResolver(false)),
+      publishArtifact in Compile := true,
```
helgridly, line 31: I hoped this would fix the no-downloaded-sources issue, but it didn't.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
breilly2: I was going to quote the same bit of `ParallelTestExecution` that @aednichols did. My reading of that is that we don't need to do any synchronization ourselves. Do others disagree?

helgridly: These vars are shared by all suites (by design) because they're in a companion object. If they might need synchronization then might as well leave this here.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -12,6 +12,11 @@ trait Rawls extends RestClient with LazyLogging {
       logger.info(s"Deleting billing project: $projectName")
       deleteRequest(url + s"api/admin/billing/$projectName")
     }
+
+    def claimProject(projectName: String, cromwellAuthBucket: String, newOwner: String)(implicit token: AuthToken): Unit = {
```
jmthibault79, line 16: Add a short comment about why releaseProject is elsewhere

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
```diff
@@ -50,21 +50,23 @@ trait RestClient extends Retry with LazyLogging {
     Await.result(responseFuture, 5.minutes)
   }
 
+  def extractResponseString(response: HttpResponse): String = {
```
jmthibault79, line 53: 👍 this stupid thing is all over FireCloudClient too btw

jmthibault79 (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
```diff
@@ -0,0 +1,127 @@
...
+    their suite with this handler before requesting a GPAlloc project. This allows us to keep a "stack" of suites
+    that may individually attempt to call releaseAllGPAllocedProjects; we ignore such calls unless they come from the
+    bottom-most suite in the stack. If the tests are configured correctly, this bottom-most suite will either be the
+    outermost super-Suite in which the others are nested, or then we're running a single suite on its own and all is well.
+
+    Once releaseAllGPAllocedProjects is called successfully, we flip a switch that tells us to barf if someone attempts
+    to GPAlloc further projects; that would indicate that tests are set up wrong and are thus at risk of upsetting Rawls+Sam
+    with duplicate projects.
+     */
+  private var bNeverCleanUpAgain = false
```
rtitle: Not sure if this boolean is going to be accessed in parallel -- probably not since it's only used in `beforeAll`/`afterAll`? Still it might be good to use `AtomicBoolean` to protect against concurrency issues.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
```diff
@@ -0,0 +1,127 @@
...
+  def requestGPAllocedProject(testSuite: Suite)(implicit authToken: AuthToken): Option[GPAllocProject] = {
+    if( ! suiteStack.contains(testSuite.suiteName) ) {
+      //The most likely developer error that would trigger this codepath: mixing in GPAllocFixtures to your suite,
+      //then overriding beforeAll without calling super(). This stops us being able to clear up claimed projects.
+      throw new WorkbenchException(
+        "GPAlloc handler doesn't know who you are. Have you forgotten to call super() in your BeforeAndAfterAll overrides?")
+    }
+
+    val opt = GPAlloc.projects.requestProject
+    opt foreach { p => gpAllocedProjects.put(p.projectName, authToken) }
```
rtitle, line 73: So we're storing the auth token in the `ConcurrentHashMap` - what if the token expires before the tests finish? Does releasing the projects fail or is there something to handle that?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
rtitle: Yikes. A better way might be:
```
    val responseStringFuture: Future[String] = response.entity.toStrict(5 minutes).map(_.data.utf8String)
    Await.result(responseStringFuture, 5 minutes)
```
Also not a fan of sprinkling `Await.result`s in this code but I'll choose my battles.

rtitle (approved):
Looks good. The stuff handling suite set up/tear down is a _little_ hairy, but it makes sense and does seem to be the ScalaTest-recommended way of doing this.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
helgridly: Releasing the project will fail (but then the allocator will auto-clear it up in two hours anyway). Consider this a nudge to have your tests not take an hour ;)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
helgridly: done

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
helgridly: done

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GPAllocFixtures.scala
helgridly: agree it's probably unnecessary, but i did it anyway

breilly2: Nit: Just be careful of unnecessary thread safety affecting readability. In this case, someone reading `AtomicBoolean` could reasonably assume that it was determined to be necessary, possibly communicating an incorrect model of the instance's thread context.

PR #82, opened 2018-02-14
GAWB-3093: Adding a new method to add and remove user from a Billing Project in trait BillingFixtures

aweng98:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -1,4 +1,5 @@
 package org.broadinstitute.dsde.workbench.fixture
+
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.{Config, UserPool}
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole.BillingProjectRole
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -2,10 +2,12 @@ package org.broadinstitute.dsde.workbench.fixture
 
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.{Config, UserPool}
+import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole.BillingProjectRole
-import org.broadinstitute.dsde.workbench.service.{Orchestration, Rawls}
 import org.broadinstitute.dsde.workbench.service.test.CleanUp
+import org.broadinstitute.dsde.workbench.service.{Orchestration, Rawls}
 import org.scalatest.TestSuite
+
 import scala.util.Random
 
 trait BillingFixtures extends CleanUp { self: TestSuite =>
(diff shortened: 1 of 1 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -28,5 +28,13 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
       } catch nonFatalAndLog(s"Error deleting billing project in withBillingProject clean-up: $billingProjectName")
     }
   }
+
+  def addUserInBillingProject(billingProjectName: String, email: String, role: BillingProjectRole)(implicit token: AuthToken): Unit = {
+    try {
+      Orchestration.billing.addUserToBillingProject(billingProjectName, email, role)
```
breilly2: Failing to add the user to the billing project should probably be a fatal error. I don't think you need to put this in a `try`. `CleanUp` puts a try around each of the cleanup functions already so you don't have to worry about it.

Is there something else you're trying to protect against that I'm not thinking of?

aweng98: I think you're right. removing the `try`

breilly2 (approved):
👍  There will probably be other cases where a pattern like this would be helpful if it works out. This will be a good test of it.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,7 +1,7 @@
 package org.broadinstitute.dsde.workbench.fixture
-
```
aednichols: Remove if not intentional

(It seems that we generally favor leaving a line between the `package` statement and `import` statements, but this is spit-and-polish that I generally don't get worked up about.)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -28,5 +28,11 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
       } catch nonFatalAndLog(s"Error deleting billing project in withBillingProject clean-up: $billingProjectName")
     }
   }
+
+  def addUserInBillingProject(billingProjectName: String, email: String, role: BillingProjectRole)(implicit token: AuthToken): Unit = {
```
aednichols: I still favor the idea of creating a fixture like
```
withUserInBillingProject(projectName, user.email, BillingProjectRole.Owner) {
   [...]
}
```
because it is avoids side-effects and is generally more Scala-ey.

The signature would be something like
```
def withUserInBillingProject(billingProjectName: String, email: String, role: BillingProjectRole)(testCode: () => Any)(implicit token: AuthToken): Unit
```

aednichols (approved):
👍

Looks fine to me as-is, but maybe Hussein should also review as TL?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
helgridly: There seems to be the `withFoo` pattern (aka loan pattern) and the "make a thing then `register cleanUp`" pattern. I don't have a good understanding of why both exist.

aednichols: The loan pattern is popular because the `register cleanUp` is often forgotten 😄 

In fact, that is exactly what Alex is working on fixing in this PR's parent ticket

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
aweng98: unintentional removal of line. it's back now.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
aweng98: Existing method `withBillingProject` (aka loan pattern) doesn't add & remove user from the Billing project. Some test authors forget to remove user from Billing project in test code as a final step. Thus for adding `register cleanup` now. Hussein's comments gave me an idea to refactor `withBillingProject` so that it has an optional param List[emails]. It would automatically add user and then remove user from the Billing project.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -16,17 +18,32 @@ trait BillingFixtures extends CleanUp { self: TestSuite =>
     Random.alphanumeric.take(length).mkString.toLowerCase
   }
 
-  def withBillingProject(namePrefix: String)
+  def withBillingProject(namePrefix: String, memberEmails: List[String] = List())
```
jmthibault79, line 21: Remember to bump the service-test version number because this is a breaking change

jmthibault79: Never mind, there's a default.  👍

MatthewBemis (approved):

MattPutnam (approved):
👍

PR #90, opened 2018-02-28
Make GoogleStorageDAO.deleteBucket handle bucket not existing

rtitle:
Needed by Leo for pause/resume clusters.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
@@ -66,12 +66,12 @@ class HttpGoogleStorageDAO(appName: String,
     val deleteObjectsFuture = if (recurse) {
       val listObjectsRequest = storage.objects().list(bucketName.value)
       retryWithRecoverWhen500orGoogleError { () =>
-        executeGoogleRequest(listObjectsRequest)
+        Option(executeGoogleRequest(listObjectsRequest))
       } {
         case e: HttpResponseException if e.getStatusCode == StatusCodes.NotFound.intValue => None
       }.flatMap { objects =>
         // Handle null responses from Google
-        val items = Option(objects).flatMap(objs => Option(objs.getItems)).map(_.asScala).getOrElse(Seq.empty)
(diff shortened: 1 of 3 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -65,9 +65,13 @@ class HttpGoogleStorageDAO(appName: String,
     // If `recurse` is true, first delete all objects in the bucket
     val deleteObjectsFuture = if (recurse) {
       val listObjectsRequest = storage.objects().list(bucketName.value)
-      retryWhen500orGoogleError(() => executeGoogleRequest(listObjectsRequest)).flatMap { objects =>
+      retryWithRecoverWhen500orGoogleError { () =>
+        Option(executeGoogleRequest(listObjectsRequest))
```
jmthibault79, line 69: Isn't this going to swallow all exceptions?

rtitle: It should just recover from 404s, making this `Future` a no-op. Other errors should result in a failed Future. We recover from 404s below in `deleteBucketRequest` as well.

I hit this in Leo when resuming a paused cluster -- it triggers this code to delete the init bucket, but the init bucket has already been deleted when the cluster started up the first time. So we don't want this method to fail on 404 errors.

jmthibault79: I mean the `Option()` here.  The recover for 404 looks fine.

rtitle: Oh sorry. `Option(_)` shouldn't swallow exceptions (unlike `Try`), but it should handle null responses. (I never know when Google APIs are going to return null)

jmthibault79: Ah!  Did not realize that.  Looks good.

jmthibault79 (approved):

PR #93, opened 2018-03-01
Add claimBillingProject() and releaseBillingProject()

jmthibault79:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -12,6 +12,12 @@ import org.scalatest.TestSuite
 
 import scala.util.Random
 
+
+/**
+  * Mix in this trait to allow your test to access billing projects managed by the GPAlloc system, or create new
+  * billing projects of your own.  Using GPAlloc will generally be much faster, limit the creation of billing projects
+  * to those tests which truly require them.
+  */
 trait BillingFixtures extends CleanUp {
   self: TestSuite =>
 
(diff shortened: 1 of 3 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -57,33 +72,72 @@ trait BillingFixtures extends CleanUp {
...
+    * Release a billing project back to GPAlloc when you are done with it.
+    * Consider using `withCleanBillingProject()` instead if you don't need to control the use of projects.
+    *
+    * @param project the GPAllocProject to release
+    * @param ownerCreds the Credentials of the current owner of the project
+    * @param memberEmails a List of member emails (as Strings) to remove as project members
+    */
+  def releaseBillingProject(project: GPAllocProject, ownerCreds: Credentials, memberEmails: List[String] = List()): Unit = {
+    val ownerToken = ownerCreds.makeAuthToken()
+    val adminToken = UserPool.chooseAdmin.makeAuthToken()
```
helgridly: i was going to say that this has to be the same admin token that registered the project, but it doesn't -- it's an admin token only for the purpose of "being an admin to call the rawls endpoint". phew

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -57,33 +72,72 @@ trait BillingFixtures extends CleanUp {
...
     //request a GPAlloced project as the potential new owner
     val newOwnerToken = newOwnerCreds.makeAuthToken()
-    GPAlloc.projects.requestProject(newOwnerToken) match {
-      case Some(project) =>
-        //the Rawls endpoint to register a precreated project needs to be called by a Rawls admin
-        //but it also takes the new owner's UserInfo in order to create the resource as them in Sam
-        val adminToken = UserPool.chooseAdmin.makeAuthToken()
-        val newOwnerUserInfo = UserInfo(OAuth2BearerToken(newOwnerToken.value), WorkbenchUserId("0"), WorkbenchEmail(newOwnerCreds.email), 3600)
-        Rawls.admin.claimProject(project.projectName, project.cromwellAuthBucketUrl, newOwnerUserInfo)(adminToken)
+    GPAlloc.projects.requestProject(newOwnerToken) map { project =>
```
helgridly: If GPAlloc doesn't have a project for you (which it might not), you're going to get a None here. That's going to be annoying for test writers to handle

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -57,33 +72,72 @@ trait BillingFixtures extends CleanUp {
...
-  def withCleanBillingProject(newOwnerCreds: Credentials, memberEmails: List[String] = List())(testCode: (String) => Any): Unit = {
+  /**
+    * Manually claim a project provisioned by GPAlloc and optionally add members.
+    * Consider using `withCleanBillingProject()` instead if you don't need to control the use of projects.
+    *
+    * @param newOwnerCreds Credentials representing a billing project owner to pass to billing project endpoints
+    * @param memberEmails a List of member emails (as Strings) to add as members of this project
+    * @return Some(GPAllocProject) if it succeeded, None if it did not
+    */
+  def claimBillingProject(newOwnerCreds: Credentials, memberEmails: List[String] = List()): Option[GPAllocProject] = {
```
helgridly: nit: rename to `claimGPAllocProject` so it's clear

helgridly (commented):
sideways thumb. as someone who wants to use the claim/release functions in a test, how do YOU want to handle the case where GPAlloc doesn't have a project for you?

MatthewBemis (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
jmthibault79: ok

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
jmthibault79: Yep, though we do need callers to know whether they get a GP Alloc'd project or a new one.

That implies to me a return type of `Either[GPAllocProject, GoogleProject]` though this doesn't do anything to relieve the caller of the responsibility to handle both cases.

jmthibault79:
One way for the calling test to deal with gpalloc returning None would be for it to call createBrandNew and then keep track of which it called.  Or it could use the `Either[]` construction.  The more I think about this, the more I like `claim()` returning an Either.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
dvoet: if the pool is exhausted, why doesn't gpalloc just create a new project and add it to the pool?

helgridly: [It does.](https://github.com/broadinstitute/gpalloc/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/gpalloc/service/GPAllocService.scala#L48) But that process takes minutes, so we return `HTTP Sorry No` and let the caller decide what they want to do.

helgridly: If you use `withCleanBillingProject`, that fallback is handled for you.

jmthibault79:
New plan: use a type that encapsulates whether it was GPAlloc'd or new, and knows how to clean up after itself.

helgridly (approved):
nice :+1:

PR #94, opened 2018-03-06
Fix GoogleStorageDAO NPE

rtitle:
I encountered this in Leo tests.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
@@ -193,9 +193,9 @@ class HttpGoogleStorageDAO(appName: String,
     val objects = listObjectsRecursive(getter) map { pagesOption =>
       pagesOption.map { pages =>
         pages.flatMap { page =>
-          Option(page.getItems.asScala) match {
+          Option(page.getItems) match {
             case None => List.empty
-            case Some(objects) => objects.toList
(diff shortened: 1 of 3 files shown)


helgridly (approved):
how sneaky

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -187,9 +187,9 @@ class HttpGoogleStorageDAO(appName: String,
     val objects = listObjectsRecursive(getter) map { pagesOption =>
       pagesOption.map { pages =>
         pages.flatMap { page =>
-          Option(page.getItems.asScala) match {
+          Option(page.getItems) match {
             case None => List.empty
-            case Some(objects) => objects.toList
+            case Some(objects) => objects.asScala.toList
```
jmthibault79, line 198: Good catch!  But let's rename this or the other `objects` in line 187

rtitle: Ah good call, yeah I don't like it when variables shadow other variables

helgridly:
@rtitle what's the status on this?

rtitle:
Oh I forgot about this. It fixes a NPE if you call `listObjectsWithPrefix` on an empty bucket. Leo tests call this method on the staging bucket, which is never empty, so it works. I'd still like to merge this however, in case anyone else runs into it.

PR #95, opened 2018-03-07
add logging when getting an access token fails

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
@@ -29,13 +29,6 @@ trait AuthToken extends LazyLogging {
           s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
           s"Access Token: ${cred.getAccessToken} \n" +
           s"SA Private Key ID: ${cred.getServiceAccountPrivateKeyId}")
-      case e: Exception =>
-        logger.error("Encountered 4xx error getting access token. Details: \n" +
-          s"Service Account: ${cred.getServiceAccountId} \n" +
-          s"User: ${cred.getServiceAccountUser} \n" +
-          s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
-          s"Access Token: ${cred.getAccessToken} \n" +
-          s"SA Private Key ID: ${cred.getServiceAccountPrivateKeyId}")
         throw e
     }
     cred.getAccessToken
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
@@ -20,7 +20,6 @@ trait AuthToken extends LazyLogging {
     val cred = buildCredential()
     try {
       cred.refreshToken()
-      throw new Exception("blah!")
     } catch {
       case e: TokenResponseException =>
         logger.error("Encountered 4xx error getting access token. Details: \n" +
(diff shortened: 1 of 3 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
```diff
@@ -16,7 +18,18 @@ trait AuthToken {
...
+    try {
+      cred.refreshToken()
+    } catch {
+      case e: TokenResponseException =>
+        logger.error("Encountered 4xx error getting access token. Details: \n" +
+          s"Service Account: ${cred.getServiceAccountId} \n" +
+          s"User: ${cred.getServiceAccountUser} \n" +
+          s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
+          s"Access Token: ${cred.getAccessToken} \n" +
+          s"SA Private Key ID: ${cred.getServiceAccountPrivateKeyId}")
```
aednichols: I would love for this exception handler to check if the service account actually exists and print that along with the output

We have had scenarios where one test deleted the FireCloud service account, and the following tests couldn't authenticate

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
```diff
@@ -1,10 +1,12 @@
 package org.broadinstitute.dsde.workbench.auth
 
+import com.google.api.client.auth.oauth2.TokenResponseException
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
+import com.typesafe.scalalogging.LazyLogging
 
-trait AuthToken {
+trait AuthToken extends LazyLogging {
```
aweng98, line 9: logging is a good enhancement.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
```diff
@@ -16,7 +18,18 @@ trait AuthToken {
 
   private def makeToken(): String = {
     val cred = buildCredential()
-    cred.refreshToken()
+    try {
+      cred.refreshToken()
```
aweng98, line 22: week 1 on this job, I had this in mind, but never tested out. 
```
  private def makeToken(): String = {
    val cred = buildCredential()
    val secToExpire: Option[java.lang.Long] = Option(cred.getExpiresInSeconds)
    if (secToExpire.forall(sec => sec < 300)) {  // if remaining time is less than 5 min
      cred.refreshToken()
    }
    cred.getAccessToken
  }
```

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
aweng98: other exception if any is just `case ex: Exception => throw ex`, right? I (we  😉) had some practice on `case` yesterday.

aednichols (approved):
👍 Seems fine, can implement suggestions if this isn't enough information

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
MatthewBemis: @aednichols i think that's an unrelated problem. IIRC, those tests were deleting the SAs from firecloud, but not from google. these SAs only need to be in google

@aweng98 any other exceptions will be thrown anyway. i'm just catching this particular type so i can inject some logging and then re-throw it

aweng98 (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
MatthewBemis: i will log the expire time because i think you're onto something with that, but i don't want to change the `makeToken` implementation too much yet without knowing the real problem. hopefully this logging will lead to real solutions :)

PR #98, opened 2018-03-15
stop ignoring cleanup failures

dvoet:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
@@ -37,9 +37,7 @@ trait WorkspaceFixtures { self: WebBrowserSpec with TestSuite =>
       testCode(workspaceName)
     } finally {
       if (cleanUp) {
-        try {
-          api.workspaces.delete(namespace, workspaceName)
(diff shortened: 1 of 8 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -37,9 +37,7 @@ trait WorkspaceFixtures { self: WebBrowserSpec with TestSuite =>
       testCode(workspaceName)
     } finally {
       if (cleanUp) {
-        try {
-          api.workspaces.delete(namespace, workspaceName)
-        } catch nonFatalAndLog(s"Error deleting workspace in withWorkspace clean-up: $namespace/$workspaceName")
+        api.workspaces.delete(namespace, workspaceName)
```
breilly2: This is going to need a change to the test of deleting a workspace.

breilly2 (approved):
Not successfully cleaning up causes failures because we hit Google limits. Failing on cleanup failures will make us even more prone to transient Google problems through no fault of the meaningful guts of the tests. Experience has shown the former to be more impactful than expected so I guess we should try this.

I am unhappy that it's coming to this but I will go on the record to approve it. Ultimately I think we need to write more tests that don't rely on real Google and reduce the tests that do.

aweng98 (approved):

PR #101, opened 2018-03-22
Add Rawls endpoints for GAWB-3293 testing

jmthibault79:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
@@ -1,9 +1,12 @@
 package org.broadinstitute.dsde.workbench.service
 
+import com.fasterxml.jackson.databind.JsonNode
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.Config
 import org.broadinstitute.dsde.workbench.model.UserInfo
+
+import scala.util.Try
 
 trait Rawls extends RestClient with LazyLogging {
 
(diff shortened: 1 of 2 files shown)


aweng98 (commented):
trait `BillingFixtures.makeRandomId` is `RandomUtil`, right?

jmthibault79:
Some klugey API parsing here, to enable only what's necessary for my ticket.

The "right" way for tests to interact with APIs is worth a discussion and will be captured in GAWB-3319.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -50,6 +53,58 @@ trait Rawls extends RestClient with LazyLogging {
       parseResponse(getRequest(url + s"api/workspaces"))
     }
   }
+
+  object submissions {
+    // dear reader: "workspace namespace" is an archaic term for billing project which is still used in a few places
```
breilly2: Then we should start using the preferred term in new code.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/MethodData.scala
```diff
@@ -45,9 +47,168 @@ object MethodData {
     documentation = "",
     payload = "task hello {\n  String name\n\n  command {\n    echo 'hello ${name}!'\n  }\n  output {\n    File response = stdout()\n  }\n  runtime {\n    docker: \"ubuntu\"\n  }\n}\n\nworkflow test {\n  call hello\n}"
   )
+
+  val SubWorkflowChildMethod = Method(
```
breilly2: This stuff is probably needed for a specific test, right? If so, it should be closer to that test. It bothered me a little when `MethodData` was in the firecloud-ui repo but it bothers me a lot now that it is in workbench-libs.

breilly2 (changes requested):
Unless there's a really good reason to have it here, I want to see test data closer to the test.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/MethodData.scala
helgridly: I am inclined to agree. If it's used in more than one test, it should be here; if not, it should be with the test.

Offtopic/related/the other side of the same coin: it'd be really nice to have some fixtures that the test environment spins up with (like our user pool). Workspaces and such.

jmthibault79: Makes sense.  I'll move it to the test.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
jmthibault79: true

breilly2 (approved):
👍 Thanks for making the change!

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -50,6 +54,56 @@ trait Rawls extends RestClient with LazyLogging {
       parseResponse(getRequest(url + s"api/workspaces"))
     }
   }
+
+  object submissions {
+    def launchWorkflow(billingProject: String, workspaceName: String, methodConfigurationNamespace: String, methodConfigurationName: String, entityType: String, entityName: String, expression: String, useCallCache: Boolean, workflowFailureMode: String = "NoNewCalls")(implicit token: AuthToken): String = {
+      val body = Map("methodConfigurationNamespace" -> methodConfigurationNamespace, "methodConfigurationName" -> methodConfigurationName, "entityType" -> entityType, "entityName" -> entityName, "expression" -> expression, "useCallCache" -> useCallCache, "workflowFailureMode" -> workflowFailureMode)
+      logger.info(s"Creating a submission: $billingProject/$workspaceName config: $methodConfigurationNamespace/$methodConfigurationName with body $body")
+      val response = postRequest(url + s"api/workspaces/$billingProject/$workspaceName/submissions", body)
```
aweng98, line 62: All good, only a short comment on the placement of log. I wish everyone logs action after action took place as the standard practice.

aweng98 (approved):

helgridly (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
jmthibault79: Disagree.  Logging-after means that there's no log when the action fails.

PR #102, opened 2018-03-23
add option to trim uuid and not prefix

helgridly:
**PR checklist**

In all cases:

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/package.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/package.scala
@@ -53,7 +53,7 @@ package object google {
     //if we're trimming the prefix, we make the prefix short enough to accommodate the UUID.
     //if we're trimming the UUID, we might still need to trim the prefix anyway if it's enormous.
     val trimmedPrefix = if(trimPrefix) sb.take(maxBucketNameLen - uuid.length) else sb.take(maxBucketNameLen)
-    val trimmedUUID = if(trimPrefix) uuid else sb.take(Math.min(maxBucketNameLen - trimmedPrefix.length, uuid.length))
+    val trimmedUUID = if(trimPrefix) uuid else uuid.take(Math.min(maxBucketNameLen - trimmedPrefix.length, uuid.length))
 
     // must not start with "goog" or contain the string "google"
     val processedName = trimmedPrefix.replaceAllLiterally("goog", "g00g")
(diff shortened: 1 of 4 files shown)


model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/package.scala
```diff
@@ -28,29 +28,36 @@ package object google {
...
 
-    // max length of 63 chars, including the uuid
     val uuid = UUID.randomUUID.toString
-    val maxNameLength = 63 - uuid.length - 1
-    if (sb.length > maxNameLength) sb.setLength(maxNameLength)
+
+    //if we're trimming the prefix, we make the prefix short enough to accommodate the UUID.
+    //if we're trimming the UUID, we might still need to trim the prefix anyway if it's enormous.
+    val trimmedPrefix = if(trimPrefix) sb.take(maxBucketNameLen - uuid.length) else sb.take(maxBucketNameLen)
+    val trimmedUUID = if(trimPrefix) uuid else sb.take(Math.min(maxBucketNameLen - trimmedPrefix.length, uuid.length))
```
jmthibault79: Shouldn't this be `uuid.take()` ?

model/src/test/scala/org/broadinstitute/dsde/workbench/model/google/GcsPathParserSpec.scala
```diff
@@ -20,6 +20,15 @@ class GcsPathParserSpec extends FlatSpecLike with Matchers with EitherValues {
     generateUniqueBucketName("myClusterWhichHasAVeryLongNameBecauseIAmExtremelyVerboseInMyDescriptions").value.length shouldBe 63
   }
 
+  "gcs" should "generate valid bucket names when trimming the uuid" in {
```
jmthibault79, line 23: add test case for true

jmthibault79: never mind, default is true

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/package.scala
helgridly: how did my tests not catch this

model/src/test/scala/org/broadinstitute/dsde/workbench/model/google/GcsPathParserSpec.scala
helgridly: the default is true, it's the block above

PR #105, opened 2018-04-04
[Leo Issue 215] Reduce time for cluster create request

vkumra-broad:
**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
@@ -11,6 +11,7 @@ import scala.concurrent.Future
 trait GoogleStorageDAO {
 
   def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName]
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, entity: GcsEntity, role: GcsRole,  lifecycleAge: Int, lifecycleType: GcsLifecycleType = Delete): Future[GcsBucketName]
   def deleteBucket(bucketName: GcsBucketName, recurse: Boolean): Future[Unit]
   def bucketExists(bucketName: GcsBucketName): Future[Boolean]
 
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
@@ -10,7 +10,7 @@ import scala.concurrent.Future
 
 trait GoogleStorageDAO {
 
-  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, reader: List[GcsEntity] = List.empty, owner: List[GcsEntity] = List.empty): Future[GcsBucketName]
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, readers: List[GcsEntity] = List.empty, owners: List[GcsEntity] = List.empty): Future[GcsBucketName]
   def deleteBucket(bucketName: GcsBucketName, recurse: Boolean): Future[Unit]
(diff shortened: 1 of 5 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -300,4 +301,22 @@ class HttpGoogleStorageDAO(appName: String,
     }
   }
 
+  override def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, readerEntity: List[GcsEntity], ownerEntity: List[GcsEntity]): Future[GcsBucketName] = {
```
jmthibault79: Move up in file to other `createBucket()`

jmthibault79:
Looks good.  Add a note about this in the google CHANGELOG and replace the versions in that file with `TRAVIS-REPLACE-ME`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -11,6 +11,7 @@ import scala.concurrent.Future
 trait GoogleStorageDAO {
 
   def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName]
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, readerEntity: List[GcsEntity], ownerEntity: List[GcsEntity]): Future[GcsBucketName]
```
helgridly: just looking at the signatures: can we replace the 2-param function with default values on the 4-param?

rtitle: +1, can probably write one in terms of the other:
```
def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName] = {
  createBucket(billingProject, bucketName, List.empty, List.empty)
}
```

rtitle: Or even get rid of the other one and write this as:

```
def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, readerEntity: List[GcsEntity] = List.empty, ownerEntity: List[GcsEntity] = List.empty): Future[GcsBucketName]
```

rtitle: Also regarding variable naming, maybe `readerEntity --> readers` and `ownerEntity --> owners` would be more clear.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.workbench.google
 
+import java._
```
rtitle: ? I suspect this is not used.

rtitle (changes requested):
Just for removing code duplication of the 2 `createBucket` methods

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageDAO.scala
```diff
@@ -10,7 +10,7 @@ import scala.concurrent.Future
 
 trait GoogleStorageDAO {
 
-  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName): Future[GcsBucketName]
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, reader: List[GcsEntity] = List.empty, owner: List[GcsEntity] = List.empty): Future[GcsBucketName]
```
jmthibault79: Pluralize: readers, owners

rtitle (approved):

PR #108, opened 2018-04-11
Retry token creation on response code 400

gpolumbo-broad:
Addresses error scene here:  https://fc-jenkins.dsp-techops.broadinstitute.org/job/firecloud-ui-fiab-test-runner/2014/consoleFull

Per @dvoet I have made the change to retry token creation if the response code is 400 or 401.  

**PR checklist**

In all cases:

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -79,7 +79,7 @@ object Settings {
   val serviceTestSettings = commonSettings ++ List(
     name := "workbench-service-test",
     libraryDependencies ++= serviceTestDependencies,
-    version := createVersion("0.8")
+    version := createVersion("0.9")
   ) ++ publishSettings
 
   val rootSettings = commonSettings ++ noPublishSettings ++ noTestSettings
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -79,7 +79,7 @@ object Settings {
   val serviceTestSettings = commonSettings ++ List(
     name := "workbench-service-test",
     libraryDependencies ++= serviceTestDependencies,
-    version := createVersion("0.9")
+    version := createVersion("0.8")
   ) ++ publishSettings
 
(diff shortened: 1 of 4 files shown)


aweng98:
```
Optimizing OAuth2 requests
If your app doesn't share credentials across servers, processes, and threads, it may be sending an excessive number of requests to Google. This can cause our servers to enforce rate limits on your app, resulting in performance degradation.
```
An interesting reading from https://developers.google.com/adwords/api/docs/guides/authentication. automation testing definitely falls into this category.

serviceTest/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 This file documents changes to the `workbench-service-test` library, including notes on how to upgrade to new versions.
 
+## 0.9
```
MatthewBemis: you'll also want to add a line in this section like ```SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.9-TRAVIS-REPLACE-ME"````

(see previous versions)

MatthewBemis (approved):
1 reminder then :+1:

project/Settings.scala
```diff
@@ -79,7 +79,7 @@ object Settings {
   val serviceTestSettings = commonSettings ++ List(
     name := "workbench-service-test",
     libraryDependencies ++= serviceTestDependencies,
-    version := createVersion("0.8")
+    version := createVersion("0.9")
```
dvoet: I don't think this warrants a version tick, it is backwards compatible

dvoet (changes requested):
👍 on code but no need to increase the version

dvoet (approved):

dvoet:
but still need to do the travis change me thing for the hash

MatthewBemis (approved):

PR #110, opened 2018-04-19
[Leo Issue 317] Adding GcsEntityType Project

kyuksel:
**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" 
 
 Contains generic, externally-facing model classes used across Workbench.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.10-255a466"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.11-TRAVIS-REPLACE-ME"`
 
 [Changelog](model/CHANGELOG.md)
 
@@ -35,7 +35,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
 
 To depend on the `MockGoogle*` classes, additionally depend on:
 
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
 
 To depend on the `MockGoogle*` classes, additionally depend on:
 
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -4,7 +4,7 @@ This file documents changes to the `workbench-google` library, including notes o
 
 ## 0.16
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
 
 ### Added
 - org.broadinstitute.dsde.workbench.google.GoogleIamDAO#findServiceAccount(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchEmail)
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -4,7 +4,7 @@ This file documents changes to the `workbench-google` library, including notes o
 
 ## 0.16
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
 
 ### Added
 - org.broadinstitute.dsde.workbench.google.GoogleIamDAO#findServiceAccount(serviceAccountProject: GoogleProject, serviceAccountEmail: WorkbenchEmail)
--- a/google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
+++ b/google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
@@ -172,17 +172,3 @@ class GoogleJsonSpec extends FlatSpecLike with Matchers {
     rqRead shouldBe gooRq
   }
 }
-
-class GcsEntitySpec extends FlatSpecLike with Matchers {
-  private val emailGcsEntity = EmailGcsEntity(User, WorkbenchEmail("foo@bar.com"))
-
-  "EmailGcsEntity stringification" should "work" in {
-    emailGcsEntity.toString shouldBe "user-foo@bar.com"
-  }
-
-  private val projectGcsEntity = ProjectGcsEntity(Viewers, ProjectNumber("398512454")).toString
-
-  "ProjectGcsEntity stringification" should "work" in {
-    projectGcsEntity.toString shouldBe "project-viewers-398512454"
-  }
-}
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
```
rtitle: Also set `TRAVIS-REPLACE-ME` in https://github.com/broadinstitute/workbench-libs/blob/develop/google/CHANGELOG.md

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
```diff
@@ -57,10 +57,12 @@ object GcsEntityTypes {
   sealed trait GcsEntityType extends ValueObject
   case object User extends GcsEntityType { val value = "user" }
   case object Group extends GcsEntityType { val value = "group" }
+  case object Project extends GcsEntityType { val value = "project" }
```
rtitle, line 75: I'm wondering if we need to change the model a little more to support project-level ACLs. Currently we have:

```
 case class GcsEntity(email: WorkbenchEmail, entityType: GcsEntityType) {
   override def toString: String = s"${entityType.value}-${email.value}"
 }
```

This works for users and groups, where the target is an email (see https://cloud.google.com/storage/docs/json_api/v1/bucketAccessControls). But for projects the format is: `project-team-projectId`, where `team` is one of (owners, editors, viewers) and `projectId` is a string. I don't think `GcsEntity` as it exists right now would support this.

rtitle: Maybe `GcsEntity` itself should be a trait, so we could have something like this:

```
sealed trait GcsEntity {
  val entityType: GcsEntityType
}
case class EmailGcsEntity(email: WorkbenchEmail, entityType: GcsEntityType) extends GcsEntity {
  override def toString: String = s"${entityType.value}-${email.value}"
}
case class ProjectGcsEntity(team: String, projectNumber: String, entityType: GcsEntityType) extends GcsEntity {
  override def toString: String = s"${entityType.value}-$team-$projectNumber"
} 
```

Just an idea.

kyuksel: I like your idea. One challenge is `jsonFormat`ing since `{Email,Project}GcsEntity` have different members. In particular for deserialization, we know only at runtime, what the entity type is. This may require an ugly regex on the `entity` string. Let me know if you have a better idea. 

I could also break the `GcsEntity` ADT but then users would need to distinctly specify what entity type they mean to use, which is probably too much inconvenience.

rtitle: Yeah, I guess the `GcsEntity` deserializer would need to parse the string into the correct `GcsEntity` subclass. There's probably no way around that. One note is I'm not sure these types _need_ JSON serializers since they're for internal use, not exposed over REST APIs. But by convention we've been specifying serializers for all our model classes.

kyuksel:
From the guidelines on `CONTRIBUTING.md`, it feels like this should be a major version bump since `GcsEntity` is now a trait and the users of it (which is only Leonardo as far as I can tell but still) will need to use one of `EmailGcsEntity` or `ProjectGcsEntity` as implementation. Any thoughts?

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
```diff
@@ -101,5 +124,4 @@ object GoogleModelJsonSupport {
   implicit val GcsLifecycleTypeFormat = ValueObjectFormat(GcsLifecycleTypes.withName)
   implicit val GcsRoleFormat = ValueObjectFormat(GcsRoles.withName)
   implicit val GcsEntityTypeFormat = ValueObjectFormat(GcsEntityTypes.withName)
-  implicit val GcsEntityFormat = jsonFormat2(GcsEntity)
```
kyuksel, line 104: It didn't feel worth creating a custom `RootJsonFormat` extension for `GcsEntity` since it isn't exposed via the APIs or serialized to JSON elsewhere.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
kyuksel: Yea, I realized (unfortunately after starting creating a custom formatter) that the runtime parsing of the `entity` field of the [`BucketAccessControl` resource representation JSON](https://cloud.google.com/storage/docs/json_api/v1/bucketAccessControls) received from Google is irrelevant to this matter and dropped it.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
```diff
@@ -167,4 +171,18 @@ class GoogleJsonSpec extends FlatSpecLike with Matchers {
 
     rqRead shouldBe gooRq
   }
-}
\ No newline at end of file
+}
+
+class GcsEntitySpec extends FlatSpecLike with Matchers {
```
rtitle: It doesn't seem like this belongs in `GoogleUtilitiesSpec` -- seems like it should be its own class in https://github.com/broadinstitute/workbench-libs/tree/develop/model/src/test/scala/org/broadinstitute/dsde/workbench/model/google

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
```diff
@@ -4,12 +4,13 @@ import java.time.Instant
...
 import org.broadinstitute.dsde.workbench.model._
-import org.broadinstitute.dsde.workbench.model.google.GcsEntityTypes.GcsEntityType
-import org.broadinstitute.dsde.workbench.model.google.GcsRoles.GcsRole
+import org.broadinstitute.dsde.workbench.model.google.GcsEntityTypes.{GcsEntityType, Project}
+import org.broadinstitute.dsde.workbench.model.google.ProjectTeamTypes.ProjectTeamType
 import spray.json.{JsString, JsValue, RootJsonFormat}
 
 // Projects
 case class GoogleProject(value: String) extends ValueObject
+case class ProjectNumber(value: String) extends AnyVal
```
rtitle: Should probably extend `ValueObject` not `AnyVal`

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
```diff
@@ -53,20 +54,42 @@ object GcsRoles {
...
+object ProjectTeamTypes {
+  sealed trait ProjectTeamType extends ValueObject
+  case object Viewers extends ProjectTeamType { val value = "viewers" }
+  case object Editors extends ProjectTeamType { val value = "editors" }
+  case object Owners extends ProjectTeamType { val value = "owners" }
+
+  def withName(name: String): ProjectTeamType = name.toLowerCase() match {
+    case "viewers" => Viewers
+    case "editors" => Editors
+    case "owners" => Owners
```
rtitle, line 66: These `withName` methods should have a default case which throws `WorkbenchException`

google/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ This file documents changes to the `workbench-google` library, including notes o
 
 ## 0.16
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-b070188"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
```
rtitle: Since we changed `GcsEntity` this probably warrants a version bump from 0.16 to 0.17

rtitle: The change is in workbench-model, not google, so update the CHANGELOG and README accordingly.

Also since we're changing GcsEntity it probably warrants a minor version bump (0.10 to 0.11 I think).

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
rtitle: 👍

rtitle (changes requested):
A few minor requests, see above. The `GcsEntity` ADT itself looks good.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilitiesSpec.scala
```diff
@@ -8,6 +8,10 @@ import com.google.api.client.googleapis.json.GoogleJsonError.ErrorInfo
 import com.google.api.client.googleapis.json.{GoogleJsonError, GoogleJsonResponseException}
 import com.google.api.client.http._
 import org.broadinstitute.dsde.workbench.metrics.{StatsDTestUtils, WorkbenchInstrumented}
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
```
rtitle: Probably unused imports?

rtitle (approved):
👍 !

PR #116, opened 2018-05-03
Scalatest Selenium version upgrade and AuthToken cache solution

aweng98:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
@@ -7,10 +7,10 @@ import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 
 
-case class Credentials (val email: String, val password: String) extends LazyLogging {
+case class Credentials (email: String, password: String) extends LazyLogging {
   def makeAuthToken(): AuthToken = {
     val auth = Credentials.cache.get(this)
-    logger.debug(s"AuthToken: ${auth}, ${auth.value}")
+    // logger.debug(s"AuthToken: ${auth}, ${auth.value}")
     auth
   }
 }
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
@@ -10,7 +10,7 @@ import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 case class Credentials (val email: String, val password: String) extends LazyLogging {
   def makeAuthToken(): AuthToken = {
     val auth = Credentials.cache.get(this)
-    logger.debug(s"AuthToken: ${auth}")
+    logger.debug(s"AuthToken: ${auth}, ${auth.value}")
     auth
   }
 }
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
@@ -1,7 +1,23 @@
 package org.broadinstitute.dsde.workbench.config
 
+import java.util.concurrent.TimeUnit
+
+import com.google.common.cache.{CacheBuilder, CacheLoader}
 import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 
-case class Credentials (email: String, password: String) {
-  def makeAuthToken(): AuthToken = UserAuthToken(this)
+
+case class Credentials (val email: String, val password: String) {
+  def makeAuthToken(): AuthToken = Credentials.cache.get(this)
+}
+
+object Credentials {
+  val cache = CacheBuilder.newBuilder()
+    .expireAfterWrite(3600, TimeUnit.SECONDS)
+    .build(
+      new CacheLoader[Credentials, AuthToken] {
+        def load(credentials: Credentials): AuthToken = {
+          UserAuthToken(credentials)
+        }
+      }
+    )
 }
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
@@ -3,11 +3,16 @@ package org.broadinstitute.dsde.workbench.config
 import java.util.concurrent.TimeUnit
 
 import com.google.common.cache.{CacheBuilder, CacheLoader}
+import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 
 
-case class Credentials (val email: String, val password: String) {
-  def makeAuthToken(): AuthToken = Credentials.cache.get(this)
+case class Credentials (val email: String, val password: String) extends LazyLogging {
+  def makeAuthToken(): AuthToken = {
+    val auth = Credentials.cache.get(this)
+    logger.debug(s"AuthToken: ${auth}")
+    auth
+  }
 }
 
 object Credentials {
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -8,26 +8,17 @@ import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail, Workbe
 import org.broadinstitute.dsde.workbench.service.{GPAlloc, Orchestration, Rawls}
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole.BillingProjectRole
-import org.broadinstitute.dsde.workbench.service.test.CleanUp
+import org.broadinstitute.dsde.workbench.service.test.{CleanUp, RandomUtil}
 import org.broadinstitute.dsde.workbench.service.util.ExceptionHandling
 import org.scalatest.TestSuite
-
-import scala.util.Random
-
 
 /**
   * Mix in this trait to allow your test to access billing projects managed by the GPAlloc system, or create new
   * billing projects of your own.  Using GPAlloc will generally be much faster, limit the creation of billing projects
   * to those tests which truly require them.
   */
-trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
+trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp with RandomUtil {
   self: TestSuite =>
-
-  // copied from WebBrowserSpec so we don't have to self-type it
-  // TODO make it common to API and browser tests
-  private def makeRandomId(length: Int = 7): String = {
-    Random.alphanumeric.take(length).mkString.toLowerCase
-  }
 
   protected def addMembersToBillingProject(projectName: String, memberEmails: List[String])(implicit token: AuthToken): Unit = {
     memberEmails foreach { email =>
@@ -59,7 +50,7 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
   }
 
   private def createNewBillingProject(namePrefix: String, memberEmails: List[String] = List())(implicit token: AuthToken): String = {
-    val billingProjectName = namePrefix + "-" + makeRandomId()
+    val billingProjectName = randomIdWithPrefix(namePrefix)
     Orchestration.billing.createBillingProject(billingProjectName, Config.Projects.billingAccountId)
     addMembersToBillingProject(billingProjectName, memberEmails)
     billingProjectName
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -92,6 +92,7 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
           deleteBillingProject(billingProject, memberEmails)
         }
       case Failure(e) =>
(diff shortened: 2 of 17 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/MethodData.scala
```diff
@@ -45,6 +45,16 @@ object MethodData {
     documentation = "",
     payload = "task hello {\n  String name\n\n  command {\n    echo 'hello ${name}!'\n  }\n  output {\n    File response = stdout()\n  }\n  runtime {\n    docker: \"ubuntu\"\n  }\n}\n\nworkflow test {\n  call hello\n}"
   )
+
+  val DataSpecMethod = Method(
```
aweng98: this method is to be used for DataSpec exclusively.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/WebBrowserSpec.scala
```diff
@@ -58,27 +58,26 @@ trait WebBrowserSpec extends WebBrowserUtil with ExceptionHandling with LazyLogg
...
       options.addArguments("--proxy-server=http://127.0.0.1:8080")
     }
     // Note that download.prompt_for_download will be ignored if download.default_directory is invalid or doesn't exist
     options.setExperimentalOption("prefs", Map(
       "download.default_directory" -> fullDownloadPath,
       "download.prompt_for_download" -> "false").asJava)
-    val capabilities = DesiredCapabilities.chrome
-    capabilities.setCapability(ChromeOptions.CAPABILITY, options)
-    capabilities
+    options
```
aweng98, line 74: get rid of selenium warning

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/WebBrowserUtil.scala
```diff
@@ -48,7 +50,7 @@ trait WebBrowserUtil extends WebBrowser {
       * @param timeOutInSeconds number of seconds to wait for the condition to be true
       * @param webDriver implicit WebDriver for the WebDriverWait
       */
-    def condition(condition: => Boolean, timeOutInSeconds: Long = defaultTimeOutInSeconds)(implicit webDriver: WebDriver): Unit = {
+    def condition(condition: => Boolean, timeOutInSeconds: Long = defaultTimeOutInSeconds)(implicit webDriver: WebDriver): Boolean = {
```
aweng98, line 53: change return type from `Unit` to `Boolean` to extends its' usefulness.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -431,7 +431,7 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
       val trialProjects: TrialProjects = countTrialProjects()
       if (trialProjects.available < count) {
         postRequest(apiUrl(s"api/trial/manager/projects?operation=create&count=${count - trialProjects.available}"))
-        Retry.retry(20.seconds, 10.minutes)({
+        Retry.retry(30.seconds, 20.minutes)({
```
aweng98, line 434: Free trial project could take more than 10 minutes to create depending on Google service.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/MethodData.scala
jmthibault79: Then it should live in `DataSpec` instead of here

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -82,15 +82,17 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
...
-    try {
-      testCode(billingProjectName)
-    } catch {
-      case t: Exception =>
-        logger.error("BillingFixtures.withBrandNewBillingProject Exception: ", t)
-        throw t // end test execution
-    } finally {
-      deleteBillingProject(billingProjectName, memberEmails)
+    Try(
+      createNewBillingProject(namePrefix, memberEmails)
```
jmthibault79: Why change this?  Before this change, we would clean up a partially-created project which hit a failure partway through.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GroupFixtures.scala
```diff
@@ -20,23 +21,23 @@ trait GroupFixtures extends ExceptionHandling with LazyLogging { self: TestSuite
                (implicit token: AuthToken): Unit = {
     val groupName = Util.appendUnderscore(namePrefix) + Util.makeUuid
 
-    try {
```
jmthibault79: same problem here

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
aweng98: so old way doesn't clean up anything because it was outside of try-catch-finally block.
My thought was this, there is nothing to cleanup because nothing created successfully.  I may be wrong about that. If billing project creation fails half way or timed out from api call, What does `deleteBillingProject` cleans up? something or nothing? I didn't want to see exception from clean up when it tries to clean up nothing. it leads to more exception.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
```diff
@@ -1,7 +1,23 @@
...
 import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 
-case class Credentials (email: String, password: String) {
-  def makeAuthToken(): AuthToken = UserAuthToken(this)
+
+case class Credentials (val email: String, val password: String) {
+  def makeAuthToken(): AuthToken = Credentials.cache.get(this)
 }
+
+object Credentials {
```
aweng98: a good idea from Rob Title!

reason for change:
1) api calls take bandwidth and time.
2) infrequently causes test to fail when one test is still using a token while other tests request new token for the same test user. Newly issued token turn old token to expire.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
jmthibault79: OK.  Why remove the `logger.error()` though?

aweng98:
jenkin tests passed
https://fc-jenkins.dsp-techops.broadinstitute.org/job/firecloud-ui-PR-fiab-test/1903/

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
aweng98: good catch! I overlooked. It's back in all fixtures now.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
aweng98: so Retry.retry actually doesn't fail when specified time is up. Anyone notice it too?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/WebBrowserSpec.scala
```diff
@@ -58,27 +58,26 @@ trait WebBrowserSpec extends WebBrowserUtil with ExceptionHandling with LazyLogg
     }
   }
 
-  private def getChromeIncognitoOption(downloadPath: String, headless: Boolean): DesiredCapabilities = {
+  private def getChromeIncognitoOption(downloadPath: String, headless: Boolean): ChromeOptions = {
     val fullDownloadPath = if (headless) s"/app/$downloadPath" else new File(downloadPath).getAbsolutePath
     logger.info(s"Chrome download path: $fullDownloadPath")
     val options = new ChromeOptions
     options.addArguments("--incognito")
+    options.addArguments("--no-experiments")
```
aweng98, line 66: I think this option makes `withSignInReal` works again on local chromedriver.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
jmthibault79: No - I mean the logger error in the test code catch block which is no longer present.

aweng98: ah for real it's back.

aweng98:
test passed again. one unrelated failure.
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/5480/

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
```diff
@@ -1,7 +1,28 @@
...
+import java.util.concurrent.TimeUnit
+
+import com.google.common.cache.{CacheBuilder, CacheLoader}
+import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.workbench.auth.{AuthToken, UserAuthToken}
 
-case class Credentials (email: String, password: String) {
-  def makeAuthToken(): AuthToken = UserAuthToken(this)
+
+case class Credentials (val email: String, val password: String) extends LazyLogging {
```
rtitle: Minor: you don't need to specify `val` in case classes

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -82,15 +82,22 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
...
+    ) match {
+      case Success(billingProject) =>
+        try {
+          testCode(billingProject)
+        } catch {
+          case ex: Exception =>
+            logger.error("", ex)
```
rtitle: It seems you lost the error message here

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -82,15 +82,22 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp {
...
+            fail(ex)
```
rtitle: Should this method be throwing `AssertionError`? Seems better to just log-and-rethrow the exception (which is what the previous code was doing).

aweng98:
passed https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/5563/

aweng98:
passed https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/5564/

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GroupFixtures.scala
```diff
@@ -20,23 +21,28 @@ trait GroupFixtures extends ExceptionHandling with LazyLogging { self: TestSuite
...
-        logger.error("GroupFixtures.withGroup Exception: ", t)
-        throw t // end test execution
-    } finally {
-      memberEmails foreach { email =>
-        Orchestration.groups.removeUserFromGroup(groupName, email, GroupRole.Member)
-      }
-      Orchestration.groups.delete(groupName)
+    } match {
+      case Success(s) =>
+        try {
```
rtitle: Again I'm not clear on the need for a nested `try`. It looks like you only clean up groups if the inner try fails, not if the outer one fails. But the outer `try` could have _partially_ succeeded and then failed for 1 email, in which case you'd still want to clean up. Is there harm in having the cleanup in the outermost finally block?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/MethodFixtures.scala
```diff
@@ -7,30 +7,42 @@ import org.broadinstitute.dsde.workbench.service.util.ExceptionHandling
...
-    } catch {
-      case t: Exception =>
-        logger.error("MethodFixtures.withMethod Exception: ", t)
-        throw t // end test execution
-    } finally {
-      if (cleanUp) {
+
+    Try {
+      for (i <- 1 to numSnapshots)
+        Orchestration.methods.createMethod(method.creationAttributes + ("name" -> methodName))
```
rtitle: Again if this succeeds for some elements but 1 of them throws an exception, the successful ones wouldn't get cleaned up.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
rtitle: I thought it should return the result of the most recent attempt if the time is up.

rtitle (changes requested):
My main comments are about the change to nested `try`s, which I think could miss cleaning up data in some cases.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
aweng98:
```
def retry[T](interval: FiniteDuration, timeout: FiniteDuration)(op: => Option[T]): Option[T] = {
    val iterations = (timeout / interval).round.toInt
    retry(Seq.fill(iterations)(interval))(op)
  }
```
No. retry stops when Seq is exhausted, not bound by duration time after all.

aweng98:
Undone code change in all fixture files. It was too much of change in one PR. I will make a new PR separate for fixture code later.

jmthibault79 (approved):
Looks good after removing fixture change.  Please add something to the CHANGELOG about Credentials.

rtitle (approved):
:shipit:

PR #120, opened 2018-05-11
Use standard SQL for parameterized query

ghost:
We're using an older version of BQ which is defaulted to legacy SQL. I was going by the current version documentation, in which standard SQL is the default. Set legacy SQL to false. 

**PR checklist**

In all cases:


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleBigQueryDAO.scala
```diff
@@ -48,6 +48,7 @@ class HttpGoogleBigQueryDAO(appName: String,
     val job = new Job()
       .setConfiguration(new JobConfiguration()
         .setQuery(new JobConfigurationQuery()
+          .setUseLegacySql(java.lang.Boolean.FALSE)
```
jmthibault79: Please add a comment about why this is needed

davidangb (approved):
I'd just use `false` instead of `java.lang.Boolean.FALSE` and let scalac handle the optimizations - but no biggie either way. :+1:

jmthibault79 (approved):

PR #128, opened 2018-05-23
genericize appendWithUuid & encode URIs

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
@@ -16,6 +16,7 @@ import org.broadinstitute.dsde.workbench.util.Retry
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContextExecutor, Future}
+import scala.util.Try
 
 trait RestClient extends Retry with LazyLogging {
   implicit val system = ActorSystem()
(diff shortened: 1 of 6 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/util/Util.scala
```diff
@@ -19,6 +19,13 @@ object Util extends LazyLogging {
     }
   }
 
+  def appendDelimiter(string: String, delimiter: String): String = {
+    string match {
+      case "" => ""
```
jmthibault79: I'm confused by this case, at least in a function called `appendDelimiter`, because it does not do so here.  Skipping this function call and calling `randomUuid` directly would have the same result.

MatthewBemis: i agree it's weird, and to be honest i don't know why it;s there. this was a copy-paste-modify of `appendUnderscore` above which does the same thing. i didn't know if there was some weird case that this was protecting against so i left it...

requires further investigation...

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala
```diff
@@ -34,6 +35,18 @@ trait RestClient extends Retry with LazyLogging {
     headers.Authorization(OAuth2BearerToken(token.value))
   }
 
+  def encodeUri(path: String): String = {
+    val pattern = """(https?)?:\/{2}+([\dA-z_.-]+)+(:[\d]+)??(\/[~0-9A-z\#\+\%@\.\/_ -]+)(\?[0-9A-z\+\%@\/&\[\];=_-]+)?""".r
```
MatthewBemis, line 39: it was stupid that i had to do this but i tried many other options. this was the only way that I could get the URI to properly encode without encoding too much or too little.

in a nutshell this solution rips out the scheme, the host, the port (if it exists), the path (if it exists), and the params (if they exist) and encodes only the necessary parts by relying on the encoding in the Uri class of akka-http. it handles things properly if you pass in each piece separated, but encodes too much (and barfs on some inputs) if you pass in the entire path at once- hence the deconstruction 

why akka-http's URI library can't do the ripping apart on my behalf is beyond me

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/util/Util.scala
MatthewBemis: only thing i can think of is that it's to protect against cases where a certain resource (in google or in firecloud) can't start with a special character, but yes `randomUuid` would have the same result

breilly2: Wow, this code has been shuffled up a bit. I think I originally wrote `appendUnderscore` as an internal function for putting an underscore between a prefix and a generated value, but not prepend an underscore if no prefix is given. I remember going back and forth between taking a `String` and an `Option[String]` so this code style may be left over from an attempt at using `Option[String]`. I might do it differently now, and I certainly never intended `appendUnderscore` to be a public function, divorced from the context of its intended purpose.

gpolumbo-broad (approved):

dvoet (approved):

PR #131, opened 2018-05-29
downloadpath

aweng98:
work to enable DataDownloadSpec tests to run in parallel

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains common classes and utilities for writing tests against Workbench REST web services.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.9-fc954ca" % "test" classifier "tests"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.9-TRAVIS-REPLACE-ME" % "test" classifier "tests"`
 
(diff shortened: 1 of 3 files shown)


breilly2:
Before you spend too much more time on this, I want to point out that this is going to be difficult to get to work. As you probably noticed, FireCloud does not make the default download filename unique so the only way to keep tests from stepping on each other is to save to different directories. Those directories need to be writable by Chrome (in order to keep Chrome from showing a system save dialog) and readable by the test. Since they're running in different docker containers, this is a challenge.

The tests currently use a single directory for downloads and open up the permissions here: https://github.com/broadinstitute/firecloud-automated-testing/blob/master/run-tests.sh#L85 before launching the tests. This script has sufficient privileges to change the directory permissions.

Since the test would need to choose a unique directory name but doesn't have enough privileges to `chmod` from the test, I'm not sure it's going to work without some significant changes... and I don't immediately have any ideas for what those changes could be. 🙁

Let me know if you'd like to discuss further.

aweng98:
@breilly2 Hi, I have a separate PR in firecloud-ui. It wasn't easy. Learning by trial & error, I found a way. Here is one `DataDownloadSpec` test log.
https://fc-jenkins.dsp-techops.broadinstitute.org/job/firecloud-ui-fiab-test-runner/3497/artifact/test-reports/TEST-org.broadinstitute.dsde.firecloud.test.metadata.DataDownloadSpec.log

gdlex4015 (approved):

PR #138, opened 2018-06-19
Add GoogleIamDAO.testIamPermission

eap:
This PR adds the testIamPermission method to GoogleIamDAO and child classes.

**PR checklist**
Determination: no version bump. Reason: addition of new public functions on classes.

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" 
 
 Contains generic, externally-facing model classes used across Workbench.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.11-2ce3359"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.11-TRAVIS-REPLACE-ME`
 
 [Changelog](model/CHANGELOG.md)
 
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google
 
 To depend on the `MockGoogle*` classes, additionally depend on:
 
-`"org.broadinstitute.dsde.workbench" %% "workbench-google"  % "0.16-f339f30" % "test" classifier "tests"`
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-f339f30"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.16-TRAVIS-REPLACE-ME"`
```
rtitle, line 42: Thanks for finding and doing this trick. :) I think there's a line below this (L46) that also needs to be TRAVIS-REPLACE-ME'd.

eap: Done

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
```diff
@@ -22,6 +22,10 @@ case class ServiceAccountKeyId(value: String) extends ValueObject
 case class ServiceAccountPrivateKeyData(value: String) extends ValueObject with Base64Support
 case class ServiceAccountKey(id: ServiceAccountKeyId, privateKeyData: ServiceAccountPrivateKeyData, validAfter: Option[Instant], validBefore: Option[Instant])
 
+// Test IAM types.
+case class IamPermission(value: String) extends ValueObject
+case class IamTestResponse(permissions: Option[List[IamPermission]])
```
rtitle: Seems like `IamTestResponse` may not be used? If not then delete it.

rtitle:
Looks good, a couple minor comments above. Will check Leo PR next.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/google/GoogleModel.scala
eap: Good catch, also, looks like I should have updated the changelog and version hash for workbench-model. That is also done now.

rtitle (approved):

PR #140, opened 2018-06-21
Added functions to serviceTest to utilize adopt and scratch project endpoints in Orch

ghost:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
@@ -515,6 +515,30 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
       trialProjectReports
     }
 
+    def adoptTrialProject(projectName: String)(implicit token: AuthToken): TrialProjectReport = {
+      logger.info(s"API post request: api/trial/manager/projects?operation=adopt&projectName=$projectName")
+      Try(postRequest(apiUrl(s"api/trial/manager/projects?operation=adopt&projectName=$projectName"))) match {
+        case Success(response) =>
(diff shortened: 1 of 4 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -515,6 +515,30 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
+      logger.info(s"API post request: api/trial/manager/projects?operation=scratch&projectName=$projectName")
+      Try(postRequest(apiUrl(s"api/trial/manager/projects?operation=scratch&projectName=$projectName"))) match {
+        case Success(response) =>
+          implicit val impTrialProjectReport: RootJsonFormat[TrialProjectReport] = jsonFormat4(TrialProjectReport)
+          val trialProjectReport = response.parseJson.convertTo[TrialProjectReport]
+          logger.info(s"Scratched free tier project: ${trialProjectReport.name}")
+          trialProjectReport
+        case Failure(ex) => throw new Exception(s"Unable to scratch free tier project: $projectName", ex)
+      }
+    }
```
davidangb: Do you need to wrap these in `Try`? Without the try, the underlying exception should bubble up. If you would rather keep the code as written so you can control the exception and return the nice error message, I support you.

davidangb (commented):
Could you also mark `createTrialProjects` as deprecated in this PR? Doing it all in one PR would mean we don't need another hash update. I'm fine if you would rather do that in a future PR.

davidangb:
also this is failing unit tests - you should be able to run those locally. I think you need a jsonformat for `UserStatusDetails` - see `reportTrialProjects` for an example.

ghost:
Marked `createTrialProjects` as deprecated. Would you please check the wording and let me know if it could be improved?

davidangb (approved):
:+1: looks good. Deprecation wording is fine!

PR #146, opened 2018-06-26
retry claimGPAlloc billing project if errored

aweng98:
Do not fail automation tests when error occurs in GPAlloc claiming new billing project. Retries up to 3 times.

https://broadinstitute.atlassian.net/browse/GAWB-3637

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
@@ -132,26 +132,28 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp wi
         case Some(project) =>
           //the Rawls endpoint to register a precreated project needs to be called by a Rawls admin
           //but it also takes the new owner's UserInfo in order to create the resource as them in Sam
-          val adminToken = UserPool.chooseAdmin.makeAuthToken()
+          val admin = UserPool.chooseAdmin
+          val adminToken = admin.makeAuthToken()
           val newOwnerUserInfo = UserInfo(OAuth2BearerToken(newOwnerToken().value), WorkbenchUserId("0"), WorkbenchEmail(newOwnerEmail), 3600)
           try {
             Rawls.admin.claimProject(project.projectName, project.cromwellAuthBucketUrl, newOwnerUserInfo)(adminToken)
-            if (ownerEmails.nonEmpty) {
-              addMembersToBillingProject(project.projectName, ownerEmails, BillingProjectRole.Owner)(newOwnerToken())
-            }
-            if (userEmails.nonEmpty) {
-              addMembersToBillingProject(project.projectName, userEmails, BillingProjectRole.User)(newOwnerToken())
-            }
-            ClaimedProject(project.projectName, gpAlloced = true)
           } catch {
             case e: Exception =>
               // Rawls claim project request sometimes fail
               // e.g. of error "Cannot create billing project [gpalloc-qa-master-2z4jdey] in database because it already exists"
-              logger.warn(s"ERROR occurred in claimGPAllocProject. Release unusable GPAlloc billing project ${project.projectName}.")
-              releaseGPAllocProject(project.projectName, newOwnerEmail)(newOwnerToken)
+              logger.warn(s"ERROR occurred in claimGPAllocProject. Release unusable billing project ${project.projectName}.")
+              Rawls.admin.releaseProject(project.projectName, newOwnerUserInfo)(adminToken)
+              GPAlloc.projects.releaseProject(project.projectName)(newOwnerToken())
               Thread sleep Random.nextInt(30000)
               throw e
           }
+          if (ownerEmails.nonEmpty) {
+            addMembersToBillingProject(project.projectName, ownerEmails, BillingProjectRole.Owner)(newOwnerToken())
+          }
+          if (userEmails.nonEmpty) {
+            addMembersToBillingProject(project.projectName, userEmails, BillingProjectRole.User)(newOwnerToken())
+          }
+          ClaimedProject(project.projectName, gpAlloced = true)
         case _ =>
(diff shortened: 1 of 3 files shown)


helgridly:
The error you're seeing here shows up when rawls+sam don't correctly clean up a project that's been released. This change doesn't fix the underlying error, it just hides it (and also hides all other possible errors). IMO we should fix the underlying issue and leave this here so we can tell if it's been fixed. Chatting with @dvoet there's a fix to ~Sam~ Rawls that might help, which I'll quickly make and link here.

helgridly:
proposed alternative (still in talks with doge) https://github.com/broadinstitute/rawls/pull/954

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -104,6 +109,9 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp wi
     claimGPAllocProject(newOwnerCreds.email, ownerEmails, userEmails)(newOwnerCreds.makeAuthToken _)
   }
 
+  implicit val system = ActorSystem()
+  implicit val ec: ExecutionContextExecutor = system.dispatcher
```
rtitle: Minor: you can also do this to get access to the implicit ExecutionContext:
```
import system.dispatcher
```

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -117,22 +125,39 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp wi
...
+          }
+        case _ =>
+          logger.warn("claimGPAllocProject got no project back from GPAlloc. Falling back to making a brand new one...")
+          val billingProjectName = createNewBillingProject("billingproj", ownerEmails, userEmails)(newOwnerToken())
+          ClaimedProject(billingProjectName, gpAlloced = false)
+      }
+    }
+
+    Await.result(retryFuture, 15.minutes) match {
+      case Right((_,project)) => project
```
rtitle: Oof. I see why this code is needed to convert a `RetryableFuture` to a `Future`. However the `Retry` trait already has an implicit conversion for this:
https://github.com/broadinstitute/workbench-libs/blob/develop/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala#L98

The reason the above conversion isn't working here is that the method is package-private. :( This doesn't seem right to me -- I wonder if this would work:
1. Change `protected[util] implicit def retryableFutureToFuture` to just `protected`
2. Change the type of `retryFuture` to a `Future[ClaimedProject]` (this should force the implicit conversion to kick in)
3. Change the code here to `Await.result(retryFuture, 15.minutes)` (no subsequent match statement)

It's not really a big deal either way but would be nice to reuse the conversion we've already defined.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -117,22 +125,39 @@ trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp wi
...
-
-        addMembersToBillingProject(project.projectName, ownerEmails, BillingProjectRole.Owner)(newOwnerToken())
-        addMembersToBillingProject(project.projectName, userEmails, BillingProjectRole.User)(newOwnerToken())
-
-        ClaimedProject(project.projectName, gpAlloced = true)
-      case _ =>
-        logger.warn("claimGPAllocProject got no project back from GPAlloc. Falling back to making a brand new one...")
-        val billingProjectName = createNewBillingProject("billingproj", ownerEmails, userEmails)(newOwnerToken())
-        ClaimedProject(billingProjectName, gpAlloced = false)
+    val retryFuture: RetryableFuture[ClaimedProject] = retry[ClaimedProject](failureLogMessage = "Retry claim GPAlloc billing project") { () =>
```
rtitle, line 129: So just summarizing, we are doing 3 main things here:
1. Request project from gpalloc
2. Register project in Rawls
3. Call Orch N times to add users to the project

If ANY of these fail, all 3 will be retried. Furthermore, if #1 succeeds but #2 or #3 fail, we will clean up created resources in gpalloc/Rawls and still retry all 3 things.

If #1 succeeds but doesn't return a project, we will create a brand new project and NOT retry.

This all seems reasonable, it just helps me to type it out. Overall 👍 from me.

rtitle:
Oh I just saw @helgridly's comment above -- I agree it is better to fix the underlying error (if known) than just adding retries. My comments above are about the retry logic.

rtitle (approved):

dvoet (approved):

helgridly:
Can someone explain to me why this got resurrected? It masks underlying issues. The failures that cause this are real, we shouldn't just be handwaving them away.

rtitle:
It seems like we're getting "bad" projects from gpalloc somehow, causing a lot of test failures. This PR is a bandaid in that it doesn't fix the underlying issue, it just retries the call to gpalloc.

Here are more details -- I see the following sequence of events:
1. A test successfully requests `gpalloc-qa-master-2z4jdey` from gpalloc
2. Rawls registration times out because the project has no billing (e.g. it goes into a retry loop with Google 403 errors and eventually times out)
3. Test retries call to Rawls with the same project: this time it immediately fails with `Cannot create billing project [gpalloc-qa-master-2z4jdey] in database because it already exists` because the project was partially registered in step 2.
4. Test fails, and does not release project because the above exception is not handled
5. gpalloc timer catches the abandoned project 2 hours later
6. gpalloc successfully scrubs it (which SHOULD reset billing!)
7. goto 1

The mystery is why billing doesn't appear to be set for the project even after step 6. I believe @dvoet is opening a Google support ticket for this. I was also reviewing GPAlloc code for potential race conditions, but nothing stands out to me yet.

Because we don't understand the root cause and this code only affects tests, I think it might be good to merge this PR to hopefully stabilize things. Anyway, we can revisit next week - maybe fresh eyes will help.

PR #157, opened 2018-08-10
add googleSubjectId as a separate field

Qi77Qi:
add googleSubjectId to `WorkbenchUser`

**PR checklist**

In all cases:

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -14,14 +14,14 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
 ## Bootstrapping
 
 1. Check [CONTRIBUTING.md](CONTRIBUTING.md) to decide whether or not you need a major or minor version bump and update `Settings.scala` if so
-1. `sbt +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
 1. Search output for a line like `published workbench-service-test_2.12 to https://broadinstitute.jfrog.io/broadinstitute/libs-release-local;build.timestamp=1517520351/org/broadinstitute/dsde/workbench/workbench-service-test_2.12/0.1-99285a4-SNAP/workbench-service-test_2.12-0.1-99285a4-SNAP.jar` and copy the `0.1-99285a4-SNAP` part
 1. Update versions in dependent projects as needed
 
 ## Development
 
 1. Write code
-1. `sbt +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` if you want to publish a specific project (Note: **`publish-local`**)
 
 ## Making Intermediate Commits
 
@@ -32,7 +32,7 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
 ## Sharing Branch Artifacts via Artifactory
 
 1. Commit changes to branch (push optional)
-1. `sbt +publish -Dproject.isSnapshot=true`
+1. `sbt +publish -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish -Dproject.isSnapshot=true`
 1. Search output for a line like `published workbench-service-test_2.12 to https://broadinstitute.jfrog.io/broadinstitute/libs-release-local;build.timestamp=1517520351/org/broadinstitute/dsde/workbench/workbench-service-test_2.12/0.1-99285a4-SNAP/workbench-service-test_2.12-0.1-99285a4-SNAP.jar` and copy the `0.1-99285a4-SNAP` part
 1. Share version # with all your friends
 
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -14,14 +14,14 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
 ## Bootstrapping
 
 1. Check [CONTRIBUTING.md](CONTRIBUTING.md) to decide whether or not you need a major or minor version bump and update `Settings.scala` if so
-1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "project <project_name>" +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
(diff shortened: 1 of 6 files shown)


model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -20,12 +21,13 @@ sealed trait WorkbenchSubject
 
 case class WorkbenchEmail(value: String) extends ValueObject
 
-case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchEmail)
-case class WorkbenchUserId(value: String) extends WorkbenchSubject with ValueObject
+final case class GoogleSubjectId(value: String) extends WorkbenchSubject with ValueObject
+final case class WorkbenchUserId(value: String) extends ValueObject
+final case class WorkbenchUser(googleSubjectId: GoogleSubjectId, email: WorkbenchEmail)
```
dvoet: Why doesn't WorbenchUser have WorkbenchUserId? And googleSubjectId should be an Option, no?

GETTING_STARTED.md
```diff
@@ -14,14 +14,14 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
...
-1. `sbt +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
 1. Search output for a line like `published workbench-service-test_2.12 to https://broadinstitute.jfrog.io/broadinstitute/libs-release-local;build.timestamp=1517520351/org/broadinstitute/dsde/workbench/workbench-service-test_2.12/0.1-99285a4-SNAP/workbench-service-test_2.12-0.1-99285a4-SNAP.jar` and copy the `0.1-99285a4-SNAP` part
 1. Update versions in dependent projects as needed
 
 ## Development
 
 1. Write code
-1. `sbt +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` if you want to publish a specific project (Note: **`publish-local`**)
```
dvoet: spelling "porject"

GETTING_STARTED.md
```diff
@@ -32,7 +32,7 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
 ## Sharing Branch Artifacts via Artifactory
 
 1. Commit changes to branch (push optional)
-1. `sbt +publish -Dproject.isSnapshot=true`
+1. `sbt +publish -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish -Dproject.isSnapshot=true`
```
dvoet: spelling "porject"

GETTING_STARTED.md
```diff
@@ -14,14 +14,14 @@ export ARTIFACTORY_PASSWORD=$(docker run -v $HOME:/root --rm broadinstitute/dsde
 ## Bootstrapping
 
 1. Check [CONTRIBUTING.md](CONTRIBUTING.md) to decide whether or not you need a major or minor version bump and update `Settings.scala` if so
-1. `sbt +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
+1. `sbt +publish-local -Dproject.isSnapshot=true`, or `sbt "porject <project_name>" +publish-local -Dproject.isSnapshot=true` (Note: **`publish-local`**)
```
dvoet: spelling "porject"

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
Qi77Qi: @dvoet `WorkbenchUser` is used in lots of existing api requests, where uuid doesn't exist. I added `LdapUser` to represent what we can get back from ldap, where `googleSubjectId` is optional and `workbenchUserId: WorkbenchUserId` is there too.

dvoet:
Thinking about this a little further, I don't think this should change so much. All that is changing is that we are adding an Optional googleSubjectId. WorkbenchUserId and its usages should remain the same. We may be changing the _values_ of WorkbenchUserId but usage and meaning are not changing.

We _should_ add a googleSubjectId which is _not_ a WorkbenchSubject. Pets should still be in terms of  WorkbenchUserId. Pets are specific to the user which is identified by WorkbenchUserId.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
dvoet: I think this should look like what you have for LdapUser and we don't need LdapUser.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/UserInfo.scala
```diff
@@ -2,4 +2,5 @@ package org.broadinstitute.dsde.workbench.model
 
 import akka.http.scaladsl.model.headers.OAuth2BearerToken
 
-case class UserInfo(accessToken: OAuth2BearerToken, userId: WorkbenchUserId, userEmail: WorkbenchEmail, tokenExpiresIn: Long)
+case class UserInfo(accessToken: OAuth2BearerToken, userId: WorkbenchUserId, googleSubjectId: Option[GoogleSubjectId], userEmail: WorkbenchEmail, tokenExpiresIn: Long)
```
gpolumbo-broad: This class is probably used throughout Workbench.  Changing it might have broad implications.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -11,17 +11,18 @@ object WorkbenchIdentityJsonSupport {
...
 
 sealed trait WorkbenchSubject
 
-case class WorkbenchEmail(value: String) extends ValueObject
-
-case class WorkbenchUser(id: WorkbenchUserId, email: WorkbenchEmail)
-case class WorkbenchUserId(value: String) extends WorkbenchSubject with ValueObject
+final case class WorkbenchEmail(value: String) extends ValueObject
+final case class GoogleSubjectId(value: String) extends ValueObject
+final case class WorkbenchUser(id: WorkbenchUserId, googleSubjectId: Option[GoogleSubjectId], email: WorkbenchEmail)
```
gpolumbo-broad, line 24: I'm similarly worried about the impact of changing the signature of this class.

gpolumbo-broad (commented):
I may be making an invalid assumption and this may just be what needs to happen but I'm concerned about the ripple effect these changes may cause throughout workbench.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
dvoet: for backwards compatibility, will it work to move googleSubjectId to the end of the parameter list and give it a default of None?

Qi77Qi: that's a good point. I'll do that

PR #166, opened 2018-09-20
Added `GoogleProjectDAO.isProjectActive`

rtitle:
This will be used by Leo for zombie cluster detection. 

Example: free trial project is created; user creates a cluster; free trial expires; Leo retains reference to zombie cluster which no longer exists in Google.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleProjectDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleProjectDAO.scala
@@ -1,13 +1,14 @@
 package org.broadinstitute.dsde.workbench.google
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.StatusCodes
+import com.google.api.client.http.HttpResponseException
 import com.google.api.services.cloudresourcemanager.CloudResourceManager
 import com.google.api.services.cloudresourcemanager.model.{Operation, Project}
 import com.google.api.services.compute.ComputeScopes
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes._
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes.GoogleCredentialMode
 import org.broadinstitute.dsde.workbench.metrics.GoogleInstrumentedService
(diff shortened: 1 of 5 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleProjectDAO.scala
```diff
@@ -39,4 +45,32 @@ class HttpGoogleProjectDAO(appName: String,
     })
   }
 
+  override def isProjectActive(projectName: String): Future[Boolean] = {
```
rtitle, line 48: I wasn't quite sure if `isProjectActive` and `isBillingActive` should be folded up into the same method. They are separate Google API calls which is why I made them separate.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleProjectDAO.scala
```diff
@@ -21,10 +23,14 @@ class HttpGoogleProjectDAO(appName: String,
 
   override implicit val service = GoogleInstrumentedService.Projects
 
-  private lazy val cloudResManager = {
+  private def cloudResManager = {
```
kyuksel, line 26: Why the `lazy val` => `def` change?

kyuksel (commented):
Looks good. A question inline. 

> I wasn't quite sure if `isProjectActive` and `isBillingActive` should be folded up into the same method. They are separate Google API calls which is why I made them separate.

I wouldn't fold them up into the same method, for clarity and modularity.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleProjectDAO.scala
rtitle: Because I've been burned by `lazy val` before. `lazy val` in this context means the objects may be used concurrently by multiple request threads. `def` means a new instance will be created per use. I don't _think_ it's an issue with the Google APIs (they are thread safe); I just changed them to `def` as a good habit. :)

kyuksel (approved):

PR #168, opened 2018-09-30
dependency lib and sbt version upgrades

Qi77Qi:
* upgrade cats-core to 1.4.0 in util lib 
* add some useful scalac options
* add bloop so that we can compile/test much faster.
Do `sbt bloopInstall`, and then you can do things like `bloop test workbenchMetrics`

**PR checklist**

In all cases:

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -33,7 +33,7 @@ object Settings {
   )
 
   val commonCrossCompileSettings = Seq(
-    crossScalaVersions := List("2.11.8", "2.12.3")
+    crossScalaVersions := List("2.11.8", "2.12.7")
   )
 
   //sbt assembly settings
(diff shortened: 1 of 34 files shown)


project/Settings.scala
```diff
@@ -46,15 +46,15 @@ object Settings {
...
+    scalaVersion  := "2.12.7",
     resolvers ++= commonResolvers,
     scalacOptions ++= commonCompilerSettings
   )
 
   val utilSettings = commonSettings ++ List(
     name := "workbench-util",
     libraryDependencies ++= utilDependencies,
-    version := createVersion("0.3")
+    version := createVersion("0.4")
```
dvoet, line 88: I think all modules except model need a version bump because they all depend on util

dvoet (changes requested):

project/Settings.scala
Qi77Qi: errrrrrrrrr

dvoet (approved):

marctalbott (approved):

PR #172, opened 2018-10-16
change sam function listResourcePolicies

marctalbott:
**PR checklist**

In all cases:

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/workbench/model/SamModel.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/workbench/model/SamModel.scala
@@ -1,0 +1,15 @@
+package org.broadinstitute.dsde.workbench.model
+
+import spray.json.DefaultJsonProtocol
+
+object SamJsonSupport {
+  import DefaultJsonProtocol._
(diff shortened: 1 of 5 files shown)


model/src/main/scala/org/broadinstitute/dsde/workbench/model/SamModel.scala
```diff
@@ -0,0 +1,15 @@
...
+object SamJsonSupport {
+  import DefaultJsonProtocol._
+  import org.broadinstitute.dsde.workbench.model.WorkbenchIdentityJsonSupport._
+
+  implicit val AccessPolicyMembershipFormat = jsonFormat3(AccessPolicyMembership.apply)
+
+  implicit val AccessPolicyResponseEntryFormat = jsonFormat3(AccessPolicyResponseEntry.apply)
+}
+
+case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[String], roles: Set[String])
```
Qi77Qi: just fyi, you can prefix case class with `final`, so that nothing can extend from these case classes, which is probably what we want

dvoet (approved):

gpolumbo-broad (approved):

Qi77Qi (approved):

PR #173, opened 2018-10-19
add googleFirestore support

Qi77Qi:
add googlefirestore support
https://broadinstitute.atlassian.net/browse/GAWB-3859

Can't seem to use https://typelevel.org/cats-tagless/ for some reason (sbt hangs when publishing jar). Otherwise, futureFirestore can just be 
```
ioFirestore(db).mapK{
    new ~>[IO, Future]{
      override def apply[A](
        fa: IO[A]
      ): Future[A] = fa.unsafeToFuture()
    }
  }
```

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
@@ -1,0 +1,16 @@
+package org.broadinstitute.dsde.workbench.google
+import java.time.Instant
+
+import com.google.cloud.firestore.{DocumentSnapshot, Firestore, Transaction}
+
+import scala.language.higherKinds
+
+trait GoogleFirestoreOps[F[_]] {
+  def set(collectionName: CollectionName, document: Document, dataMap: Map[String, Any]): F[Instant]
+  def get(collectionName: CollectionName, document: Document): F[DocumentSnapshot]
+  def transaction[A](ops: (Firestore, Transaction) => A): F[A]
+}
+
+final case class CollectionName(asString: String) extends AnyVal
+final case class Document(asString: String) extends AnyVal
+final case class FieldKey(asString: String) extends AnyVal
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
@@ -8,7 +8,7 @@ import scala.language.higherKinds
 trait GoogleFirestoreOps[F[_]] {
   def set(collectionName: CollectionName, document: Document, dataMap: Map[String, Any]): F[Instant]
   def get(collectionName: CollectionName, document: Document): F[DocumentSnapshot]
-  def transaction[A](ops: (Firestore, Transaction) => A): F[A]
+  def transaction[A](ops: (Firestore, Transaction) => F[A]): F[A]
 }
 
 final case class CollectionName(asString: String) extends AnyVal
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOpsInterpreters.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOpsInterpreters.scala
@@ -11,75 +11,90 @@ import com.google.cloud.firestore._
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
 
-
-object GoogleFirestoreOpsInterpreters{
-  def ioFirestore(db: Firestore)(implicit ec: ExecutionContext): GoogleFirestoreOps[IO] = new GoogleFirestoreOps[IO] {
-    override def set(collectionName: CollectionName, document: Document, dataMap: Map[String, Any]): IO[Instant] = {
-      val docRef = db.collection(collectionName.asString).document(document.asString)
-      IO.async[WriteResult]{
-        cb =>
-          ApiFutures.addCallback(
-            docRef.set(dataMap.asJava),
-            callBack(cb),
-            executor
-          )
-      }.map(x => Instant.ofEpochSecond(x.getUpdateTime.getSeconds))
+object GoogleFirestoreOpsInterpreters {
+  def ioFirestore(db: Firestore)(
+      implicit ec: ExecutionContext
+  ): GoogleFirestoreOps[IO] = new GoogleFirestoreOps[IO] {
+    override def set(
+        collectionName: CollectionName,
+        document: Document,
+        dataMap: Map[String, Any]): IO[Instant] = {
+      val docRef =
+        db.collection(collectionName.asString).document(document.asString)
+      IO.async[WriteResult] { cb =>
+          ApiFutures
+            .addCallback(docRef.set(dataMap.asJava), callBack(cb), executor)
+        }
+        .map(x => Instant.ofEpochSecond(x.getUpdateTime.getSeconds))
     }
 
-    override def get(collectionName: CollectionName, document: Document): IO[DocumentSnapshot] = {
-      IO.async[DocumentSnapshot]{
-        cb =>
-          ApiFutures.addCallback(
-            db.collection(collectionName.asString).document(document.asString).get(),
-            callBack(cb),
-            executor
-          )
+    override def get(collectionName: CollectionName, document: Document): IO[DocumentSnapshot] =
+      IO.async[DocumentSnapshot] { cb =>
+        ApiFutures.addCallback(
+          db.collection(collectionName.asString)
+            .document(document.asString)
+            .get(),
+          callBack(cb),
+          executor
+        )
       }
-    }
 
-    def transaction[A](ops: (Firestore, Transaction) => IO[A]): IO[A] = {
-      IO.async[A]{
-        cb =>
-          ApiFutures.addCallback(
-            db.runTransaction(new Transaction.Function[A]{
-              override def updateCallback(transaction: Transaction): A = ops(db, transaction).unsafeRunSync()
-            }),
-            callBack(cb),
-            executor
-          )
+    def transaction[A](ops: (Firestore, Transaction) => IO[A]): IO[A] =
+      IO.async[A] { cb =>
+        ApiFutures.addCallback(
+          db.runTransaction(new Transaction.Function[A] {
+            override def updateCallback(transaction: Transaction): A =
+              ops(db, transaction).unsafeRunSync()
+          }),
+          callBack(cb),
+          executor
+        )
       }
-    }
 
     private val executor = new Executor {
       override def execute(command: Runnable): Unit = ec.execute(command)
     }
   }
 
-  def futureFirestore(db: Firestore)(implicit ec: ExecutionContext): GoogleFirestoreOps[Future] = new GoogleFirestoreOps[Future] {
+  def futureFirestore(db: Firestore)(
+      implicit ec: ExecutionContext
+  ): GoogleFirestoreOps[Future] = new GoogleFirestoreOps[Future] {
     val iofs = ioFirestore(db)
-    override def set(collectionName: CollectionName,
-                     document: Document,
-                     dataMap: Map[String, Any]): Future[Instant] = iofs.set(collectionName, document, dataMap).unsafeToFuture()
-    override def get(collectionName: CollectionName,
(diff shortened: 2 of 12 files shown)


helgridly (commented):
I think this is probably fine, but I don't have much context here -- the link to the JIRA says it's about locking in Sam, and I don't see how this solves that problem. Also, I'd find it helpful to see some examples of how to use this code -- I'd usually look to tests for that, but that's hard in library-land.

Qi77Qi:
@helgridly  The ticket is about adding locking in sam to prevent race condition in a few places (first one is for creating SA), and we're using firestore as the external db for implementing locking. This PR is just for adding a small utility lib for accessing firestore. There's some example of usage https://github.com/broadinstitute/sam/pull/230/files#diff-807992852986fafb6263c0b0c2bc2d3d ...I agree it's probably a good idea to add some example of using this code here too. I'll update this PR.

For testing, I don't see any exisitng google dao code is actually testing against google, so I thought that's intended.

helgridly:
Agreed that we don't have a precedent for unit testing googly-DAOs in wb-libs; that was just me griping, not a request :) Thanks for the link to the other PR

project/Settings.scala
```diff
@@ -103,7 +103,7 @@ object Settings {
   val googleSettings = commonSettings ++ List(
     name := "workbench-google",
     libraryDependencies ++= googleDependencies,
-    version := createVersion("0.17"),
+    version := createVersion("0.18"),
```
dvoet: this is additive only and is backwards compatible, right? Then we don't need a version bump.
https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md#no-version-bump-necessary

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.dsde.workbench.google
+import java.time.Instant
+
+import com.google.cloud.firestore.{DocumentSnapshot, Firestore, Transaction}
+
+import scala.language.higherKinds
+
+trait GoogleFirestoreOps[F[_]] {
+  def set(collectionName: CollectionName, document: Document, dataMap: Map[String, Any]): F[Instant]
```
dvoet, line 9: why does this return an Instant?

dvoet (commented):
How are access credentials passed in? I don't see anything explicit which worries me.

Qi77Qi:
@dvoet , from google's doc, creds are set by `export GOOGLE_APPLICATION_CREDENTIALS=pathToCredJson`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOps.scala
Qi77Qi: it's the `Instant` when the record is being updated

dvoet:
I don't think we set that... we use different credentials for different things so need tighter control. We need to figure out how to pass in credentials programmatically like we do for our other google access daos.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreOpsInterpreters.scala
```diff
@@ -0,0 +1,84 @@
...
+                     document: Document,
+                     dataMap: Map[String, Any]): Future[Instant] = dao.set(collectionName, document, dataMap).unsafeToFuture()
+    override def get(collectionName: CollectionName,
+                     document: Document): Future[DocumentSnapshot] = dao.get(collectionName, document).unsafeToFuture()
+    override def transaction[A](ops: (Firestore, Transaction) => A): Future[A] = {
+      dao.transaction(ops).unsafeToFuture()
+    }
+  }
+
+  def firestore(credential: InputStream): Resource[IO, Firestore] = Resource.make[IO, Firestore](IO(FirestoreOptions
```
dvoet: Instead of an InputStream take a GoogleCredentialMode (see the other google DAOs)

dvoet (changes requested):

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+  private[dsde] def getLockStatus(tx: Transaction, documentReference: DocumentReference)(implicit af: Async[F]): F[LockStatus] =
+    for {
+      nullableDocumentSnapshot <- af.async[DocumentSnapshot] { cb =>
+        ApiFutures.addCallback(
+          tx.get(documentReference),
+          callBack(cb),
+          executor
+        )
+      }
+      status <- Option(nullableDocumentSnapshot).fold[F[LockStatus]](af.pure(Available)) { documentSnapshot =>
```
dvoet: prefer case statement to fold

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+    } yield ()
+
+  def retry[F[_], A](fa: F[A], interval: FiniteDuration, maxRetries: Int)
+                         (implicit sf: Sync[F], timer: Timer[F]): F[A] = {
+    fa.handleErrorWith {
+      case e =>
+        if (maxRetries > 0)
+          timer.sleep(interval) *> retry(fa, interval, maxRetries - 1)
+        else
+          sf.raiseError(new WorkbenchException(s"can't get lock: ${e}"))
```
dvoet: `s"timeout waiting for lock: ${e}"`

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfig, val googleFirestoreOps: GoogleFirestoreOps[F])(implicit ec: ExecutionContext, timer: Timer[F]){
+
+  // This is executor for java client's call back to return to. Hence it should be main implicit execution context
+  val executor = new Executor {
+    override def execute(command: Runnable): Unit = ec.execute(command)
+  }
+
+  def withLock(lock: LockPath)(implicit af: Async[F]): Resource[F, Unit] = Resource.make{
+    retry(acquireLock(lock), config.retryInterval, config.maxRetry)
+  }(_ => releaseLock(lock))
```
dvoet, line 29: does this release the lock even if it was not acquired (e.g. timeout waiting for lock)?

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+    googleFirestoreOps.transaction{
+      (db, tx) =>
+        val docRef = db.collection(lockPath.collectionName.asString).document(lockPath.document.asString)
+        sf.delay(tx.delete(docRef))
+    }
+  }
+}
+
+object DistributedLock {
+  private[dsde] val EXPIRESIN = "expiresIn"
```
dvoet: since this is an absolute time call it expiresAt, expiresIn indicates an interval (e.g. expires in 30 seconds as opposed to expires at 3:30PM)

Qi77Qi:
it should be safe to bump grpc since sam service test runs just fine

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+  private[dsde] val EXPIRESINTIMEUNIT = TimeUnit.MILLISECONDS
+
+  def apply[F[_]](lockPathPrefix: String, config: DistributedLockConfig, googleFirestoreOps: GoogleFirestoreOps[F])(implicit ec: ExecutionContext, timer: Timer[F]): DistributedLock[F] = new DistributedLock(lockPathPrefix: String, config, googleFirestoreOps)
+
+  private[dsde] def setLock[F[_]](tx: Transaction, documentReference: DocumentReference, expiresIn: FiniteDuration)(implicit sf: Sync[F], timer: Timer[F]): F[Unit] =
```
dvoet: why are setLock and retry in the companion object? I think we should keep them in the same place as the other functions.

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfig, val googleFirestoreOps: GoogleFirestoreOps[F])(implicit ec: ExecutionContext, timer: Timer[F]){
+
+  // This is executor for java client's call back to return to. Hence it should be main implicit execution context
+  val executor = new Executor {
+    override def execute(command: Runnable): Unit = ec.execute(command)
+  }
+
+  def withLock(lock: LockPath)(implicit af: Async[F]): Resource[F, Unit] = Resource.make{
```
dvoet: Sometimes we use a Sync and sometimes Async... can we fix on one? These implicits are also named af and sf which have not meaning to me. Is there a reason they are passed in to each function individually instead of at the instance level like the execution context and timer?

dvoet (changes requested):

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
Qi77Qi: `Async` extends from `Sync`, see https://typelevel.org/cats-effect/typeclasses/. So the difference being I can use `setLock` with some data type that doesn't have `Async` instance (say `State`), vs `withLock` can only be used with data type that has `Async` instance, which can be handy in testing. It's true in app code, we'll probably always use `IO` or some other things, so there's not much difference there.

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
Qi77Qi: For `setLock`, I don't expect user to call `setLock` directly on a `DistributedLock` instance. Yet being in companion object is handy for testing and from console. But I don't mind putting it in `DistribtuedLock` if you feel strongly about it.

For `retry`, it doesn't have much to do with `DistributedLock`. We probably should move it somewhere else if it's worth it. But there're other libraries provides similar functionality. So I think for now, companion object seems the best place for it.

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
Qi77Qi: I don't think it would. The signature for second parameter is `release: A => F[Unit]`, which means there has to be an `A`.

Also
```
Resource.make(IO.raiseError[Int](new Exception))(a => IO(println(a)))
res4: Resource[IO, Int] = Allocate(Map(RaiseError(java.lang.Exception), scala.Function1$$Lambda$35/1529306539@a18649a, 1))

qi-automation@ res4.use(x => IO.unit).unsafeRunSync()
java.lang.Exception
  ammonite.$sess.cmd4$.<init>(cmd4.sc:1)
  ammonite.$sess.cmd4$.<clinit>(cmd4.sc)

```

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
dvoet: maybe look at org.broadinstitute.dsde.workbench.util.Retry

As for setLock I think we should move it. I am trying to get to consistency. The more inconsistencies, the more chance for misunderstandings, the more questions of why this or that, the more general confusion looking at the code.

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
dvoet: I understand all that. But back to consistency. There is no real reason to use Sync. I understand that it is all that is needed for setLock. In some other context it would probably be fine. But it is strictly internal. Making it different raises the question of why which makes it harder to comprehend.

Also I would say just make it implicit at the class level instead of on each function.

dvoet (approved):

PR #148, opened 2018-06-29
reduce end-user oauth scopes

davidangb:
This PR changes the OAuth scopes that are granted for user tokens and service account tokens. Users default to a small set of `openid, email, profile`, while service accounts are unchanged and add `cloud-platform` and `devstorage.full_control` to the previous list. The user scopes are overridable on a per-test basis.

This PR complements broadinstitute/firecloud-ui#1336. That other PR relies on the per-test overrides to request billing permissions for `BillingSpec`, while keeping the smaller set of scopes for all other tests.

**PR checklist**
- [x] For each library you've modified here, decide whether it requires a major, minor, or no version bump. (Click [here](/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md) for guidance)

If you're doing a **major** or **minor** version bump to any libraries:
- [x] Bump the version in `project/Settings.scala` `createVersion()`
- [x] Update `CHANGELOG.md` for those libraries
- [x] I promise I used `@deprecated` instead of deleting code where possible

In all cases:
- [x] Replace the appropriate version hashes in `README.md` and the `CHANGELOG.md` for any libs you changed with `TRAVIS-REPLACE-ME` to auto-update the version string
- [x] Get two thumbsworth of PR review
- [x] Verify all tests go green (CI _and_ coverage tests)
- [ ] Squash commits and **merge to develop**
- [ ] Delete branch after merge

PR #176, opened 2018-10-25
BillingFixtures.claimGPAllocProject uses the right scopes when it needs to create a project

davidangb:
This is potentially fallout from #148, though I don't know why we haven't seen a problem in the interim.

Tests call `BillingFixtures.claimGPAllocProject`.

If gpalloc has no available projects, `claimGPAllocProject` will create one on the fly.

Before this PR, it tried to create a project using the auth token of the current user (the end user as which the test is running). That token is not likely to have billing scope (see #148). This PR explicitly requests a different token that does have billing scope, in order to create the project.

This should address GAWB-3947 - well, it will once I create follow-on PRs to adopt the updated lib into various test repos.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains common classes and utilities for writing tests against Workbench REST web services.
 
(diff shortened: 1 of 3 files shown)


helgridly (approved):
nice!

dvoet (approved):

PR #179, opened 2018-10-30
turn on useful scalac options for 2.12

Qi77Qi:
* Add more linting to the code
* Upgrade 2.11.8 to 2.11.12, so that partial unification can be enabled

cc @rtitle since I removed `appName` https://github.com/broadinstitute/workbench-libs/pull/179/files#diff-c3ecfd86c34a3f78dc35e2761189917eR14 due to fatal error for unused variable.

I feel we probably don't need to release new versions for these changes since they're mostly syntax change. But let me know if you have concerns.

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
@@ -103,7 +103,7 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
                              POA: Boolean = false,
                              NCU: Boolean = false,
                              prefix: String = "")(implicit token: AuthToken): String = {
-      val request = Map(
+      val request: Map[String, Any] = Map(
         "DS" -> DS,
         "NMDS" -> NMDS,
         "NCTRL" -> NCTRL,
@@ -168,7 +168,7 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
       postRequest(apiUrl(s"api/workspaces"), request)
     }
 
-    def clone(originNamespace: String, originName: String, cloneNamespace: String, cloneName: String, authDomain: Set[String] = Set.empty, membersGroupName: String, attributes: Map[String, String])
+    def clone(originNamespace: String, originName: String, cloneNamespace: String, cloneName: String, authDomain: Set[String] = Set.empty)
              (implicit token: AuthToken): Unit = {
       logger.info(s"Copying workspace: $originNamespace/$originName authDomain: $authDomain")
 
@@ -469,10 +469,10 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
       val successfulResponseKeys = Seq("Success", "NoChangeRequired")
 
       response.parseJson.asJsObject.fields.map {
-        case f@x if successfulResponseKeys.contains(f._1) =>
+        case f@_ if successfulResponseKeys.contains(f._1) =>
           logger.info(s"${f._1}: ${f._2.toString()}")
           return
-        case f@y =>
+        case f@_ =>
           logger.error(s"${f._1}: ${f._2.toString()}")
           throw new Exception(s"Unable to $update trial user: $userEmail. Error message: $response")
       }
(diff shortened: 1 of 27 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -327,13 +326,13 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
-        createMethod(SimpleMethod.creationAttributes + ("name"->methodName))
-      methodName
-    }
+//    def createMethod(testname:String, method:Method, numSnapshots: Int = 1)
+//                    (implicit token: AuthToken): String = {
+//      val methodName = uuidWithPrefix(testname)
+//      for (_ <- 1 to numSnapshots)
+//        createMethod(SimpleMethod.creationAttributes + ("name"->methodName))
+//      methodName
+//    }
```
dvoet: why is this gone?

dvoet (changes requested):
it is hard to tell, but are there api changes that require version bumps? But you do need to update the readme, etc. with TRAVIS_REPLACE_ME

marctalbott (commented):
Are there more places that Google DAOs need to be changed? I'm thinking specifically of the DAOs in [this file](https://github.com/broadinstitute/workbench-libs/blob/develop/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/dao/Google.scala), but perhaps there are other places as well?

Qi77Qi:
@marctalbott  I added back `appName`, so it should be good now

helgridly:
A minor bump is fine here. I'll update the contributing doc to be clear about this; I only intended major to be for "requires significant refactoring".

util/src/main/scala/org/broadinstitute/dsde/workbench/util/ExecutionContexts.scala
```diff
@@ -0,0 +1,33 @@
...
+    * For more info: https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c
+    */
+  def cachedThreadPool[F[_]](
+                              implicit sf: Sync[F]
+                            ): Resource[F, ExecutionContext] = {
+    val alloc = sf.delay(Executors.newCachedThreadPool)
+    val free  = (es: ExecutorService) => sf.delay(es.shutdown())
+    Resource.make(alloc)(free).map(ExecutionContext.fromExecutor)
+  }
+}
```
dvoet, line 33: any reason you can make one function:
`def threadPool[F[_]](executorService: =>ExecutorService)`

dvoet (approved):
1 little thing

util/src/main/scala/org/broadinstitute/dsde/workbench/util/ExecutionContexts.scala
Qi77Qi: I think the reason why we shouldn't pass in a `ExecutorService` is because if we do that, we have no way of preventing user from using this passed in executorservice outside the resource, and if user does that, they'll get exceptions

helgridly (approved):

util/src/main/scala/org/broadinstitute/dsde/workbench/util/ExecutionContexts.scala
```diff
@@ -0,0 +1,33 @@
+package org.broadinstitute.dsde.workbench.util
+import java.util.concurrent.{ExecutorService, Executors}
+
+import cats.effect.{Resource, Sync}
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+
+
+object ExecutionContexts {
```
rtitle, line 9: Cool! cc @breilly2 as we were talking about cats-effect patterns in Avram

rtitle:
Looks good to me, though it's already merged :)

Qi77Qi:
@rtitle  I originally requested your review since I removed something you added, but I ended up reverting that, so I figured I don't need to bother you :)

util/src/main/scala/org/broadinstitute/dsde/workbench/util/ExecutionContexts.scala
Qi77Qi: @breilly2  make sure to turn on [this](https://github.com/broadinstitute/workbench-libs/pull/179/files#diff-6c323177457b467c4590f2132ba7f87aR34) if you guys decide to adopt cats-effect. It'll save lots of time trying to debug complication error

PR #181, opened 2018-11-02
policy sync endpoints for sam automation tests

marctalbott:
**PR checklist**

In all cases:


model/src/main/scala/org/broadinstitute/dsde/workbench/model/SamModel.scala
```diff
@@ -9,7 +9,11 @@ object SamJsonSupport {
...
 
   implicit val AccessPolicyResponseEntryFormat = jsonFormat3(AccessPolicyResponseEntry.apply)
+
+  implicit val CreateResourceRequestFormat = jsonFormat3(CreateResourceRequest.apply)
 }
 
-final case class AccessPolicyMembership(memberEmails: Set[WorkbenchEmail], actions: Set[String], roles: Set[String])
-final case class AccessPolicyResponseEntry(policyName: String, policy: AccessPolicyMembership, email: WorkbenchEmail)
\ No newline at end of file
+final case class AccessPolicyMembership(memberEmails: Set[String], actions: Set[String], roles: Set[String])
```
helgridly: aww really? this makes it a breaking change

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Sam.scala
```diff
@@ -108,12 +108,12 @@ object Sam extends Sam {
       deleteRequest(url + s"api/groups/v1/$groupName")
     }
 
-    def addUserToPolicy(groupName: String, policyName: String, memberEmail: String)(implicit token: AuthToken): Unit = {
+    def addUserToManagedGroupPolicy(groupName: String, policyName: String, memberEmail: String)(implicit token: AuthToken): Unit = {
```
helgridly: if you want to rename these functions, go ahead, but leave the old ones there + mark them as deprecated + make them call the new ones

helgridly (changes requested):
i think there are a couple minor things you can do to make this an entirely nonbreaking change

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Sam.scala
Qi77Qi: if we know this function is not used by any code other than sam, I'd rather just rename them to reduce maintenance overhead.

dvoet (approved):

model/CHANGELOG.md
```diff
@@ -2,9 +2,17 @@
 
 This file documents changes to the `workbench-model` library, including notes on how to upgrade to new versions.
 
+## 0.13
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.13-TRAVIS-REPLACE-ME"`
+
+### Changed
+
+- Remove duplicated `SamModel`
```
Qi77Qi, line 11: I think you'll need to bump other modules that depend on `model`, which is all modules I think.

helgridly: yes (sadly)

marctalbott: Just to be clear, you mean that all the other libs need minor version bumps as well, is that correct?

helgridly: yes

helgridly (approved):

PR #188, opened 2018-11-21
add googleStorage lib wrapper around new google storage client

Qi77Qi:
wip to gather some thoughts.
Motivation
1. Current google library is no longer in active development. https://github.com/googleapis/google-api-java-client#library-maintenance
2. I'd like a google lib exposes `IO` interface instead of only `Future`
3. New storage library has a [java emulator](https://github.com/googleapis/google-cloud-java/blob/master/TESTING.md#testing-code-that-uses-storage). It doesn't cover all features though. 
4. rename the `Ops` class to `Alg` just so it's more consistent with most tagless final implementations

Only implemented methods needed in https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/google/GoogleKeyCache.scala

TODO:
* check out logging for grpc calls
* https://github.com/broadinstitute/workbench-libs/pull/188#discussion_r237731810

**PR checklist**

In all cases:


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreAlg.scala
```diff
@@ -5,7 +5,7 @@ import com.google.cloud.firestore.{DocumentSnapshot, Firestore, Transaction}
 
 import scala.language.higherKinds
 
-trait GoogleFirestoreOps[F[_]] {
+trait GoogleFirestoreAlg[F[_]] {
```
dvoet: What is Alg short for? Can you avoid the abbreviation?

I am guessing it is Algebra but spelling that out is probably more confusing.. How about `GoogleFirestoreService`?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageAlg.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.google.cloud.storage.BucketInfo.LifecycleRule
+import fs2.Stream
+import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject}
+
+import scala.language.higherKinds
+
+trait GoogleStorageAlg[F[_]] {
+  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String): Stream[F, GcsObjectName]
+
+  // not memory safe. Use listObjectsWithPrefix if you're worried about OOM
```
dvoet: put this in scaladoc

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+import scala.concurrent.ExecutionContext
+
+// AsyncFlatSpec currently doesn't work with scalacheck's forAll. It'll be supported in scalatest 3
+class GoogleStorageInterpreterSpec extends AsyncFlatSpec with Matchers {
+  "ioStorage storeObject" should "be able to upload an object" in {
+    val bucketName = genGcsBucketName.sample.get
+    val objectName = genGcsBlobName.sample.get
+    val objectBody = genGcsObjectBody.sample.get
+    val res = localStorage.storeObject(bucketName, objectName, objectBody, objectType).attempt.map(x => x.isRight shouldBe(true))
+    res.unsafeToFuture()
```
dvoet: I have been getting grumpy about having to call `res.unsafeToFuture` at the end of every test. It is so easy to forget and not even notice. Is there some way we can get scalatest to know about IO? Or maybe we just have a utility function:
```
def ioTest[T](test: => IO[T]) = test.unsafeToFuture()
```
then tests can look like
```
"ioStorage storeObject" should "be able to upload an object" in ioTest {
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
```diff
@@ -0,0 +1,89 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+object GoogleStorageInterpreters {
+  def ioStorage(db: Storage, blockingEc: ExecutionContext, maxPageSize: Long = 1000)(
+    implicit cs: ContextShift[IO]
+  ): GoogleStorageAlg[IO] = new GoogleStorageAlg[IO] {
+    override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String): Stream[IO, GcsObjectName] = for {
+        pages <- Stream.eval(cs.evalOn(blockingEc)(IO(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize), BlobListOption.currentDirectory()))))
```
dvoet: I don't think this handles queries with results that span pages. This only gets the first page and never subsequent pages.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+import org.broadinstitute.dsde.workbench.google.Generators._
+import org.broadinstitute.dsde.workbench.google.GoogleStorageInterpreterSpec._
+import org.broadinstitute.dsde.workbench.google.GoogleStorageInterpreters._
+import org.scalacheck.Gen
+import org.scalatest.{AsyncFlatSpec, Matchers}
+
+import scala.concurrent.ExecutionContext
+
+// AsyncFlatSpec currently doesn't work with scalacheck's forAll. It'll be supported in scalatest 3
+class GoogleStorageInterpreterSpec extends AsyncFlatSpec with Matchers {
```
dvoet: add a test where `listObjectsWithPrefix` has more results than fit on a single page, suggest make `maxPageSize` small

dvoet (changes requested):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreAlg.scala
dvoet: clarification: Algebra is not a confusing term, it is quite precise. And while this may be an algebra, I don't think it conveys what we want. 

This is the thing that knows how to talk to Firestore. In the past we have used the term data access object (DAO). I don't really care if we want to use a different term but it should be something that indicates what it does instead of what it is.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
Qi77Qi: from doc
```
  /**
   * Returns an iterable that traverses all of the elements of the underlying data source. The data
   * is fetched lazily page by page, where each page may contain multiple elements. A new page is
   * fetched whenever the elements of any particular page are exhausted.
   */
  Iterable<ResourceT> iterateAll();
```

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
Qi77Qi: I don't think scalatest will know about IO since it doesn't depends on cats-effect.
I think it'll error if you have `IO[Assertion]` at the end of each test instead of `Future[Assertion]`.
I'll create the ioTest helper if you prefer

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreAlg.scala
Qi77Qi: wouldn't it be more confusing if we use some custom name?
With `Alg`, we can refer people to https://typelevel.org/cats-tagless/, and it's easier for them to follow the pattern.
Maybe I should link https://typelevel.org/cats-tagless/ if that helps? most blog posts will refer this to `Alg` or `Algebra` I think. I'm fine with `Algebra` too if that's better

marctalbott (approved):
Looks fine, although I can't say I fully understand the fs2 streams and all the functions that go along with them like `compile`, `drain`, `foldMonoid`, etc.

Qi77Qi:
Since there has been many times code compiles fine in 2.12 but not in 2.11. I feel we should just drop 2.11 for modules that no projects depends on.

There's some change related to travis
1. replace `++$TRAVIS_SCALA_VERSION` with `+` so that all sbt commands will cross build based on `crossScalaVersions` in sbt
2.  remove 2.11 in `.travis.yml` since we're already building both versions in one build

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
```diff
@@ -0,0 +1,124 @@
...
+                                     )(
+  implicit cs: ContextShift[IO],
+  timer: Timer[IO]
+) extends GoogleStorageService[IO] {
+  private def retryStorageIO[A](ioa: IO[A]): Stream[IO, A] = Stream.retry(ioa, retryConfig.retryInitialDelay, retryConfig.retryNextDelay, retryConfig.maxAttempts, retryConfig.retryable)
+
+  override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[IO, GcsObjectName] = {
+    for {
+      firstPage <- retryStorageIO(cs.evalOn(blockingEc)(IO(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize.longValue()), BlobListOption.currentDirectory()))))
+      page <- Stream.unfold(firstPage){
```
Qi77Qi: `Stream.unfold` will keep running the effects until `None`

.travis.yml
```diff
@@ -2,7 +2,6 @@ jdk:
 - oraclejdk8
 language: scala
 scala:
-- 2.11.12
 - 2.12.7
```
Qi77Qi: removing `2.11.12` here, but all sbt commands are running with `+`, so it'll cross build respecting `crossScalaVersions` in sbt definition.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    retryStorageIO(cs.evalOn(blockingEc)(IO(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value))))).compile.drain
+  }
+
+  override def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule]): IO[Unit] = {
+    val bucketInfo = BucketInfo.of(bucketName.value)
+      .toBuilder
+      .setLifecycleRules(lifecycleRules.asJava)
+      .build()
+    retryStorageIO(cs.evalOn(blockingEc)(IO(db.update(bucketInfo)))).compile.drain
```
Qi77Qi: `.compile.drain` is for running the stream just for effects.
For instance:
```
qi-sam@ Stream.eval(IO(println(1))) >>  Stream.eval(IO(println(2)))
res3: Stream[IO[x], Unit] = Stream(..)

qi-sam@ res3.compile.drain.unsafeRunSync()
1
2

```

`>>` is the same as `flatMap(_ => secondStream)`. The operator exists for all monads once you import `cats.implicits._`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
```diff
@@ -0,0 +1,124 @@
...
+  override def unsafeListObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): IO[List[GcsObjectName]] = listObjectsWithPrefix(bucketName, objectNamePrefix).compile.toList
+
+  override def unsafeGetObject(bucketName: GcsBucketName, blobName: GcsBlobName): IO[Option[String]] = {
+    val getBlobs = cs.evalOn(blockingEc)(IO(db.get(BlobId.of(bucketName.value, blobName.value)))).map(Option(_))
+
+    val res = for {
+      blob <- OptionT(retryStorageIO(getBlobs))
+      r <- OptionT.liftF(Stream.emits(blob.getContent()).covary[IO].through(text.utf8Decode))
+    } yield r
+    res.value.compile.foldMonoid
```
Qi77Qi: `.compile.foldMonoid` over `Option` monoid instance.
`fold` is for reducing stream down to a value.
`foldMonoid` means reducing the stream down to a pure value with monoid's `+` function

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreters.scala
```diff
@@ -0,0 +1,124 @@
...
+    } yield ()
+  }
+
+  /**
+    * @return true if deleted; false if not found
+    */
+  override def removeObject(bucketName: GcsBucketName, blobName: GcsBlobName): IO[Boolean] = {
+    val deleteObject = cs.evalOn(blockingEc)(IO(db.delete(BlobId.of(bucketName.value, blobName.value))))
+    for {
+      deleted <- retryStorageIO(deleteObject).compile.lastOrError
```
Qi77Qi: `.compile.lastOrError` will raise error if the stream is empty

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleFirestoreInterpreter.scala
```diff
@@ -0,0 +1,85 @@
...
+import com.google.api.core.{ApiFutureCallback, ApiFutures}
+import com.google.auth.oauth2.ServiceAccountCredentials
+import com.google.cloud.firestore._
+import org.broadinstitute.dsde.workbench.google.GoogleFirestoreInterpreter.callBack
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+private[google] class GoogleFirestoreInterpreter[F[_]](db: Firestore)(
+              implicit effect: Effect[F],
```
Qi77Qi: not sure why git doesn't recognize this as a renamed file..but there's no significant change except I changed constraint to `Effect`

Qi77Qi: use `Effect` constraint, because I need to `unsafeRunSync` on the code that's submitted to google's callback api

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+private[google] class GoogleStorageInterpreter[F[_]](db: Storage,
+                                      blockingEc: ExecutionContext,
+                                      retryConfig: RetryConfig
+                                     )(
+  implicit cs: ContextShift[F],
+  timer: Timer[F],
+  asyncF: Async[F]
+) extends GoogleStorageService[F] {
+  private def retryStorageF[A](ioa: F[A]): Stream[F, A] = Stream.retry(ioa, retryConfig.retryInitialDelay, retryConfig.retryNextDelay, retryConfig.maxAttempts, retryConfig.retryable)
```
rtitle: This retry stuff isn't really `Storage`-specific. It'd be cool to move this to a common location in case other Google interpreters want to use the streaming pattern as an alternative to https://github.com/broadinstitute/workbench-libs/blob/develop/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName] = {
+    for {
+      firstPage <- retryStorageF(cs.evalOn(blockingEc)(asyncF.delay(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize.longValue()), BlobListOption.currentDirectory()))))
+      page <- Stream.unfold(firstPage){
+        currentPage =>
+          Option(currentPage).map(p => (p, p.getNextPage))
+      }
+      objects <- Stream.fromIterator[F, Blob](page.iterateAll().iterator().asScala).map{
```
rtitle: It looks like `iterateAll` returns all the elements, _on all the pages_ (e.g. it lazily loads a new page when the current page is exhausted). Since we're already `unfold`ing pages, I wonder if `getValues()` is the method we want to call here, which only operates on the current page.

See: https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/paging/Page.java

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+  asyncF: Async[F]
+) extends GoogleStorageService[F] {
+  private def retryStorageF[A](ioa: F[A]): Stream[F, A] = Stream.retry(ioa, retryConfig.retryInitialDelay, retryConfig.retryNextDelay, retryConfig.maxAttempts, retryConfig.retryable)
+
+  override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName] = {
+    for {
+      firstPage <- retryStorageF(cs.evalOn(blockingEc)(asyncF.delay(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize.longValue()), BlobListOption.currentDirectory()))))
+      page <- Stream.unfold(firstPage){
+        currentPage =>
+          Option(currentPage).map(p => (p, p.getNextPage))
```
rtitle: Does `p.getNextPage()` make a Google API call? If so I'm wondering if it should be retried like the initial call to `list()`.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+      }
+      objects <- Stream.fromIterator[F, Blob](page.iterateAll().iterator().asScala).map{
+        blob => GcsObjectName(blob.getName, Instant.ofEpochMilli(blob.getCreateTime))
+      }
+    } yield objects
+  }
+
+  override def unsafeGetObject(bucketName: GcsBucketName, blobName: GcsBlobName): F[Option[String]] = {
```
rtitle: Should we also define a safe `getObject` which returns the object contents as a `Stream`? Would this return a `Stream[F, Byte]`, or could it be chunked somehow (I'm not remembering what fs2 supports here)?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+    val blobInfo = BlobInfo.newBuilder(bucketName.value, objectName.value).setContentType(objectType).build()
+
+    val storeObject = cs.evalOn(blockingEc)(asyncF.delay(db.create(blobInfo, objectContents)))
+    for {
+      _ <- retryStorageF(storeObject).compile.drain
+    } yield ()
+  }
+
+  override def removeObject(bucketName: GcsBucketName, blobName: GcsBlobName): F[Boolean] = {
+    val deleteObject = cs.evalOn(blockingEc)(asyncF.delay(db.delete(BlobId.of(bucketName.value, blobName.value))))
```
rtitle: This pattern is everywhere, I wonder if a helper function would be useful?
```
  def liftAsync[A](thunk: => A): F[A] = {
    cs.evalOn(blockingEc)(asyncF.delay(thunk))
  }
```
Or is that more confusing? I dunno..

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+object GoogleStorageInterpreter {
+  private val storageRetryableCode = Set(408, 429, 500, 502, 503, 504) //error codes are copied from com.google.cloud.storage.StorageException.RETRYABLE_ERRORS
+
+  val defaultRetryConfig = RetryConfig(
+    2 seconds,
+    x => x * 2,
+    5,
+    {
+      case e: StorageException => if(storageRetryableCode.contains(e.getCode)) true else false
```
rtitle: Nit:
```
case e: StorageException => storageRetryableCode.contains(e.getCode)
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName]
+
+  /**
+    * not memory safe. Use listObjectsWithPrefix if you're worried about OOM
+    */
+  def unsafeListObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000)(implicit sf: Sync[F]): F[List[GcsObjectName]] = listObjectsWithPrefix(bucketName, objectNamePrefix).compile.toList
+
+  def storeObject(bucketName: GcsBucketName, objectName: GcsBlobName, objectContents: Array[Byte], objectType: String): F[Unit]
+
+  def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule]): F[Unit]
```
rtitle: Do we care about using `com.google.cloud.storage._` classes in this interface? Should we replicate everything in `workbench-model`? I'm honestly not sure. Right now it looks like we use the Google classes for `Acl` and `LifecycleRule`, whereas everything else (like `GcsBucketName`, `GcsObjectName`) comes from workbench-model.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+    * @return true if deleted; false if not found
+    */
+  def removeObject(bucketName: GcsBucketName, objectName: GcsBlobName): F[Boolean]
+
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, acl: List[Acl]): F[Unit]
+}
+
+final case class GcsBlobName(value: String) extends AnyVal
```
rtitle: This should maybe go in workbench-model along with the other Gcs model classes.

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLockAlgebra.scala
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.dsde.workbench.util
+
+import cats.effect.Resource
+import org.broadinstitute.dsde.workbench.google.util.LockPath
+
+trait DistributedLockAlgebra[F[_]] {
+  def withLock(lock: LockPath): Resource[F, Unit]
```
rtitle: Cool, nice use of cats-effect `Resource` to implement a lock. So the caller can do something like this:
```
distributedLock.withLock(LockPath(...)).use(_ => doSomething())
```
and it will automatically acquire and release the lock, correct?

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -38,14 +39,14 @@ class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfi
         lockStatus <- getLockStatus(tx, docRef)
         _ <- lockStatus match {
           case Available => setLock(tx, docRef, lock.expiresIn)
-          case Locked => af.raiseError[Unit](new WorkbenchException(s"can't get lock: $lock"))
+          case Locked => asyncF.raiseError[Unit](new WorkbenchException(s"can't get lock: $lock"))
```
rtitle: So it fails fast if the lock can't be acquired, as opposed to blocking (up to some timeout) waiting for the lock to be available?

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
```diff
@@ -72,14 +73,14 @@ class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfi
       data = Map(
```
rtitle, line 78: This data is to implement a ttl on the lock, correct? (Just checking my understanding of this code)

google/src/test/scala/org/broadinstitute/dsde/workbench/google/Generators.scala
```diff
@@ -0,0 +1,19 @@
+package org.broadinstitute.dsde.workbench.google
+
+import java.nio.charset.Charset
+import java.time.Instant
+
+import org.broadinstitute.dsde.workbench.model.google._
+import org.scalacheck.Gen
```
rtitle: Ooh scalacheck - lots of cool things in this PR :)

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+import org.broadinstitute.dsde.workbench.google.GoogleStorageInterpreterSpec._
+import org.scalacheck.Gen
+import org.scalatest.{AsyncFlatSpec, Matchers}
+
+import scala.concurrent.ExecutionContext
+
+// AsyncFlatSpec currently doesn't work with scalacheck's forAll. It'll be supported in scalatest 3
+class GoogleStorageInterpreterSpec extends AsyncFlatSpec with Matchers {
+  "ioStorage storeObject" should "be able to upload an object" in ioAssertion {
+    val bucketName = genGcsBucketName.sample.get
```
rtitle: Interesting - is there a reason you're just taking 1 sample from the generators as opposed to something like:
```
forAll(genGcsBucketName, genGcsBlobName, genGcsObjectBody) { (bucketName, objectName, objectBody) =>
  // your assertion
}
```
which would generate many permutations of data and run the assertion each time? As written I feel like you’re not really using the power of ScalaCheck.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+    } yield {
+      allObjectsWithPrefix.map(_.value) should contain theSameElementsAs (blobNameWithPrefix.map(_.value))
+    }
+  }
+}
+
+object GoogleStorageInterpreterSpec {
+  implicit val cs = IO.contextShift(ExecutionContext.global)
+  implicit val timer = IO.timer(ExecutionContext.global)
+  val db = LocalStorageHelper.getOptions().getService()
```
rtitle: Cool that Google provides this local in-memory implementation for testing.

rtitle (changes requested):
This looks awesome! The tagless final approach and abstraction of the effect type works well here. The use of fs2 for retries and streaming of large data is nice too, I'd love to see that expand to other areas of workbench code.

Once it's finalized, it'd be nice to update Leo to use `GoogleStorageInterpreter` as well as Sam. We do things like create/delete buckets, set acls, and list objects. I think the Leo GCS code is mostly localized [here](https://github.com/DataBiosphere/leonardo/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/BucketHelper.scala).

I left some comments above, take them or leave them. My only real concern is with the implementation of `listObjectsWithPrefix` -- based on my reading of the javadoc it looks like it might return duplicate entries because it's streaming through pages and calling `iterateAll()` on each page. I'm not 100% sure of the behavior though -- may be worth a test with actual buckets.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
Qi77Qi: see the comments on line 14 :)
I'm using `AsyncFlatSpec`. I wish it works with `forAll`

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
Qi77Qi: this function is just for acquiring a lock. `withLock` will actually retry and semantically block

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLock.scala
Qi77Qi: yep...

google/src/main/scala/org/broadinstitute/dsde/workbench/util/DistributedLockAlgebra.scala
Qi77Qi: yep

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: I thought `iterateAll` will retrieve all pages in the beginning too. But I wrote a test https://github.com/broadinstitute/workbench-libs/pull/188/files#diff-405a305f0d21ea56fc302bdd7467022bR87, and with storage emulator, it only returned one page. That's why I added `unfold` to retrieve all pages

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: if we want to be memory safe, I think we'll need to write to a file. Google storage doesn't expose an api that returns a memory safe data type.

I think we should definitely add this at some point, but since this PR is more for supporting https://github.com/broadinstitute/sam/pull/249 at the moment, which is just retrieving a key file. I think we can add it in a different PR.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: I have no idea why I did `if someBoolean true else false` lol

google/src/test/scala/org/broadinstitute/dsde/workbench/google/Generators.scala
Qi77Qi: I was thinking of adding scalacheck generators for wb-model too. I created some in sam repo https://github.com/broadinstitute/sam/blob/develop/src/test/scala/Generator.scala, but some of those probably can be moved to wb-model for sharing

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
Qi77Qi: Yea, it's only for a few google services and they're not full featured

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: I don't think it'll actually save much code by extracting this out. But I added 
```
  def retryGoogleF[F[_]: Timer: RaiseThrowable, A](retryConfig: RetryConfig)(fa: F[A]): Stream[F, A] = Stream.retry[F, A](fa, retryConfig.retryInitialDelay, retryConfig.retryNextDelay, retryConfig.maxAttempts, retryConfig.retryable)

```
in package object.

Not sure if that's better or worth though

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: I kinda prefer to have specific `evalOn` calls since we should be cautious about shifting to blockingEc

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
Qi77Qi: I think exposing google's data types directly gives more options to callers. They're just `String`s, so I think they're fine. Our current model case classes are actually less ideal because google's library api changed. Another reason I don't quite like current google models is there's some json code tied with them, which shouldn't be necessary anymore since new library uses grpc

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
Qi77Qi: I see wb-model as firecloud/wb related model. Say a new service might want to bring in wb-models, but it doesn't care about calling google, and they probably don't care about google related models.

Qi77Qi:
@rtitle  I'm not sure I understand the concern about duplicates completely. I added https://github.com/broadinstitute/workbench-libs/pull/188/files#diff-405a305f0d21ea56fc302bdd7467022bR101, I think duplicate entries are overwritten when storing objects.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
rtitle: Gotcha, makes sense.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
rtitle: Ah ok, should have read the comment :)

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
rtitle: Interesting, ok. Maybe it's worth a test with a real bucket with multiple pages? I'm just worried that if `iterateAll` works as advertised you might end up with duplicate entries by calling `iterateAll` for each page.

For example if you have:
```
page1
  obj1
  obj2
page2
  obj3
  obj4
```
would we get
```
Stream(
  /* iterateAll on page1 */ obj1, obj2, obj3, obj4, 
  /* iterateAll on page2 */ obj1, obj2, obj3, obj4)
```
?

I could be completely misunderstanding though..

rtitle:
^ elaborated in [this comment](https://github.com/broadinstitute/workbench-libs/pull/188#discussion_r237731810)

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
Qi77Qi: oh I see...I'll try this out tomorrow.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreter.scala
```diff
@@ -0,0 +1,119 @@
...
+      .build()
+    retryStorageF(cs.evalOn(blockingEc)(asyncF.delay(db.update(bucketInfo)))).compile.drain
+  }
+}
+
+object GoogleStorageInterpreter {
+  private val storageRetryableCode = Set(408, 429, 500, 502, 503, 504) //error codes are copied from com.google.cloud.storage.StorageException.RETRYABLE_ERRORS
+
+  val defaultRetryConfig = RetryConfig(
+    2 seconds,
```
dvoet: I would start with 1 second and add a jitter of up to another second `org.broadinstitute.dsde.workbench.util.addJitter`

dvoet (commented):
1 more comment on retries then we need to figure out how to get the logs we want. That might not be actual code changes here but we should add it to the scaladoc or something.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+import fs2.Stream
+import com.google.cloud.storage.contrib.nio.testing.LocalStorageHelper
+import org.broadinstitute.dsde.workbench.google.Generators._
+import org.broadinstitute.dsde.workbench.google.GoogleStorageInterpreterSpec._
+import org.scalacheck.Gen
+import org.scalatest.{AsyncFlatSpec, Matchers}
+
+import scala.concurrent.ExecutionContext
+
+// AsyncFlatSpec currently doesn't work with scalacheck's forAll. It'll be supported in scalatest 3
```
danbills: scalatest 3 is out

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+  def unsafeListObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000)(implicit sf: Sync[F]): F[List[GcsObjectName]] = listObjectsWithPrefix(bucketName, objectNamePrefix).compile.toList
+
+  def storeObject(bucketName: GcsBucketName, objectName: GcsBlobName, objectContents: Array[Byte], objectType: String): F[Unit]
+
+  def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule]): F[Unit]
+
+  /**
+    * not memory safe
+    */
+  def unsafeGetObject(bucketName: GcsBucketName, blobName: GcsBlobName): F[Option[String]]
```
danbills: What is an `Object`?  i.e. can `String` b improved?  to me it looks like the contents of the file?  Also what's with the option, i.e. what does it mean when it's `None`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+  */
+trait GoogleStorageService[F[_]] {
+  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName]
+
+  /**
+    * not memory safe. Use listObjectsWithPrefix if you're worried about OOM
+    */
+  def unsafeListObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000)(implicit sf: Sync[F]): F[List[GcsObjectName]] = listObjectsWithPrefix(bucketName, objectNamePrefix).compile.toList
+
+  def storeObject(bucketName: GcsBucketName, objectName: GcsBlobName, objectContents: Array[Byte], objectType: String): F[Unit]
```
danbills: consider using Scodec bitvector instead of `Array[Byte]`?  What about objectType, can `String` be improved?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule]): F[Unit]
+
+  /**
+    * not memory safe
+    */
+  def unsafeGetObject(bucketName: GcsBucketName, blobName: GcsBlobName): F[Option[String]]
+
+  /**
+    * @return true if deleted; false if not found
```
danbills: IMO This should be an enumeration

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
```diff
@@ -0,0 +1,41 @@
...
+    */
+  def removeObject(bucketName: GcsBucketName, objectName: GcsBlobName): F[Boolean]
```
danbills: as above

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
danbills: this `String` could probably be refined via "Refined" library

google/src/main/scala/org/broadinstitute/dsde/workbench/google/util/DistributedLock.scala
```diff
@@ -6,19 +6,23 @@ import java.util.concurrent.{Executor, TimeUnit}
...
+import org.broadinstitute.dsde.workbench.util.DistributedLockAlgebra
 
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 
-class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfig, val googleFirestoreOps: GoogleFirestoreOps[F])(implicit ec: ExecutionContext, timer: Timer[F], af: Async[F]){
-
+class DistributedLock[F[_]](lockPathPrefix: String, config: DistributedLockConfig, val googleFirestoreOps: GoogleFirestoreService[F])(implicit ec: ExecutionContext, timer: Timer[F], asyncF: Async[F]) extends DistributedLockAlgebra[F]{
+  val retryable: Function1[Throwable, Boolean] = {
```
danbills: total nitpick but why not use syntax sugar for `Function1` type declaration `(Throwable) => Boolean`

danbills (commented):
Wondering why this isn't being used yet?  We could use it in Cromwell's DRS impl.

Lots of little questions but I really like the overall approach!

google/src/test/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageInterpreterSpec.scala
Qi77Qi: I realize the comment is a lie...It's really due to `forAll` is expecting a pure value...see https://github.com/rickynils/scalacheck/issues/214

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
Qi77Qi: `Object` here is just a blob. Return type is `String` because I used I piped it through `text.utf8Decode` before `.compile` it into `F`...Since lots of times for small file objects, `String` is probably what we want anyway. `None` means the blob doesn't exist in google.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
Qi77Qi: I don't know scodec well enough, but possibly...This PR is pretty big as is though, so I'll probably leave this as is and we can make future improvements.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleStorageService.scala
Qi77Qi: definitely worth improvements! In an other PR tho :)

google/src/main/scala/org/broadinstitute/dsde/workbench/google/util/DistributedLock.scala
Qi77Qi: no reason other than I happened to be using `Function1`

dvoet (approved):

Qi77Qi:
I'm merging this PR for now but comments are still welcome and I'll address retrospectively

rtitle (commented):
Nice, 👍 . I like implementing it as a separate sub-module.

PR #189, opened 2018-11-27
GAWB-3902: Add IOException as a retry condition

ansingh7115:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.workbench.auth
 
 import java.io.IOException
-import java.net.SocketTimeoutException
 
 import akka.http.scaladsl.model.StatusCodes
 import com.google.api.client.auth.oauth2.TokenResponseException
@@ -36,7 +35,7 @@ trait AuthToken extends LazyLogging {
     Retry.retry(5.seconds, 1.minute)({
 
       val cred = buildCredential()
-      val baseLogMessage =
+      val baseLogMessage = "Details: \n" +
         s"Service Account: ${cred.getServiceAccountId} \n" +
         s"User: ${cred.getServiceAccountUser} \n" +
         s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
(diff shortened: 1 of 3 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
```diff
@@ -31,20 +33,27 @@ trait AuthToken extends LazyLogging {
...
-            s"Service Account: ${cred.getServiceAccountId} \n" +
-            s"User: ${cred.getServiceAccountUser} \n" +
-            s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
-            s"Access Token: ${cred.getAccessToken} \n" +
-            s"Token Expires: in ${cred.getExpiresInSeconds} seconds \n" +
-            s"SA Private Key ID: ${cred.getServiceAccountPrivateKeyId}")
+          logger.error(s"Encountered ${e.getStatusCode} error getting access token." + baseLogMessage)
+          None
+        case f: IOException => {
+          logger.error(s"Error getting access token with error message: ${f.getMessage} \n" + baseLogMessage)
```
dvoet: `logger.error` has a variation that takes a throwable. Always use that instead embedding `f.getMessage` in the message. Otherwise we lose exception information like the stack trace and the exception class.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/auth/AuthToken.scala
```diff
@@ -31,20 +33,27 @@ trait AuthToken extends LazyLogging {
...
     Retry.retry(5.seconds, 1.minute)({
+
       val cred = buildCredential()
+      val baseLogMessage = "Details: \n" +
+        s"Service Account: ${cred.getServiceAccountId} \n" +
+        s"User: ${cred.getServiceAccountUser} \n" +
+        s"Scopes: ${cred.getServiceAccountScopesAsString} \n" +
+        s"Access Token: ${cred.getAccessToken} \n" +
+        s"Token Expires: in ${cred.getExpiresInSeconds} seconds \n" +
+        s"SA Private Key ID: ${cred.getServiceAccountPrivateKeyId}"
```
dvoet, line 44: Extracting this to a val in test code is ok. In production code I would make it a def. The problem is that this is a bunch of needless string concatenation in the success case which can lead to performance issues in high traffic areas.

dvoet (changes requested):

dvoet (approved):

gpolumbo-broad (approved):

PR #201, opened 2019-01-17
Add pubsub

Qi77Qi:
https://broadinstitute.atlassian.net/browse/GAWB-4066

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
@@ -1,8 +1,23 @@
 package org.broadinstitute.dsde.workbench.google2
 
+import cats.effect.{Async, ContextShift, Resource, Timer}
 import fs2.Sink
 import io.circe.Encoder
 
 trait GooglePublisher[F[_]] {
+  /**
+    * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
+    */
   def publish[A: Encoder]: Sink[F, A]
+
+  /**
+    * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
+    */
+  def publishString: Sink[F, String]
+}
+
+object GooglePublisher {
+  def resource[F[_]: Async: Timer: ContextShift, A: Encoder](config: PublisherConfig): Resource[F, GooglePublisher[F]] = for {
+    publisher <- GooglePublisherInterpreter.publisher(config)
+  } yield GooglePublisherInterpreter(publisher, config.retryConfig)
 }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
@@ -1,0 +1,10 @@
+package org.broadinstitute.dsde.workbench.google2
+
+import com.google.pubsub.v1.ProjectTopicName
+import fs2.Sink
+import io.circe.Encoder
+
+trait GooglePublisher[F[_]] {
+  def createTopic(projectTopicName: ProjectTopicName): F[Unit]
+  def publish[A: Encoder]: Sink[F, A]
+}
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
@@ -1,10 +1,8 @@
 package org.broadinstitute.dsde.workbench.google2
 
-import com.google.pubsub.v1.ProjectTopicName
 import fs2.Sink
 import io.circe.Encoder
 
 trait GooglePublisher[F[_]] {
-  def createTopic(projectTopicName: ProjectTopicName): F[Unit]
   def publish[A: Encoder]: Sink[F, A]
 }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
@@ -8,7 +8,7 @@ trait GooglePublisher[F[_]] {
   /**
     * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
     */
-  def publish[A: Encoder]: Sink[F, A]
+  def publish[MessageType: Encoder]: Sink[F, MessageType]
 
   /**
     * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
@@ -17,7 +17,7 @@ trait GooglePublisher[F[_]] {
 }
 
 object GooglePublisher {
-  def resource[F[_]: Async: Timer: ContextShift, A: Encoder](config: PublisherConfig): Resource[F, GooglePublisher[F]] = for {
+  def resource[F[_]: Async: Timer: ContextShift, MessageType: Encoder](config: PublisherConfig): Resource[F, GooglePublisher[F]] = for {
     publisher <- GooglePublisherInterpreter.publisher(config)
   } yield GooglePublisherInterpreter(publisher, config.retryConfig)
 }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
@@ -9,43 +9,55 @@ import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient, TopicAdminSettin
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.{ProjectTopicName, PubsubMessage}
-import com.google.rpc.Code
 import fs2.Sink
 import io.circe.Encoder
 import io.circe.syntax._
+import org.broadinstitute.dsde.workbench.RetryConfig
 
-private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer: ContextShift](
+private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer](
                                                          publisher: Publisher,
-                                                         topicAdminClient: TopicAdminClient
+                                                         retryConfig: RetryConfig
                                      ) extends GooglePublisher[F] {
-  def publish[A: Encoder]: Sink[F, A] = Sink {
-    a =>
-      val byteString = ByteString.copyFromUtf8(a.asJson.toString()) //This will turn a case class into raw json string
-      val message = PubsubMessage.newBuilder().setData(byteString).build()
-      Async[F].async[String]{
-        cb =>
-          ApiFutures.addCallback(
-            publisher.publish(message),
-            callBack(cb),
-            MoreExecutors.directExecutor()
-          )
-      }.void
+  def publish[A: Encoder]: Sink[F, A] = in => {
+    in.flatMap{
+      a =>
+        val byteString = ByteString.copyFromUtf8(a.asJson.noSpaces) //This will turn a case class into raw json string
+        retryGoogleF(retryConfig)(asyncPublishMessage(byteString))
+    }
   }
+
+  def publishString: Sink[F, String] = in => {
+    in.flatMap{
+      str =>
+        val byteString = ByteString.copyFromUtf8(str)
+        retryGoogleF(retryConfig)(asyncPublishMessage(byteString))
+    }
+  }
+
+  private def asyncPublishMessage(byteString: ByteString): F[Unit] = Async[F].async[String]{
+    cb =>
+      val message = PubsubMessage.newBuilder().setData(byteString).build()
+      ApiFutures.addCallback(
+        publisher.publish(message),
+        callBack(cb),
+        MoreExecutors.directExecutor()
+      )
+  }.void
 }
 
 object GooglePublisherInterpreter {
   def apply[F[_]: Async: Timer: ContextShift](
              publisher: Publisher,
-             topicAdminClient: TopicAdminClient
-           ): GooglePublisherInterpreter[F] = new GooglePublisherInterpreter(publisher, topicAdminClient)
+             retryConfig: RetryConfig
+           ): GooglePublisherInterpreter[F] = new GooglePublisherInterpreter(publisher, retryConfig)
 
-  def publisher[F[_]: Sync](pathToJson: String, projectTopicName: ProjectTopicName): Resource[F, Publisher] =
+  def publisher[F[_]: Sync](config: PublisherConfig): Resource[F, Publisher] =
     for{
-      credentialFile <- org.broadinstitute.dsde.workbench.util.readFile(pathToJson)
+      credentialFile <- org.broadinstitute.dsde.workbench.util.readFile(config.pathToCrentialJson)
       credential = ServiceAccountCredentials.fromStream(credentialFile)
       pub <- Resource.make(
         Sync[F].delay(Publisher
-          .newBuilder(projectTopicName)
+          .newBuilder(config.projectTopicName)
           .setCredentialsProvider(FixedCredentialsProvider.create(credential))
           .build()))(p => Sync[F].delay(p.shutdown()))
       topicAdminClient <- Resource.make(
@@ -57,10 +69,12 @@ object GooglePublisherInterpreter {
       )(client => Sync[F].delay(client.shutdown()))
       _ <- Resource.liftF(
         Sync[F]
-        .delay(topicAdminClient.createTopic(projectTopicName))
+        .delay(topicAdminClient.createTopic(config.projectTopicName))
         .void
         .recover {
-          case e: io.grpc.StatusRuntimeException if(e.getStatus.getCode == Code.ALREADY_EXISTS) => ()
+          case _: com.google.api.gax.rpc.AlreadyExistsException => ()
         })
     } yield pub
 }
+
+final case class PublisherConfig(pathToCrentialJson: String, projectTopicName: ProjectTopicName, retryConfig: RetryConfig)
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
@@ -19,28 +19,28 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer](
                                                          publisher: Publisher,
                                                          retryConfig: RetryConfig
                                      ) extends GooglePublisher[F] {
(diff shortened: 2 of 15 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -0,0 +1,66 @@
...
+import io.circe.Encoder
+import io.circe.syntax._
+
+private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer: ContextShift](
+                                                         publisher: Publisher,
+                                                         topicAdminClient: TopicAdminClient
+                                     ) extends GooglePublisher[F] {
+  def publish[A: Encoder]: Sink[F, A] = Sink {
+    a =>
+      val byteString = ByteString.copyFromUtf8(a.asJson.toString()) //This will turn a case class into raw json string
```
dvoet: I would use `noSpaces` instead of `toString`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
```diff
@@ -0,0 +1,8 @@
+package org.broadinstitute.dsde.workbench.google2
+
+import fs2.Sink
+import io.circe.Encoder
+
+trait GooglePublisher[F[_]] {
+  def publish[A: Encoder]: Sink[F, A]
```
dvoet: should we also have a function that just takes a `String`? Maybe someone does not want a json message.

dvoet: we should also have a function that can publish more than one message at a time. Watch out for limits (https://cloud.google.com/pubsub/quotas):
per publish request
10MB (total size)
1,000 messages

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -0,0 +1,66 @@
...
+import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient, TopicAdminSettings}
+import com.google.common.util.concurrent.MoreExecutors
+import com.google.protobuf.ByteString
+import com.google.pubsub.v1.{ProjectTopicName, PubsubMessage}
+import com.google.rpc.Code
+import fs2.Sink
+import io.circe.Encoder
+import io.circe.syntax._
+
+private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer: ContextShift](
```
dvoet: do you need Timer and ContextShift?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -0,0 +1,66 @@
...
+                                                         topicAdminClient: TopicAdminClient
+                                     ) extends GooglePublisher[F] {
+  def publish[A: Encoder]: Sink[F, A] = Sink {
+    a =>
+      val byteString = ByteString.copyFromUtf8(a.asJson.toString()) //This will turn a case class into raw json string
+      val message = PubsubMessage.newBuilder().setData(byteString).build()
+      Async[F].async[String]{
+        cb =>
+          ApiFutures.addCallback(
+            publisher.publish(message),
```
dvoet: we need retries

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -0,0 +1,66 @@
...
+            .build()))
+      )(client => Sync[F].delay(client.shutdown()))
+      _ <- Resource.liftF(
+        Sync[F]
+        .delay(topicAdminClient.createTopic(projectTopicName))
+        .void
+        .recover {
+          case e: io.grpc.StatusRuntimeException if(e.getStatus.getCode == Code.ALREADY_EXISTS) => ()
+        })
+    } yield pub
```
dvoet: I find all the boilerplate with Sync and Resource makes the code difficult to read. All the pertinent details are hidden. Would extracting each part of the for comp into a function help?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -0,0 +1,112 @@
...
+          Subscriber
+          .newBuilder(subscription, receiver(queue)) //TODO: set credentials correctly
+            .setCredentialsProvider(FixedCredentialsProvider.create(credential))
+            .build())
+      )(s => Sync[F].delay(s.stopAsync()))
+    } yield sub
+  }
+}
+
+final case class SubsriberConfig(projectTopicName: ProjectTopicName, achDeadLine: FiniteDuration)
```
dvoet: spelling: Subs c riberConfig and ac k DeadLine

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -0,0 +1,112 @@
...
+      _ <- Resource.liftF(Async[F].delay(
+        subscriptionAdminClient.createSubscription(subscription, subsriberConfig.projectTopicName, PushConfig.getDefaultInstance, subsriberConfig.achDeadLine.toSeconds.toInt)
+      ).void.recover{
+        case e: io.grpc.StatusRuntimeException if(e.getStatus.getCode == Code.ALREADY_EXISTS) => ()
+      })
+      sub <- Resource.make(
+        Sync[F].delay(
+          Subscriber
+          .newBuilder(subscription, receiver(queue)) //TODO: set credentials correctly
+            .setCredentialsProvider(FixedCredentialsProvider.create(credential))
```
dvoet: There are some other settings that looks like we should expose, e.g. the flow control settings

dvoet (changes requested):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
Qi77Qi: I can add `pubString`. I don't think google's publisher client expose a `publishN` function.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
Qi77Qi: no, I don't...nice catch 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
Qi77Qi: I kinda prefer to have them all together here so that I don't need to jump around to see what initialization and shutdown look like for each resource. But maybe I'm not seeing what you have in mind. Let me know if you have some examples.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
Qi77Qi: since fs2 is pull based, so I don't think we need flow control settings?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -0,0 +1,80 @@
...
+            .newBuilder()
+            .setCredentialsProvider(FixedCredentialsProvider.create(credential))
+            .build()))
+      )(client => Sync[F].delay(client.shutdown()))
+      _ <- Resource.liftF(
+        Sync[F]
+        .delay(topicAdminClient.createTopic(config.projectTopicName))
+        .void
+        .recover {
+          case _: com.google.api.gax.rpc.AlreadyExistsException => ()
```
Qi77Qi: it's interesting that emulator and real pubsub didn't throw the same exception type...
from real pubsub it looks like
```
[info]   com.google.api.gax.rpc.AlreadyExistsException: io.grpc.StatusRuntimeException: ALREADY_EXISTS: Resource already exists in the project (resource=qi-pubsub-test).
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
dvoet: I think we do. My read of https://cloud.google.com/pubsub/docs/pull makes me think that the client library will suck as many messages as it can onto the client for local processing. The default limit seems to be 10k messages or 10MB. I don't really see a reason for numbers as large as that.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -0,0 +1,111 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+private[google2] class GoogleSubscriberInterpreter[F[_]: Async: Timer: ContextShift, A](
+                                                         subscriber: Subscriber,
+                                                         queue: fs2.concurrent.Queue[F, Event[A]]
+                                     ) extends GoogleSubscriber[F, A] {
+  val messages: Stream[F, Event[A]] = queue.dequeue
+
+  def start: F[Unit] = Async[F].async[Unit]{
+        cb =>
```
dvoet: can we call this callback instead of cb or something more descriptive?

Qi77Qi: it's quite typical to call this `cb` when using `async`..also calling it callback makes it easy to confuse with google's [callback](https://github.com/broadinstitute/workbench-libs/pull/201/files#diff-19849200e5d5dfcd69380863c18d949bR15) argument

dvoet (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import cats.effect.{Async, ContextShift, Resource, Timer}
+import fs2.Sink
+import io.circe.Encoder
+
+trait GooglePublisher[F[_]] {
+  /**
+    * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
+    */
+  def publish[MessageType: Encoder]: Sink[F, MessageType]
```
marctalbott, line 11: I'm not sure I understand this syntax, are these both implicits? Or is `MessageType` a kind of `Encoder`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+    * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
+    */
+  def publishString: Sink[F, String]
+}
+
+object GooglePublisher {
+  def resource[F[_]: Async: Timer: ContextShift, MessageType: Encoder](config: PublisherConfig): Resource[F, GooglePublisher[F]] = for {
```
marctalbott, line 20: Same question here, from what I understand this means `F[_]` implements (or has these traits) `Async`, `Timer`, `ContextShift`, but how does that then relate to `MessageType` and `Encoder`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -0,0 +1,134 @@
...
+          Logger[F].info(s"failed to publish $message to internal Queue due to $e")
+          consumer.nack() //pubsub will resend the message up to ackDeadlineSeconds (this is configed during subscription creation
+        case Right(_) =>
+          ()
+      }
+    }
+  }
+
+  def subscriber[F[_]: Effect: Logger, MessageType: Decoder](subsriberConfig: SubscriberConfig, queue: fs2.concurrent.Queue[F, Event[MessageType]]): Resource[F, Subscriber] = {
+    val subscription = ProjectSubscriptionName.of(subsriberConfig.projectTopicName.getProject, subsriberConfig.projectTopicName.getTopic)
```
marctalbott: spelling: subscriberConfig

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -0,0 +1,134 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+private[google2] class GoogleSubscriberInterpreter[F[_]: Async: Timer: ContextShift, MessageType](
+                                                                                         subscriber: Subscriber,
+                                                                                         queue: fs2.concurrent.Queue[F, Event[MessageType]]
+                                                                                       ) extends GoogleSubscriber[F, MessageType] {
+  val messages: Stream[F, Event[MessageType]] = queue.dequeue
+
+  def start: F[Unit] = Async[F].async[Unit]{
+    callback =>
```
marctalbott, line 29: Where does this `callback` come from?

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GooglePubSubSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+          .build())))(client => IO(client.shutdown()))
+      subscriptionAdminClient <- Resource.make(IO(SubscriptionAdminClient.create(
+        SubscriptionAdminSettings.newBuilder()
+          .setTransportChannelProvider(channelProvider)
+          .setCredentialsProvider(credentialsProvider)
+          .build())))(client => IO(client.shutdown()))
+      pub <- Resource.make(IO(Publisher.newBuilder(projectTopicName)
+        .setChannelProvider(channelProvider)
+        .setCredentialsProvider(credentialsProvider)
+        .build()))(p => /*IO(p.shutdown()) >>*/ IO.unit) //TODO: shutdown properly. Somehow this hangs the publisher unit test
```
marctalbott, line 111: Are there tickets for these TODOs already?

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GooglePubSubSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+      _ <- localPubsub(projectTopicName, queue).use {
+        case (pub, sub) =>
+          val subScribeStream = (Stream.emits(people) to pub.publish[Person]) ++ Stream.eval(sub.start)
+
+          val processEvents: Stream[IO, Unit] = sub.messages.zipWithIndex.evalMap[IO, Unit]{
+            case (event, index)=>
+              if(expectedPeople.contains(event.msg)) {
+                expectedPeople = expectedPeople.filterNot(_ == event.msg)
+                if(index.toInt == people.length - 1)
+                  IO(event.consumer.ack()).void >> terminateSubscriber.set(true)
```
marctalbott, line 65: From what I can tell, the `>>` is an alias for `flatMap`? Could we just use `flatMap` here for readability's sake or is `>>` a sufficiently common alias that I should just get used to looking at it?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
dvoet: in `resource[F[_]: Async: Timer: ContextShift, MessageType: Encoder]` the `:` means something other inheritance. The way to know is that they are in the parameterized type list. In this case it means that there must be a companion implicit object. E.g. for `MessageType` there must be an implicit object in scope of type `Encoder[MessageType]`. For `F[_]` there must be `Async[F[_]]`, `Timer[F[_]]` and `ContextShift[F[_]]`.

dvoet: https://docs.scala-lang.org/tutorials/FAQ/context-bounds.html

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -19,28 +19,28 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer](
                                                          publisher: Publisher,
                                                          retryConfig: RetryConfig
                                      ) extends GooglePublisher[F] {
-  def publish[A: Encoder]: Sink[F, A] = in => {
-    in.flatMap{
-      a =>
-        val byteString = ByteString.copyFromUtf8(a.asJson.noSpaces) //This will turn a case class into raw json string
-        retryGoogleF(retryConfig)(asyncPublishMessage(byteString))
+  def publish[MessageType: Encoder]: Sink[F, MessageType] = in => {
+    in.flatMap {
```
Qi77Qi, line 23: when I see `A`, I know right away that it can be any type...but if I see `MessageType`, I have to look closer to make sure if it's not some type alias defined elsewhere.

But if `MessageType` makes it more intuitive for other people, I'm fine with it

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GooglePubSubSpec.scala
Qi77Qi: @marctalbott  These TODOs probably don't worth tickets. It'll probably need quite some time to figure out why the publisher hangs, which might turn out some bug in emulator publisher's shutdown code. But it really doesn't affect us that much.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
Qi77Qi: this `callback` is used to terminate the `F`. When `callback` is called with either a left value or a right value, `F` will be terminated with either an exception or a successful value.

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GooglePubSubSpec.scala
Qi77Qi: `io1 >> io2` is indeed just alias for `io1.flatMap(_ => io2)`...but it's probably the most common syntax when using fs2 or some other FP libraries (it's widely used in fs2 source code too). So I think it's worth getting some familiarity.

marctalbott (approved):

PR #202, opened 2019-01-28
Add KMS interpreter and service

marctalbott:
This is still a wip as it may change depending on the automation test I'm currently working on for Sam, but this is what I've got so far

**PR checklist**

In all cases:


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleKmsInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+import scala.language.higherKinds
+
+/**
+  * Interpreter for Google KMS (Key Management Service) algebra
+  *
+  * created by mtalbott on 1/18/19
+  */
+
+private[google] class GoogleKmsInterpreter[F[_]: Sync: ContextShift](client: KeyManagementServiceClient, blockingEc: ExecutionContext) extends GoogleKmsService[F] {
+  override def createKeyRing(project: String, location: String, keyRingId: String): F[KeyRing] = {
```
Qi77Qi: there's a `GoogleProject` type I think.
I'll make custom type for `location` and `keyRingId`. Stringy types are too error prone

Same comment goes for other methods

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleKmsInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+
+      _ <- setIamPolicy(project, location, keyRingId, keyId, newPolicy)
+    } yield newPolicy
+  }
+
+  override def removeMemberFromKeyPolicy(project: String, location: String, keyRingId: String, keyId: String, member: String, role: String): F[Policy] = {
+    for {
+      currentIamPolicyOpt <- getIamPolicy(project, location, keyRingId, keyId)
+
+      otherBindings = currentIamPolicyOpt.get.getBindingsList.asScala.toList.filter(binding => !binding.getRole.equals(role))
```
Qi77Qi: calling `.get` on `Option` will throw exception. It should be avoided in general. I think for this case, you can use `map` to get inside the `Option`, and at the end do `.getOrElse(List.empty)`

Something like `currentIamPolicyOpt.map(x => x.getBindingsList.asScala.toList.filter(binding => !binding.getRole.equals(role)).getOrElse(List.empty)`

This way you don't need to worry about exceptions.


Also, double check the equality check works the way you expect

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleKmsInterpreter.scala
```diff
@@ -0,0 +1,120 @@
...
+      newBindings = currentIamPolicyOpt.get.getBindingsList.asScala.toList.filter(binding => binding.getRole.equals(role)).map { binding =>
```
Qi77Qi: same comment. Don't use `.get` on `Option`, it defeats the purpose of using `Option` in the first place.

google/src/test/scala/org/broadinstitute/dsde/workbench/google/mock/MockGoogleKmsInterpreter.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.google.iam.v1.Policy
+import org.broadinstitute.dsde.workbench.google.GoogleKmsService
+
+/**
+  * Mock Google Key Management Service Interpreter
+  *
+  * created by mtalbott 1/28/19
+  */
+
+object MockGoogleKmsInterpreter extends GoogleKmsService[IO] {
```
Qi77Qi: minor, I kinda prefer `Fakexxx`. It's the same naming as some emulators, and it's easy to know that it's new google code as opposed to older google module.

Qi77Qi:
👍  overall looks really good...left a few small comments

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleKmsInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  override def addMemberToKeyPolicy(project: GoogleProject, location: Location, keyRingId: KeyRingId, keyId: KeyId, member: String, role: String): F[Policy] = {
+    for {
+      currentIamPolicyOpt <- getIamPolicy(project, location, keyRingId, keyId)
+
+      newBinding = Binding.newBuilder()
+        .setRole(role)
+        .addMembers(member)
+        .build()
+      newPolicy = Policy.newBuilder(currentIamPolicyOpt.get)
```
dvoet: .get is bad practice... how can you handle None?

dvoet (changes requested):
1 little thing

dvoet (approved):

Qi77Qi (approved):
thought I approved it a while back..Maybe the approval is gone since you rebased?

Qi77Qi:
nvm..I was thinking of another PR

PR #205, opened 2019-02-28
add notification creater and topic creator

Qi77Qi:
Code is tested from console.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
@@ -6,33 +6,34 @@ import com.google.api.core.ApiFutures
 import com.google.api.gax.core.FixedCredentialsProvider
 import com.google.api.gax.rpc.AlreadyExistsException
 import com.google.auth.oauth2.ServiceAccountCredentials
-import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient, TopicAdminSettings}
+import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient}
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.{ProjectTopicName, PubsubMessage}
-import fs2.Sink
+import fs2.Pipe
+import io.chrisdavenport.log4cats.Logger
 import io.circe.Encoder
 import io.circe.syntax._
 import org.broadinstitute.dsde.workbench.RetryConfig
 
-private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer](
+private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer: Logger](
                                                          publisher: Publisher,
                                                          retryConfig: RetryConfig
                                      ) extends GooglePublisher[F] {
-  def publish[MessageType: Encoder]: Sink[F, MessageType] = in => {
+  def publish[MessageType: Encoder]: Pipe[F, MessageType, Unit] = in => {
     in.flatMap {
       message =>
         publishMessage(message.asJson.noSpaces) //This will turn message case class into raw json string
     }
   }
 
-  def publishString: Sink[F, String] = in => {
+  def publishString: Pipe[F, String, Unit] = in => {
     in.flatMap(publishMessage)
   }
 
   private def publishMessage(message: String) = {
     val byteString = ByteString.copyFromUtf8(message)
-    retryGoogleF(retryConfig)(asyncPublishMessage(byteString))
+    retryGoogleF(retryConfig)(asyncPublishMessage(byteString), None, s"Publishing $message")
   }
 
   private def asyncPublishMessage(byteString: ByteString): F[Unit] = Async[F].async[String]{
@@ -47,7 +48,7 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer](
 }
 
 object GooglePublisherInterpreter {
-  def apply[F[_]: Async: Timer: ContextShift](
+  def apply[F[_]: Async: Timer: ContextShift: Logger](
              publisher: Publisher,
              retryConfig: RetryConfig
            ): GooglePublisherInterpreter[F] = new GooglePublisherInterpreter(publisher, retryConfig)
@@ -56,7 +57,7 @@ object GooglePublisherInterpreter {
     for {
       credential <- credentialResource(config.pathToCredentialJson)
       publisher <- publisherResource(config.projectTopicName, credential)
-      topicAdminClient <- topicAdminClientResource(credential)
+      topicAdminClient <- GoogleTopicAdminInterpreter.topicAdminClientResource(credential)
       _ <- createTopic(config.projectTopicName, topicAdminClient)
     } yield publisher
 
@@ -68,16 +69,6 @@ object GooglePublisherInterpreter {
         .recover {
           case _: AlreadyExistsException => ()
         })
-  }
-
-  def topicAdminClientResource[F[_] : Sync](credential: ServiceAccountCredentials): Resource[F, TopicAdminClient] = {
-    Resource.make(
-      Sync[F].delay(TopicAdminClient.create(
-        TopicAdminSettings
-          .newBuilder()
-          .setCredentialsProvider(FixedCredentialsProvider.create(credential))
-          .build()))
-    )(client => Sync[F].delay(client.shutdown()))
   }
 
   private def publisherResource[F[_] : Sync](topicName: ProjectTopicName, credential: ServiceAccountCredentials): Resource[F, Publisher] = {
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -3,7 +3,6 @@ package google2
 
 import java.time.Instant
 
-import cats.data.OptionT
 import cats.effect._
 import cats.implicits._
 import com.google.auth.oauth2.ServiceAccountCredentials
@@ -11,30 +10,37 @@ import com.google.cloud.storage.BucketInfo.LifecycleRule
 import com.google.cloud.storage.Storage.{BlobListOption, BucketTargetOption}
 import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, BucketInfo, Storage, StorageException, StorageOptions}
 import fs2.{Stream, text}
+import io.chrisdavenport.log4cats.Logger
+import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject}
 
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 
-private[google2] class GoogleStorageInterpreter[F[_]](db: Storage,
+private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async: Logger](db: Storage,
                                       blockingEc: ExecutionContext,
                                       retryConfig: RetryConfig
-                                     )(
-  implicit cs: ContextShift[F],
-  timer: Timer[F],
-  asyncF: Async[F]
-) extends GoogleStorageService[F] {
-  private def retryStorageF[A]: F[A] => Stream[F, A] = fa => retryGoogleF(retryConfig)(fa)
+                                     )extends GoogleStorageService[F] {
+  private def retryStorageF[A]: (F[A], Option[TraceId], String) => Stream[F, A] = retryGoogleF(retryConfig)
 
-  override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName] = {
+  override def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000, traceId: Option[TraceId] = None): Stream[F, GcsObjectName] = {
     for {
-      firstPage <- retryStorageF(blockingF(asyncF.delay(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize.longValue()), BlobListOption.currentDirectory()))))
+      firstPage <- retryStorageF(
+        blockingF(Async[F].delay(db.list(bucketName.value, BlobListOption.prefix(objectNamePrefix), BlobListOption.pageSize(maxPageSize.longValue()), BlobListOption.currentDirectory()))),
+        traceId,
+        s"com.google.cloud.storage.Storage.list($bucketName, ${BlobListOption.prefix(objectNamePrefix)}, ${BlobListOption.pageSize(maxPageSize.longValue())}, ${BlobListOption.currentDirectory()})"
+      )
       page <- Stream.unfoldEval(firstPage){
         currentPage =>
           Option(currentPage).traverse{
             p =>
-              retryStorageF(blockingF(asyncF.delay(p.getNextPage))).compile.lastOrError.map(next => (p, next))
+              val fetchNext = retryStorageF(
+                blockingF(Async[F].delay(p.getNextPage)),
+                traceId,
+                s"com.google.api.gax.paging.Page.getNextPage"
+              )
+              fetchNext.compile.lastOrError.map(next => (p, next))
           }
       }
       objects <- Stream.fromIterator[F, Blob](page.getValues.iterator().asScala).map{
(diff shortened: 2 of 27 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -6,11 +6,11 @@ import com.google.api.core.ApiFutures
 import com.google.api.gax.core.FixedCredentialsProvider
 import com.google.api.gax.rpc.AlreadyExistsException
 import com.google.auth.oauth2.ServiceAccountCredentials
-import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient, TopicAdminSettings}
+import com.google.cloud.pubsub.v1.{Publisher, TopicAdminClient}
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.protobuf.ByteString
 import com.google.pubsub.v1.{ProjectTopicName, PubsubMessage}
-import fs2.Sink
+import fs2.Pipe
```
Qi77Qi, line 13: newer version of fs2 deprecated `Sink` because they want to reduce type aliases.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
```diff
@@ -0,0 +1,114 @@
...
+  case object ObjectArchive extends NotificationEventTypes {
+    def asString: String = "OBJECT_ARCHIVE"
+  }
+}
+
+final case class Filters(eventTypes: List[NotificationEventTypes], objectNamePrefix: Option[String])
+private[google2] final case class NotificationRequest(topic: ProjectTopicName, payloadFormat: String, eventTypes: List[NotificationEventTypes], objectNamePrefix: Option[String])
+private[google2] final case class NotificationResponse(items: Option[NonEmptyList[Notification]])
+final case class Notification(topic: ProjectTopicName)
+final case class MetadataNotificationCreaterConfig(pathToCredentialJson: String, googleUrl: Uri)
```
dvoet: Metadata does not belong in the name of this class

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
```diff
@@ -0,0 +1,114 @@
...
+        ))
+        // POST request will create multiple notifications for same bucket and topic with different ID; Hence we check if
+        // a notification for the given bucket and topic already exists. If yes, we do nothing; if no, we create it
+      _ <- if(notifications.items.exists(nel => nel.toList.exists(x => x.topic === topic)))
+          Sync[F].pure(())
+        else httpClient.expect[Notification](Request[F](
+          method = Method.POST,
+          uri = notificationUri,
+          headers = headers
+        ).withEntity(notificationBody))
```
dvoet: need retries and logging for google calls

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
```diff
@@ -0,0 +1,114 @@
...
+      } yield ()
+  }
+}
+
+object GoogleStorageNotificationCreaterInterpreter {
+  implicit val projectTopicNameEncoder: Encoder[ProjectTopicName] = Encoder.encodeString.contramap(
```
dvoet: what is contramap? I looked it up and did not find much understandable.

`Generally speaking, if you have some context F[A] for type A, and you can get an A value out of a B value — Contravariant allows you to get the F[B] context for B.`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
```diff
@@ -0,0 +1,114 @@
...
+    projectTopicName =>
+      s"projects/${projectTopicName.getProject}/topics/${projectTopicName.getTopic}"
+  )
+
+  implicit val notificationEventTypesEncode: Encoder[NotificationEventTypes] = Encoder.encodeString.contramap(eventType => eventType.asString)
```
dvoet: missing `r` at end of notificationEventTypesEncode

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  implicit val notificationRequestEncoder: Encoder[NotificationRequest] = Encoder.forProduct4(
+    "topic",
+    "payload_format",
+    "event_types",
+    "object_name_prefix"
+  )(x => NotificationRequest.unapply(x).get)
+
+  implicit val projectTopicNameDecoder: Decoder[ProjectTopicName] = Decoder.decodeString.emap { s =>
```
dvoet: why do we need to parse this? I feel like we will just get ourselves into trouble when google decides to change it. Why don't we treat the topic name as an opaque string?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdmin.scala
```diff
@@ -0,0 +1,43 @@
...
+    * * `domain:{domain}`: A Google Apps domain name that represents all the
+    *    users of that domain. For example, `google.com` or `example.com`.
+    * @return
+    */
+  def createWithPublisherMembers(projectTopicName: ProjectTopicName, members: List[Identity]):F[Unit]
+}
+
+
+object GoogleTopicAdmin {
+  def fromCredentialPath[F[_]: Logger: Sync](pathToCredential: String): Resource[F, GoogleTopicAdmin[F]] = for {
```
dvoet: are Logger or Sync required, same question below

dvoet (changes requested):
a couple minor comments but in general all calls to google need retries and we need to log all calls to google.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
Qi77Qi: When I first encounter `contramap`, I found it hard to understand abstractly too. It's usually easier to start with concrete example. Here, what I'm doing is, deriving a `Encoder[ProjectTopicName]` from the built-in `Encoder[String]`. In order to do that, `circe` provides `contramap` combinator, which does exactly that as long as I tell it "how to convert `ProjectTopicName` to `String`".

 You can think of `contramap` as the reverse of `map`. For instance, on the decoder side, I could have done `Decoder.decodeString.map(s => some function convert s to ProjectTopicName)` to get a `Decoder[ProjectTopicName]` (I didn't use simple `map` because I want to handle failures when converting from string to `ProjectTopicName`, so I used `emap`).

It's probably hard to remember `contramap` is for `Encoder`, `map` is for `Decoder`, so what I normally do is knowing I need to derive some `Encoder[A]`, `Decoder[A]` from built-in ones, just try typing either `map`,  `contramap` and see which one comes up

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
Qi77Qi: we need to parse this because we need to check if this matches the topic user wants to create. If we don't parse it, then we are expecting user to input a String like `//pubsub.googleapis.com/projects/{project-identifier}/topics/{my-topic}` as `topic`, which is very error prone

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdmin.scala
Qi77Qi: `Logger` is required because of https://github.com/broadinstitute/workbench-libs/pull/205/files#diff-0e40702893382d56e7c9ae448bc0c692R17

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreater.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
+import org.http4s.client.Client
+import org.http4s.client.middleware.{Logger, Retry, RetryPolicy}
+
+trait GoogleStorageNotificationCreater[F[_]] {
+  def createNotification(topic: ProjectTopicName, bucketName: GcsBucketName, filters: Filters): F[Unit]
+}
+
+object GoogleStorageNotificationCreater {
+  def apply[F[_]: Concurrent: Timer](httpClient: Client[F], config: MetadataNotificationCreaterConfig): GoogleStorageNotificationCreater[F] =
```
Qi77Qi: I'd like to keep this constructor because the input `httpClient` could already have retry and logging enabled

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
Qi77Qi: addressed in https://github.com/broadinstitute/workbench-libs/pull/205/files#diff-9ccb856d00f16de514c16be293b328f7R17

util/src/main/scala/org/broadinstitute/dsde/workbench/util/ExecutionContexts.scala
```diff
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.workbench.util
 import java.util.concurrent.{ExecutorService, Executors}
 
 import cats.effect.{Resource, Sync}
-import cats.implicits._
```
Qi77Qi, line 5: this is no longer needed because later version of cats-effect added `map` directly on `Resource` instead of relying on syntax extension

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
dvoet: I think I get it now, but this is the kind of thing that makes code hard to understand. The doc for this function is `Create a new Encoder by applying a function to a value of type B before encoding as an A.` The abstract types make it hard to understand and the type A is out of context so it takes a bunch of squinting to figure it out.

Unless I know a bunch of stuff already it is hard to understand what the code is doing. And the first level of documentation I can find is not very helpful as it is far too abstract.

dvoet:
also a spelling mistake throughout creator, not creater

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreater.scala
```diff
@@ -0,0 +1,22 @@
...
+
+trait GoogleStorageNotificationCreater[F[_]] {
+  def createNotification(topic: ProjectTopicName, bucketName: GcsBucketName, filters: Filters): F[Unit]
+}
+
+object GoogleStorageNotificationCreater {
+  def apply[F[_]: Concurrent: Timer](httpClient: Client[F], config: NotificationCreaterConfig): GoogleStorageNotificationCreater[F] =
+    new GoogleStorageNotificationCreaterInterpreter[F](httpClient, config)
+
+  def withRetryAndLogging[F[_]: Concurrent: Timer](httpClient: Client[F], config: NotificationCreaterConfig, retryPolicy: RetryPolicy[F]): GoogleStorageNotificationCreater[F] = {
```
dvoet: so if I under this right, it is up to the calling code to use this. I think it would be better if the calling code does not have to deal with retries (what do they know about how to retry). The same goes for logging. I agree that the calling program should be able to modify logging but it should be via config, not code.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
Qi77Qi: The good thing is once you pick it up, it's applicable to many things :P. Here is more doc about the general idea https://typelevel.org/cats/typeclasses/contravariant.html

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
dvoet: what I am getting at is that we should represent `ProjectTopicName` as a string not its constituent parts. We can have a constructor of `ProjectTopicName` that takes project and topic and forms the right string if we want.

Anyway, I have been burned by stuff like this changing and I have found it better to treat these names as opaque strings.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdminInterpreter.scala
```diff
@@ -0,0 +1,74 @@
...
+
+      policy <- Sync[F].delay(topicAdminClient.getIamPolicy(topicName)) //getIamPolicy requires `roles/pubsub.admin` role, see https://cloud.google.com/pubsub/docs/access-control
+
+      binding = Binding.newBuilder()
+        .setRole("roles/pubsub.publisher")
+        .addAllMembers(members.map(_.getValue).asJava)
+        .build()
+      updatedPolicy = Policy.newBuilder(policy).addBindings(binding).build()
+
+      _ <- Sync[F].delay(topicAdminClient.setIamPolicy(topicName, updatedPolicy))
```
dvoet: how does logging work here?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
dvoet: I am trying to poke at the common sentiment I have been hearing that our code is really hard to ramp up on. The more times we say "once you pick it up" the harder the overall code is to understand for outsiders.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreaterInterpreter.scala
Qi77Qi: @dvoet `ProjectTopicName` is a google lib type. It's what google expects if we're using google library. It'd quite inconvenient and error prone if caller needs 2 representations of "topic" in order to create the notification(string representation), and create the topic(ProjectTopicName). If google changes representation, how would string representation save us from trouble? 

In current implementation, if google does change string representation one day, sentry should report a decoding error with exactly which field failed to decode. So I think we should be able to react relatively fast in that case.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdminInterpreter.scala
Qi77Qi: I just added a log line https://github.com/broadinstitute/workbench-libs/pull/205/files/9826a9b00014ded05d8fba131327a89dd687ec78#diff-0e40702893382d56e7c9ae448bc0c692R35

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreater.scala
```diff
@@ -0,0 +1,24 @@
...
+
+trait GoogleStorageNotificationCreater[F[_]] {
+  def createNotification(topic: ProjectTopicName, bucketName: GcsBucketName, filters: Filters): F[Unit]
+}
+
+object GoogleStorageNotificationCreater {
+  def apply[F[_]: Concurrent: Timer](httpClient: Client[F], config: NotificationCreaterConfig): GoogleStorageNotificationCreater[F] =
+    new GoogleStorageNotificationCreatorInterpreter[F](httpClient, config)
+
+  def withRetryAndLogging[F[_]: Concurrent: Timer](httpClient: Client[F], config: NotificationCreaterConfig): GoogleStorageNotificationCreater[F] = {
```
dvoet: I think this one should be the default constructor (call it apply) and the current apply function should be `withoutRetryAndLogging` if we event want it at all

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdminInterpreter.scala
dvoet: We need to have a standardized log statement for every individual google call we make in all the places we make google calls in this new google2 lib. Look at what the older google lib does. I understand that this library is different so the mechanism will be different, even the format can be different. I just want the information in our logs so that when we have google problems we can be precise about what we were doing. We need to log the api call, the parameters and any error response. Elapsed time is a nice to have (what we currently have is not particularly accurate).

I know this is not necessarily specific to this PR but I have been saying this for a while. This PR can't merge until we have the solution. Then we need to make sure that the other google2 interpreters have it too.

Qi77Qi: I added some logging https://github.com/broadinstitute/workbench-libs/pull/205/files#diff-19849200e5d5dfcd69380863c18d949bR31...

1. Things like parameters are is basically logged specifically at call site such as https://github.com/broadinstitute/workbench-libs/pull/205/files#diff-ba19528522770daf7afddd8a9a256c8bR94

2. I added an optional `traceId` field. The idea is we can search for this `traceId` in kibana and see all logs for a given `traceId`.

3. I added `LoggerSyntax` helper. The idea of `log.ctxDebug[A](some message)` is we can have a consistent logging message for a given type A.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -1,17 +1,49 @@
...
+  private implicit val loggableGoogleCallEncoder: Encoder[LoggableGoogleCall] = Encoder.forProduct4(
+    "traceId",
+    "action",
+    "duration",
+    "response"
+  )(x => LoggableGoogleCall.unapply(x).get)
+
+  implicit def logSyntax[F[_]](log: Logger[F]): LoggerSyntax[F] = LoggerSyntax(log)
+
+  def retryGoogleF[F[_]: Timer: RaiseThrowable: Logger, A](retryConfig: RetryConfig)(fa: F[A], traceId: Option[TraceId], action: String): Stream[F, A] =
```
dvoet: This needs to log per attempt, not once for all attempts. Capturing overall elapsed time is useful but the intent is to capture each call individually.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/LoggerSyntax.scala
```diff
@@ -0,0 +1,49 @@
+package org.broadinstitute.dsde.workbench.google2
+
+import cats.data.Kleisli
+import io.chrisdavenport.log4cats.Logger
+import io.circe.Encoder
+import io.circe.syntax._
+
+final case class LoggerSyntax[F[_]](log: Logger[F]) extends AnyVal {
```
dvoet: docs as to what this class does, what is `ctx` in all the val and def names below? Don't abbreviate message with msg.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -43,51 +49,61 @@ private[google2] class GoogleStorageInterpreter[F[_]](db: Storage,
...
       .toBuilder
       .setAcl(acl.asJava)
       .setDefaultAcl(acl.asJava)
       .build()
 
-    retryStorageF(blockingF(asyncF.delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value))))).compile.drain
+    retryStorageF(
+      blockingF(Async[F].delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value)))),
+      None,
+      s"creating bucket $billingProject/$bucketName with acls $acl"
```
dvoet: I would like to see something that mirrors the exact api call a little better, e.g.
`s"com.google.cloud.storage.Storage.create($bucketInfo, ${BucketTargetOption.userProject(billingProject.value)})`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreatorInterpreter.scala
```diff
@@ -0,0 +1,110 @@
...
+  implicit val notificationEventTypesEncoder: Encoder[NotificationEventTypes] = Encoder.encodeString.contramap(eventType => eventType.asString)
+
+  implicit val notificationRequestEncoder: Encoder[NotificationRequest] = Encoder.forProduct4(
+    "topic",
+    "payload_format",
+    "event_types",
+    "object_name_prefix"
+  )(x => NotificationRequest.unapply(x).get)
+
+  implicit val projectTopicNameDecoder: Decoder[ProjectTopicName] = Decoder.decodeString.emap { s =>
```
dvoet, line 57: use `ProjectTopicName.parse` instead of writing our own parsing code.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -14,28 +19,36 @@ import scala.language.higherKinds
   * We follow tagless final pattern similar to https://typelevel.org/cats-tagless/
   */
 trait GoogleStorageService[F[_]] {
-  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000): Stream[F, GcsObjectName]
+  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, maxPageSize: Long = 1000, traceId: Option[TraceId] = None): Stream[F, GcsObjectName]
```
dvoet, line 25: doc what traceId is for

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/LoggerSyntax.scala
Qi77Qi: it means logging with some context `A`, which can be anything as long as there's an `Encoder[A]` in scope.

For instance, `Logger[F].ctxError[LoggableGoogleCall]("Failed", e).run(loggableGoogleCall)` and `Logger[F].ctxInfo[LoggableGoogleCall]("Succeeded").run(loggableGoogleCall)` will log `loggableGoogleCall` in the same way given same `Encoder[LoggableGoogleCall]`.

This can be useful in application logging as well. Say we have `val context = SamAPIContext(traceId: TraceId, userAgent: UserAgent)`, for a given api request we can do things like  `Logger[F].ctxDebug[SamAPIContext]("request initiated").run(context)`, `Logger[F].ctxError[SamAPIContext]("user not found", e).run(context)`, , and they will have the context logged in the same format.

Qi77Qi: https://typelevel.org/cats/datatypes/kleisli.html if you're not familiar with `Kleisli`...if you're familiar with reader monad (Anu mentioned you guys were either using it or she's used it. I can't remember exactly), it's similar to `Kleisli`

dvoet: The terms `context` and `syntax` as they are used here don't fit in how I usually think of them (or at least I don't see it yet). Can we name things in terms of what they do instead of what they are?

But looking at it a little more, I am having trouble seeing the benefits of having this class. It seems to do 2 things, serialize json and insert a message into the json. This is only used in the google retry code and some special case logic creating topics.

I think the special case for creating topics (topic already exists) should stop being a special case and the retry config should be used instead so that an `AlreadyExistsException` is not retried.

I think the message (which could just be a `success` boolean) should be part of the LoggableGoogleCall and encode json at the call site. Also check if the logger is enabled at the call site.

dvoet (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -1,17 +1,52 @@
...
+
+  def retryGoogleF[F[_]: Sync: Timer: RaiseThrowable: Logger, A](retryConfig: RetryConfig)(fa: F[A], traceId: Option[TraceId], action: String): Stream[F, A] = {
+    val faWithLogging = for {
+      startTime <- Timer[F].clock.realTime(TimeUnit.MILLISECONDS)
+      attempted <- fa.attempt
+      endTime <- Timer[F].clock.realTime(TimeUnit.MILLISECONDS)
+      _ <- attempted match {
+        case Left(e) =>
+          val loggableGoogleCall = LoggableGoogleCall(traceId, action, (endTime - startTime).milliseconds, None, "Failed")
+          Logger[F].error(e)(loggableGoogleCall.asJson.noSpaces) //google library logs error response as well. Revisit if this is too noise
```
dvoet: Log this at info level too. If the failure is successfully retried then this wasn't an error. If it ends up getting propagated then it is up to the caller to decide if it is an error (e.g. the AlreadyExistsException is not an error when creating a topic)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleTopicAdminInterpreter.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.broadinstitute.dsde.workbench.RetryConfig
+
+import scala.concurrent.duration._
+import collection.JavaConverters._
+import fs2.Stream
+import io.circe.syntax._
+import JsonCodec._
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+class GoogleTopicAdminInterpreter[F[_]: Logger: Sync: Timer](topicAdminClient: TopicAdminClient, retryConfig: RetryConfig) extends GoogleTopicAdmin[F] {
```
ansingh7115, line 21: Do we need tests for this class?

Qi77Qi: Ideally we would, but there's not much testing we can do without talking to actual google, which we don't want to do in unit tests

ansingh7115 (approved):

PR #207, opened 2019-03-14
Add more functions to GoogleStorageService and bug fixes

Qi77Qi:
* I realize in rawls, with old way of creating the bucket, permission looked a bit weird (it said something like `Storage legacy owner rule`). So I decided to add `createBucketWithAdmin` in google2, but it may not have been the cause of the issue I ran into. It doesn't hurt to add the function tho. Fixed a bug where we weren't ignoring `already exist` exception in `createBucket` function
* Fixed a bug in GoogleStorageService, when given old service account json credential, it's not setting project properly because it can't find `project_id` field in old json. So I added an optional project field, and fallback to `project_id` field.
* Moved generators in sam to `model` since they're generators for common ADTs.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
@@ -11,7 +11,7 @@ import com.google.pubsub.v1.ProjectTopicName
 import io.chrisdavenport.log4cats.Logger
 import io.circe.syntax._
 import io.circe.{Decoder, Encoder}
-import org.broadinstitute.dsde.workbench.google2.GoogleStorageNotificationCreatorInterpreter._
+import org.broadinstitute.dsde.workbench.google2.GoogleServiceHttpInterpreter._
 import org.broadinstitute.dsde.workbench.google2.JsonCodec._
 import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
@@ -22,10 +22,10 @@ import org.http4s.client.dsl.Http4sClientDsl
 import org.http4s.headers.Authorization
 import org.http4s.{AuthScheme, Credentials, Headers, Method, Request, Response, Status, Uri}
 
-// Google api doc https://cloud.google.com/storage/docs/json_api/v1/notifications
-class GoogleStorageNotificationCreatorInterpreter[F[_]: Sync: Logger](httpClient: Client[F], config: NotificationCreaterConfig, googleCredentials: GoogleCredentials) extends GoogleStorageNotificationCreater[F] with Http4sClientDsl[F] {
+class GoogleServiceHttpInterpreter[F[_]: Sync: Logger](httpClient: Client[F], config: NotificationCreaterConfig, googleCredentials: GoogleCredentials) extends GoogleServiceHttp[F] with Http4sClientDsl[F] {
   val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, googleCredentials.refreshAccessToken().getTokenValue))
 
+  // Google api doc https://cloud.google.com/storage/docs/json_api/v1/notifications
   def createNotification(topic: ProjectTopicName, bucketName: GcsBucketName, filters: Filters, traceId: Option[TraceId]): F[Unit] = {
     val notificationUri = config.googleUrl.withPath(s"/storage/v1/b/${bucketName.value}/notificationConfigs")
     val notificationBody = NotificationRequest(topic, "JSON_API_V1", filters.eventTypes, filters.objectNamePrefix)
@@ -49,15 +49,16 @@ class GoogleStorageNotificationCreatorInterpreter[F[_]: Sync: Logger](httpClient
     } yield ()
   }
 
-  def getStorageServer(project: GoogleProject, traceId: Option[TraceId]): F[Identity] = {
-    val uri = config.googleUrl.withPath(s"v1/projects/${project.value}/serviceAccount")
+  // https://cloud.google.com/storage/docs/getting-service-account
+  def getProjectServiceAccount(project: GoogleProject, traceId: Option[TraceId]): F[Identity] = {
+    val uri = config.googleUrl.withPath(s"/storage/v1/projects/${project.value}/serviceAccount")
     val headers = Headers(Authorization(Credentials.Token(AuthScheme.Bearer, googleCredentials.refreshAccessToken().getTokenValue)))
 
-    httpClient.expectOr[Identity](Request[F](
+    httpClient.expectOr[GetProjectServiceAccountResponse](Request[F](
       method = Method.GET,
       uri = uri,
       headers = Headers(authHeader)
-    ))(onError(traceId))
+    ))(onError(traceId)).map(_.serviceAccount)
   }
 
   private def onError(traceId: Option[TraceId]): Response[F] => F[Throwable] = resp => {
@@ -69,7 +70,7 @@ class GoogleStorageNotificationCreatorInterpreter[F[_]: Sync: Logger](httpClient
   }
 }
 
-object GoogleStorageNotificationCreatorInterpreter {
+object GoogleServiceHttpInterpreter {
   implicit val notificationEventTypesEncoder: Encoder[NotificationEventTypes] = Encoder.encodeString.contramap(eventType => eventType.asString)
 
   implicit val notificationRequestEncoder: Encoder[NotificationRequest] = Encoder.forProduct4(
@@ -104,6 +105,10 @@ object GoogleStorageNotificationCreatorInterpreter {
 
   implicit def identityDeocder: Decoder[Identity] = Decoder.decodeString.map(s => Identity.serviceAccount(s))
 
+  implicit def getProjectServiceAccountResponse: Decoder[GetProjectServiceAccountResponse] = Decoder.forProduct1(
+    "email_address"
+  )(GetProjectServiceAccountResponse)
+
   implicit val eqProjectTopicName: Eq[ProjectTopicName] = Eq.instance((t1, t2) => t1.getProject == t2.getProject && t1.getTopic == t2.getTopic)
 
   def credentialResourceWithScope[F[_]: Sync](pathToCredential: String): Resource[F, GoogleCredentials] = for {
@@ -136,3 +141,4 @@ private[google2] final case class NotificationResponse(items: Option[NonEmptyLis
 final case class Notification(topic: ProjectTopicName)
 final case class NotificationCreaterConfig(pathToCredentialJson: String, googleUrl: Uri)
 final case class LoggableErrorResponse(traceId: Option[TraceId], status: Status, body: String)
+final case class GetProjectServiceAccountResponse(serviceAccount: Identity)
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
@@ -38,7 +38,7 @@ class GoogleServiceHttpInterpreter[F[_]: Sync: Logger](httpClient: Client[F], co
         headers = headers
       ))
       // POST request will create multiple notifications for same bucket and topic with different ID; Hence we check if
-      // a notification for the given bucket and topic already exists. If yes, we do nothing; if no, we create it
+      // a notification for the given bucket and topic already exists. If yes, we do nothing; else, we create it
       _ <- if(notifications.items.exists(nel => nel.toList.exists(x => x.topic === topic)))
         Sync[F].pure(())
       else httpClient.expectOr[Notification](Request[F](
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
@@ -52,8 +52,6 @@ class GoogleServiceHttpInterpreter[F[_]: Sync: Logger](httpClient: Client[F], co
   // https://cloud.google.com/storage/docs/getting-service-account
   def getProjectServiceAccount(project: GoogleProject, traceId: Option[TraceId]): F[Identity] = {
     val uri = config.googleUrl.withPath(s"/storage/v1/projects/${project.value}/serviceAccount")
-    val headers = Headers(Authorization(Credentials.Token(AuthScheme.Bearer, googleCredentials.refreshAccessToken().getTokenValue)))
-
     httpClient.expectOr[GetProjectServiceAccountResponse](Request[F](
       method = Method.GET,
       uri = uri,
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -29,7 +29,6 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
                                       retryConfig: RetryConfig
                                      )extends GoogleStorageService[F] {
   private def retryStorageF[A]: (F[A], Option[TraceId], String) => Stream[F, A] = retryGoogleF(retryConfig)
-  private def retryStorageKleisli[A] = retryGoogleKleisli[F, A](retryConfig)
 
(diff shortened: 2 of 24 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -88,19 +92,55 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
-      blockingF(Async[F].delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value)))),
-      None,
+      createBucket,
+      traceId,
       s"com.google.cloud.storage.Storage.create($bucketInfo, ${billingProject})"
     ).compile.drain
   }
 
-  override def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule], traceId: Option[TraceId] = None): F[Unit] = {
+  def createBucketWithAdminRole(googleProject: GoogleProject, bucketName: GcsBucketName, adminIdentity: Identity, traceId: Option[TraceId] = None): Stream[F, Unit] = {
```
dvoet: Would it be more flexible if this were broken into 2 functions, 1 to create the bucket and 1 to set IAM?

What identity requires `roles/storage.admin` access?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageNotificationCreatorInterpreter.scala
```diff
@@ -6,41 +6,66 @@ import cats.effect.{Resource, Sync}
...
+        Sync[F].pure(())
+      else httpClient.expectOr[Notification](Request[F](
+        method = Method.POST,
+        uri = notificationUri,
+        headers = headers
+      ).withEntity(notificationBody))(onError(traceId))
+    } yield ()
+  }
+
+  def getStorageServer(project: GoogleProject, traceId: Option[TraceId]): F[Identity] = {
```
dvoet: should this be called `getStorageServiceAccount`?

Qi77Qi: just saw the comment. Weird I don't see the email in gmail. I renamed it to `projectServiceAccount` (`the email address of the service account associated with the project that contains your Cloud Storage bucket` is how google describe it`). Let me know if this sounds good. I also renamed  `GoogleStorageNotification` to `GoogleServiceHttp` so that we can put anything that requires direct http call to this class.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: I figured it's probably common that whoever created the bucket also wants admin role on the bucket, which is rawls does, so I bundled them together.
`roles/storage.admin` is given to `806222273987-ksinuueghug8u81i36lq8aof266q19hl@developer.gserviceaccount.com`(passed in parameter), this is needed for it to create notification (it really only need ` storage.buckets.update` permission, but `roles/storage.admin` is the only existing role that has it I think).

I changed it to pass in `Map[StorageRole, NonEmptyList[Identity]`. It's basically the same as the other `createBucket` but instead of using `acl`, this one uses iam.

It's also more tricky in terms of permission if we separate setIamPolicy with whoever creates the bucket.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -88,19 +94,57 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
       .setDefaultAcl(acl.asJava)
       .build()
 
+    val createBucket = blockingF(Async[F].delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value)))).void.handleErrorWith {
+      case e: com.google.cloud.storage.StorageException if(e.getCode == 409) =>
+        Logger[F].info(s"$bucketName already exists")
```
dvoet, line 103: I don't think we should be handling 409 here. That error should be passed up to the caller since this is a generic function.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -88,19 +94,57 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
+      _ <- retry(
+        createBucket,
+        s"com.google.cloud.storage.Storage.create($bucketInfo, ${googleProject})"
+      )
+      policy <- retry(
+        blockingF(Async[F].delay(db.getIamPolicy(bucketName.value))),
+        s"com.google.cloud.storage.Storage.getIamPolicy(${bucketName})"
+      )
+      policyBuilder = policy.toBuilder()
+      updatedPolicy = roles.foldLeft(policyBuilder)((curBuilder, item) => curBuilder.addIdentity(Role.of(item._1.name), item._2.head, item._2.tail: _*)).build()
```
dvoet: no abbreviates please... what is cur in curBuilder?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -88,19 +94,57 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
-      blockingF(Async[F].delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value)))),
-      None,
+      createBucket,
+      traceId,
       s"com.google.cloud.storage.Storage.create($bucketInfo, ${billingProject})"
     ).compile.drain
   }
 
-  override def setBucketLifecycle(bucketName: GcsBucketName, lifecycleRules: List[LifecycleRule], traceId: Option[TraceId] = None): F[Unit] = {
+  def createBucketWithRoles(googleProject: GoogleProject, bucketName: GcsBucketName, roles: Map[StorageRole, NonEmptyList[Identity]], traceId: Option[TraceId] = None): Stream[F, Unit] = {
```
dvoet: I still think this should be 2 functions: `createBucket` (which we already have above) and `addBucketIamPolicy`. They seem pretty separate and usable independently. But at the very least `createBucketWithRoles` should call `createBucket`.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -88,19 +94,57 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
+    def retry[A]: (F[A], String) => Kleisli[Stream[F, ?], Option[TraceId], A] = (fa, str) => Kleisli(traceId => retryGoogleF(retryConfig)(fa, traceId, str))
```
dvoet: what would the code be like without a `Kleisli`? Every time I see it I need to lookup what it is again.

dvoet (changes requested):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
dvoet: the reason it is an all-in-one function call in rawls is that for legacy ACLs, the bucket can be created and ACLs set in a single google call. That doesn't appear to be the case in the new IAM model.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: This is the concrete action of creating bucket. In what sense this is generic?
If I leave out `handleErrorWith`, it'll be in `retryStorageF`, which I think harder to read

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: I'd have to pass `tradeId` to each `retryGoogleF` explicitly and it can easily get lost in one call.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: It's the current builder. It's the accumulator for `foldLeft` function, each iteration it adds `addIdentity` to the current builder with initial builder being `policyBuilder`.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -103,33 +108,18 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
-      policy <- retry(
+      policy <- retryStorageKleisli(
         blockingF(Async[F].delay(db.getIamPolicy(bucketName.value))),
         s"com.google.cloud.storage.Storage.getIamPolicy(${bucketName})"
       )
       policyBuilder = policy.toBuilder()
-      updatedPolicy = roles.foldLeft(policyBuilder)((curBuilder, item) => curBuilder.addIdentity(Role.of(item._1.name), item._2.head, item._2.tail: _*)).build()
-      _ <- retry(
+      updatedPolicy = roles.foldLeft(policyBuilder)((currentBuilder, item) => currentBuilder.addIdentity(Role.of(item._1.name), item._2.head, item._2.tail: _*)).build()
+      _ <- retryStorageKleisli(
```
dvoet: I now remember a problem we had with this pattern where getting the iam policy and later setting it each in their separate retry does not work. The pattern is meant to ensure only 1 modifier at a time. If another process updates IAM between the get and the set then the call to set will persistently fail.

Instead get and set should be wrapped in a single retry. If the get succeeds and the set fails the get needs to be retried as well.

And then we don't need `retryStorageKleisli` either :) as there is only 1 usage of it

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
dvoet: if you are worried about it getting missed in a call to `retryGoogleF` then don't give it a default value in `retryGoogleF`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: updated

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -1,28 +1,35 @@
...
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
+import io.circe.fs2._
 
 private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async: Logger](db: Storage,
                                       blockingEc: ExecutionContext,
                                       retryConfig: RetryConfig
                                      )extends GoogleStorageService[F] {
   private def retryStorageF[A]: (F[A], Option[TraceId], String) => Stream[F, A] = retryGoogleF(retryConfig)
+  private def retryStorageKleisli[A] = retryGoogleKleisli[F, A](retryConfig)
```
dvoet: this is unused, please remove

dvoet (approved):
remove retryStorageKleisli then 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
```diff
@@ -0,0 +1,144 @@
...
+import org.http4s.client.Client
+import org.http4s.client.dsl.Http4sClientDsl
+import org.http4s.headers.Authorization
+import org.http4s.{AuthScheme, Credentials, Headers, Method, Request, Response, Status, Uri}
+
+class GoogleServiceHttpInterpreter[F[_]: Sync: Logger](httpClient: Client[F], config: NotificationCreaterConfig, googleCredentials: GoogleCredentials) extends GoogleServiceHttp[F] with Http4sClientDsl[F] {
+  val authHeader = Authorization(Credentials.Token(AuthScheme.Bearer, googleCredentials.refreshAccessToken().getTokenValue))
+
+  // Google api doc https://cloud.google.com/storage/docs/json_api/v1/notifications
+  def createNotification(topic: ProjectTopicName, bucketName: GcsBucketName, filters: Filters, traceId: Option[TraceId]): F[Unit] = {
```
ansingh7115, line 29: Could you add a comment saying what a TraceId is? (somewhere it makes sense, maybe where its is defined)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
```diff
@@ -0,0 +1,144 @@
...
+      ))
+      // POST request will create multiple notifications for same bucket and topic with different ID; Hence we check if
+      // a notification for the given bucket and topic already exists. If yes, we do nothing; else, we create it
+      _ <- if(notifications.items.exists(nel => nel.toList.exists(x => x.topic === topic)))
+        Sync[F].pure(())
+      else httpClient.expectOr[Notification](Request[F](
+        method = Method.POST,
+        uri = notificationUri,
+        headers = headers
+      ).withEntity(notificationBody))(onError(traceId))
```
ansingh7115, line 48: should we logging more than just the trace id here and in `getProjectServiceAccount`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
```diff
@@ -0,0 +1,144 @@
...
+    } yield ()
+  }
+
+  // https://cloud.google.com/storage/docs/getting-service-account
+  def getProjectServiceAccount(project: GoogleProject, traceId: Option[TraceId]): F[Identity] = {
+    val uri = config.googleUrl.withPath(s"/storage/v1/projects/${project.value}/serviceAccount")
+    val headers = Headers(Authorization(Credentials.Token(AuthScheme.Bearer, googleCredentials.refreshAccessToken().getTokenValue)))
```
ansingh7115: `headers` is unused

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
ansingh7115: +1

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.workbench
 import java.util.concurrent.TimeUnit
 
 import cats.implicits._
+import cats.data.Kleisli
```
ansingh7115: unused import

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -58,22 +60,46 @@ trait GoogleStorageService[F[_]] {
 
   /**
     * @param traceId uuid for tracing a unique call flow in logging
+    * Acl is deprecated. Use setIamPolicy if possible
     */
-  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, acl: List[Acl], traceId: Option[TraceId] = None): F[Unit]
+  def createBucket(googleProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]] = None, traceId: Option[TraceId] = None): Stream[F, Unit]
+
+  /**
+    * @param traceId uuid for tracing a unique call flow in logging
```
ansingh7115, line 68: Ah! I see you've explained what trace id is here - could you add the comment to the definition of TraceId?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleServiceHttpInterpreter.scala
Qi77Qi: http header and body logging are enabled [here](https://github.com/broadinstitute/workbench-libs/pull/207/files#diff-87adb86e5a9c3c79b1467a69aa71e76fR25)

ansingh7115 (approved):

PR #214, opened 2019-04-13
Update Orchestration Workspaces endpoint

aweng98:
This is used in new automation tests in Orchestration PR [#715](https://github.com/broadinstitute/firecloud-orchestration/pull/715)

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.workbench.service
 
+
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.model.HttpResponse
 import akka.http.scaladsl.model.headers.Cookie
@@ -19,6 +20,7 @@ import spray.json.{DefaultJsonProtocol, _}
 
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
+
 
 trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport with DefaultJsonProtocol with RandomUtil {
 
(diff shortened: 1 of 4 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -202,6 +204,32 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
         case Some(_) => logger.info(s"Bucket read access check passed: workspace $workspaceNamespace/$workspaceName bucket readable")
       }
     }
+
+    def getWorkspaces(user: Credentials)(implicit token: AuthToken): HttpResponse = {
```
marctalbott, line 208: Is the `user` parameter necessary here? How important is it to log the user's email?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -202,6 +204,32 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
+      logger.info(s"Getting user ${user.email} workspaces")
+      getRequest(apiUrl(s"api/workspaces"))
+    }
+
+    def getWorkspaceDetail(workspaceNamespace: String, workspaceName: String)(implicit token: AuthToken): HttpResponse = {
+      logger.info(s"Getting workspace: $workspaceNamespace/$workspaceName")
+      getRequest(apiUrl(s"api/workspaces/${workspaceNamespace}/${workspaceName}"))
+    }
+
+    def getAuthorizationDomainInWorkspace(workspaceNamespace: String, workspaceName: String)(implicit token: AuthToken): List[String] = {
```
marctalbott, line 218: I believe standard practice is to have the functions here correspond to just the service's endpoints and keep parsing logic in the tests themselves

marctalbott (commented):
It generally looks good, it's just raised some questions for me about how closely the functions in the service test library should correspond to the endpoints exposed by the service vs whether additional stuff should be added for logging and parsing?

PR #223, opened 2019-04-25
Update Orchestration endpoints

aweng98:
Used by firecloud-orchestration PR [722](https://github.com/broadinstitute/firecloud-orchestration/pull/722)

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
@@ -15,7 +15,7 @@ import org.broadinstitute.dsde.workbench.service.Sam.user.UserStatusDetails
 import org.broadinstitute.dsde.workbench.service.WorkspaceAccessLevel.WorkspaceAccessLevel
 import org.broadinstitute.dsde.workbench.service.test.RandomUtil
 import org.broadinstitute.dsde.workbench.service.util.Retry
-import spray.json.{DefaultJsonProtocol, _}
+import spray.json._
 
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
@@ -244,6 +244,19 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
       logger.info(s"Getting discoverable groups for workspace: $ns/$wName")
       parseResponseAs[Seq[String]](getRequest(apiUrl(s"api/library/$ns/$wName/discoverableGroups")))
     }
+
+    def getMetadata(namespace: String, workspaceName: String)(implicit token: AuthToken): String = {
+      logger.info(s"Getting library metadata for workspace: $namespace/$workspaceName")
+      parseResponse(getRequest(apiUrl(s"api/library/$namespace/$workspaceName/metadata")))
+    }
+
+    def searchPublishedLibraryDataset(namespace: String, workspaceName: String, fieldAggregations: Map[String, Any] = Map.empty,
+                                       filters: Map[String, Any] = Map.empty, researchPurpose: Map[String, Any] = Map.empty)(implicit token: AuthToken): String = {
+      logger.info(s"Searching published library dataset for workspace: $namespace/$workspaceName")
+      val request = Map("searchString" -> workspaceName, "fieldAggregations" -> fieldAggregations, "filters" -> filters, "researchPurpose" -> researchPurpose)
+      postRequest(apiUrl("api/library/search"), request)
+    }
+
   }
 
   /*
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
@@ -16,7 +16,6 @@ import org.broadinstitute.dsde.workbench.service.WorkspaceAccessLevel.WorkspaceA
 import org.broadinstitute.dsde.workbench.service.test.RandomUtil
 import org.broadinstitute.dsde.workbench.service.util.Retry
 import spray.json._
-
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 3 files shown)


marctalbott (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -244,6 +244,18 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
+
+    def getMetadata(namespace: String, workspaceName: String)(implicit token: AuthToken): String = {
+      logger.info(s"Getting library metadata for workspace: $namespace/$workspaceName")
+      parseResponse(getRequest(apiUrl(s"api/library/$namespace/$workspaceName/metadata")))
+    }
+
+    def searchPublishedLibraryDataset(namespace: String, workspaceName: String, searchParams: Map[String, Any] = Map.empty)(implicit token: AuthToken): String = {
+      logger.info(s"Searching published library dataset for workspace: $namespace/$workspaceName")
+      val request = Map("searchString" -> workspaceName, "fieldAggregations" -> Map.empty, "filters" -> Map.empty)
+      postRequest(apiUrl("api/library/search"), if (searchParams.isEmpty) request else searchParams)
```
ansingh7115: so if I'm calling this method with searchParams, I have to pass in the name of the workspace twice? once as workspaceName and again inside searchParams?

aweng98: `workspaceName` is required, but not the `searchParam`. I provide a default `searchParam` if user doesn't specify one. I guess if the `searchParam` contains a map element `"searchString" -> workspaceName`, then yes, `workspace` name is typed twice. Not ideal and not the best way of code this. I did it because it's just the easiest and quickest. I don't think this method will be called a lot.

ansingh7115: It would be good to instead have the method be something along the lines of

```
def searchPublishedLibraryDataset(namespace: String, workspaceName: String, fieldAggregations: Map[String, Any] = Map.empty, filters: Map[String, Any] = Map.empty)(implicit token: AuthToken): String = {
logger.info(s"Searching published library dataset for workspace: $namespace/$workspaceName")
      val request = Map("searchString" -> workspaceName, "fieldAggregations" ->  fieldAggregations, "filters" -> filters)
      postRequest(apiUrl("api/library/search"), request)
```
which is simpler and the caller is not redundantly passing in the workspaceName twice

aweng98: 👍

aweng98: added missing `researchPurpose` map in method's param list.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -244,6 +244,19 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
+      logger.info(s"Getting library metadata for workspace: $namespace/$workspaceName")
+      parseResponse(getRequest(apiUrl(s"api/library/$namespace/$workspaceName/metadata")))
+    }
+
+    def searchPublishedLibraryDataset(namespace: String, workspaceName: String, fieldAggregations: Map[String, Any] = Map.empty,
+                                       filters: Map[String, Any] = Map.empty, researchPurpose: Map[String, Any] = Map.empty)(implicit token: AuthToken): String = {
+      logger.info(s"Searching published library dataset for workspace: $namespace/$workspaceName")
+      val request = Map("searchString" -> workspaceName, "fieldAggregations" -> fieldAggregations, "filters" -> filters, "researchPurpose" -> researchPurpose)
+      postRequest(apiUrl("api/library/search"), request)
+    }
```
ansingh7115, line 278: actually, looking at the orch endpoint: here's what the payload is supposed to look like:

```
{
  "searchString": "string",
  "filters": {},
  "researchPurpose": {
    "DS": [
      "http://purl.obolibrary.org/obo/DOID_4"
    ],
    "NMDS": false,
    "NCTRL": false,
    "NAGR": false,
    "POA": false,
    "NCU": false
  },
  "fieldAggregations": {},
  "from": 0,
  "size": 10,
  "sortField": "string",
  "sortDirection": "string"
}
```
the `searchString` shouldn't necessarily be the workspaceName. It should be whatever the caller is trying to search. You're also missing  `from`, `size`, `sortField`, and `sortDirection`, which all should be optional params in this method (with default values).

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -244,6 +243,41 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
+      parseResponse(getRequest(apiUrl(s"api/library/$namespace/$workspaceName/metadata")))
+    }
+
+    def searchPublishedLibraryDataset(searchString: String, from: Int = 0, size: Int = 10, sortField: String = "",
+                                      sortDirection: String = "", fieldAggregations: Map[String, Any] = Map.empty,
+                                       filters: Map[String, Any] = Map.empty, researchPurpose: Map[String, Any] = Map.empty)(implicit token: AuthToken): String = {
+      logger.info(s"Searching published library dataset")
+
+      // when researchPurpose is present in json body, it cannot be empty. below fields are required
+      val researchPurposeDefault = Map[String,Any](
```
ansingh7115: Instead of doing it this way, you could move `val researchPurposeDefault` to outside this method, then change the the method param (line 254) to `researchPurpose: Map[String, Any] = researchPurposeDefault` and then change the request (line 270) to ` "researchPurpose" -> researchPurpose,`

ansingh7115 (approved):

ansingh7115:
Approved with 1 more nitpick!
Thanks for putting up with my nitpicking :)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
aweng98: 👍 updated it.

PR #227, opened 2019-05-06
AuthDomainMatcher: increase Eventually timeout from 1 minute to 5 minutes

aweng98:
This PR should address following stacktrace that was reported in [QA-544](https://broadworkbench.atlassian.net/browse/QA-544)

Exception:
List(****) did not contain element "GroupsApiSpec_workspace 07fe8ef2-ad99-4382-8b67-cfab23cd8804"

assertion still failing in Alpha sometimes. increase timeout to 120 seconds from 60 seconds should take care tests flakiness.

Use `Must` instead `should` in assertion. Quote from http://doc.scalatest.org/2.2.4/index.html#org.scalatest.MustMatchers:

"Trait MustMatchers is an alternative to Matchers that provides the exact same meaning, syntax, and behavior as Matchers, but uses the verb must instead of should. The two traits differ only in the English semantics of the verb: should is informal, making the code feel like conversation between the writer and the reader; must is more formal, making the code feel more like a written specification."

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ To start the Google PubSub emulator for unit testing:
 
 Contains common classes and utilities for writing tests against Workbench REST web services.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.16-f63a6eb" % "test" classifier "tests"`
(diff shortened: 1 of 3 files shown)


ndbolliger (approved):
👍

PR #228, opened 2019-05-08
Requester pays support

helgridly:
**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
@@ -11,23 +11,22 @@ import akka.http.scaladsl.model.{StatusCodes, _}
 import akka.stream.ActorMaterializer
 import cats.implicits._
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
-import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.client.http.{AbstractInputStreamContent, FileContent, HttpResponseException, InputStreamContent}
 import com.google.api.services.compute.ComputeScopes
 import com.google.api.services.plus.PlusScopes
 import com.google.api.services.storage.model.Bucket.Lifecycle
 import com.google.api.services.storage.model.Bucket.Lifecycle.Rule.{Action, Condition}
 import com.google.api.services.storage.model._
-import com.google.api.services.storage.{Storage, StorageRequest, StorageScopes}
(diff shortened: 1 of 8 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleStorageDAO.scala
```diff
@@ -111,6 +112,31 @@ class HttpGoogleStorageDAO(appName: String,
     }
   }
 
+  override def getRequesterPays(bucketName: GcsBucketName): Future[Boolean] = {
+    val bucketF = retryWithRecoverWhen500orGoogleError {() => {
+      executeGoogleRequest(storage.buckets().get(bucketName.value))
+      }
+    } {
+      case e: GoogleJsonResponseException if e.getDetails.getMessage.startsWith("Bucket is requester pays") =>
+        executeGoogleRequest(storage.buckets().get(bucketName.value).setUserProject("whats-this-variable")) //FIXME
```
helgridly: OK, so here's a problem. I have a test that:
1. Sets a bucket to be requester pays using the Rawls API
2. Checks, using this DAO, the requester-pays status of the bucket.

Snag: once you've set a bucket to be RP, the call "give me the bucket info so I can see if it is indeed RP" requires a project to bill, because you're the requester, and you have to pay!

So I need to pass in that project from somewhere. I want to know how to get that variable from config into this function. Here are a few options, feel free to propose your own:

A] Pass in the project to bill as an optional parameter to this DAO.
B] Subclass this DAO, pass the project into its constructor, and then override `getRequesterPays` to use it.
C] Do some shenanigans with implicits, like have a "don't use requester pays" instance variable that's used by default, and a "here's the RP project to use" variable you can use when you want to.

@dvoet @Qi77Qi @ansingh7115 Can I get your opinions on this?

Qi77Qi: I would probably do A]

Do you need to set billing projects for all api calls once requester pays is enabled?

helgridly: for all API calls that use the bucket, yeah

helgridly:
This PR just got _way_ worse. Unfortunately there's no `setUserProject` (ie the RP-project) on any base class, so the best I could do in terms of abstracting this out as a function would be to match against every subclass of Google's `StorageRequest` class and do the same thing. It's miserable and I hate it.

helgridly:
Closed because we're not doing set RP for now.

PR #232, opened 2019-05-16
add small utility for reading json file into objects

Qi77Qi:
Recently, there were two places where I see people need to read a json file into an object A. So adding a small utility function for doing that

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ In this repo:
 
 Contains utility functions and classes.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.5-4c7acd5"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.5-TRAVIS-REPLACE-ME"`
 
 [Changelog](util/CHANGELOG.md)
 
@@ -31,7 +31,7 @@ NOTE: This library uses akka-http's implementation of spray-json and is therefor
 
 Contains utilities for instrumenting Scala code and reporting to StatsD using [metrics-scala](https://github.com/erikvanoosten/metrics-scala) and [metrics-statsd](https://github.com/ReadyTalk/metrics-statsd).
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metrics" % "0.5-4c7acd5"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metrics" % "0.5-TRAVIS-REPLACE-ME"`
 
 [Changelog](metrics/CHANGELOG.md)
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -51,7 +51,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.3-e7d949f"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.2-TRAVIS-REPLACE-ME"`
```
jdcanas: previous value got clobbered

util/src/main/scala/org/broadinstitute/dsde/workbench/util/package.scala
```diff
@@ -32,4 +40,17 @@ package object util {
...
   def readFile[F[_]](path: String)(implicit sf: Sync[F]): Resource[F, FileInputStream] =
     Resource.make(sf.delay(new FileInputStream(path)))(f => sf.delay(f.close()))
+
+  /*
+   * Example:
+   * scala> org.broadinstitute.dsde.workbench.util.readJsonFileToA[IO, List[String]](java.nio.file.Paths.get("/tmp/list"), None).compile.lastOrError.unsafeRunSync
+   * res0: List[String] = List(this is great)
+   *
+   */
+  def readJsonFileToA[F[_]: Sync: ContextShift: Concurrent, A: Decoder](path: Path, blockingExecutionContext: Option[ExecutionContext] = None): Stream[F, A] = {
```
jdcanas, line 50: what is the behavior if the path passed isn't a valid path? Will the stream be 'empty' and it fail silently, or will it return some sort of throwable wrapped in the stream?

jdcanas (approved):
Looks good except for the hash clobber. I'm not too sure if the method added needs more error handling, its not clear to me the behavior if the path passed is invalid (and, more importantly, if the caller will receive a helpful message)

util/src/main/scala/org/broadinstitute/dsde/workbench/util/package.scala
Qi77Qi: the `F` will fail with appropriate error like
```
scala> org.broadinstitute.dsde.workbench.util.readJsonFileToA[IO, List[String]](java.nio.file.Paths.get("/tmp/list"), None).compile.lastOrError.unsafeRunSync
java.nio.file.NoSuchFileException: /tmp/list
  at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
  at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
  at sun.nio.fs.UnixException.rethrowAsIOException(
```

PR #244, opened 2019-06-18
Return blob for storeObject

Qi77Qi:
Added
- Add `getBlob`

Changed
- Rename previous `getObject` to `getObjectBody`, `unsafeGetObject` to `unsafeGetObjectBody`
- provide `text/plain` as default `objectType` for `storeObject`
- Bump `http4sVersion` to `0.20.3`
- `storeObject` returns `Blob`

This is needed for welder which is required for 7/10 AOU bedford release

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -21,6 +21,8 @@ import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject}
 
 import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.global
 import scala.concurrent.duration._
 
 private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async: Logger: Linebacker](db: Storage,
@@ -240,14 +242,14 @@ object GoogleStorageInterpreter {
 
   def storage[F[_]: Sync: ContextShift](
       pathToJson: String,
-      blocker: Blocker,
+      blockingEc: ExecutionContext,
       project: Option[GoogleProject] = None // legacy credential file doesn't have `project_id` field. Hence we need to pass in explicitly
   ): Resource[F, Storage] =
     for {
       credential <- org.broadinstitute.dsde.workbench.util.readFile(pathToJson)
       project <- project match { //Use explicitly passed in project if it's defined; else use `project_id` in json credential; if neither has project defined, raise error
         case Some(p) => Resource.pure[F, GoogleProject](p)
-        case None => Resource.liftF(parseProject(pathToJson, blocker).compile.lastOrError)
+        case None => Resource.liftF(parseProject(pathToJson, blockingEc).compile.lastOrError)
       }
       db <- Resource.liftF(
         Sync[F].delay(
@@ -265,8 +267,8 @@ object GoogleStorageInterpreter {
     "project_id"
   )(GoogleProject.apply)
 
-  def parseProject[F[_]: ContextShift: Sync](pathToJson: String, blocker: Blocker): Stream[F, GoogleProject] =
-     fs2.io.file.readAll[F](Paths.get(pathToJson), blocker, 4096)
+  def parseProject[F[_]: ContextShift: Sync](pathToJson: String, executionContext: ExecutionContext = global): Stream[F, GoogleProject] =
+     fs2.io.file.readAll[F](Paths.get(pathToJson), executionContext, 4096)
           .through(byteStreamParser)
           .through(decoder[F, GoogleProject])
 }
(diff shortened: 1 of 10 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -22,6 +22,7 @@ import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectN
 
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.global
```
rtitle, line 25: Why is it needed to import the implicit global execution context -- I thought methods in the interpreter allow the caller to pass an execution context?

rtitle: Oh I misread, this isn't importing the implicit (which is `scala.concurrent.ExecutionContext.Implicits.global`) and is just used as a default arg lower down. So I think this is fine

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -49,13 +49,14 @@ trait GoogleStorageService[F[_]] {
...
-  def unsafeGetObject(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: Option[TraceId] = None): F[Option[String]]
+  def unsafeGetObjectBody(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: Option[TraceId] = None): F[Option[String]]
 
   /**
     * @param traceId uuid for tracing a unique call flow in logging
     */
-  def getObject(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: Option[TraceId] = None): Stream[F, Byte]
+  def getObjectBody(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: Option[TraceId] = None): Stream[F, Byte]
 
+  def getBlob(bucketName: GcsBucketName, blobName: GcsBlobName, traceId: Option[TraceId] = None): Stream[F, Blob]
```
rtitle, line 83: Why would I use `getBlob` over `getObjectBody`? Some javadoc would be nice.

rtitle (approved):
Looks good to me

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
Qi77Qi: `Blob` is `com.google.cloud.storage.Blob`, which comes from the java lib. The original `getObjectBody` only returns the object body. But `Blob` actually exposes lots of extra things like metadata etc, so providing this method is handy so that we don't have to write a different method for each different thing we need out of blob

I'll add some doc

dvoet (changes requested):
I think the only backwards incompatible changes are renaming the getObject functions. Instead of renaming, deprecate the old name then you don't need a minor version bump

Qi77Qi:
@dvoet I also bumped http4s's version, I think it's safer to always bump minor when we upgrade dependency

dvoet:
@Qi77Qi I am not certain I agree... if the dependency upgrade is backwards compatible why would we say our library is not? Why is it safer to bump the minor version when updating a dependency when a minor version bump in workbench-lib is supposed to signify a break in compatibility?

Qi77Qi:
@dvoet ,

1.  If we don't want to bump minor, I'll have to add `@deprecate` on `storeObject` as well. (potentially there're other things I missed)
2.  I don't quite like using `@deprecation` here because it's more meaningful if we can specify since what version, but since we're using hash cross bin compact versions, this field is meaningless(imagine you see "deprecation since `v1-someHash`" while you have `vs-someOtherHash`)
3. I've had a few cases where I had to delete deprecation annotation or another case deleting the old renamed method.
4. If we use deprecation here, is bumping minor version a good time to remove them? Or what's appropriate time removing them?

dvoet:
I agree that hashes do not really work as version numbers (we are seeing other problems where sbt evicts the newer version because is comes later in the alphabet).

I am just trying to stick to the guidelines as stated. If the guidelines need to be updated then we should do that.

Even if we don't change the use of hashes in versions, I think a deprecation tag is better than nothing as we should strive to avoid incompatible changes.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -108,11 +115,11 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
   override def storeObject(bucketName: GcsBucketName,
                            objectName: GcsBlobName,
                            objectContents: Array[Byte],
-                           objectType: String,
+                           objectType: String = "text/plain",
```
dvoet, line 118: why add this default?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -229,14 +236,14 @@ object GoogleStorageInterpreter {
 
   def storage[F[_]: Sync: ContextShift](
       pathToJson: String,
-      blockingExecutionContext: ExecutionContext,
+      blockingEc: ExecutionContext,
```
dvoet: why change the name to be less descriptive... we should avoid abbreviations

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: mostly because I haven't seen any other values being used.

dvoet (approved):

PR #249, opened 2019-06-25
Steal enhanced scala config from rawls

helgridly:
No version bump necessary! Yessssss.

**PR checklist**

In all cases:

### DIFF
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/util/EnhancedScalaConfig.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/util/EnhancedScalaConfig.scala
@@ -1,0 +1,32 @@
+package org.broadinstitute.dsde.workbench.util
+
+import com.typesafe.config.{Config, ConfigException, ConfigObject}
+
+import scala.util.{Failure, Success, Try}
+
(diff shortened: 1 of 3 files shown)


MatthewBemis (commented):
`TRAVIS-REPLACE-ME` for the hash?

helgridly:
travis

### replace

# YOU

MatthewBemis (approved):
this might be the nicest code i've ever seen :+1:

helgridly:
doug wrote it originally but thanks

util/src/main/scala/org/broadinstitute/dsde/workbench/util/EnhancedScalaConfig.scala
```diff
@@ -0,0 +1,32 @@
+package org.broadinstitute.dsde.workbench.util
+
+import com.typesafe.config.{Config, ConfigException, ConfigObject}
+
+import scala.util.{Failure, Success, Try}
+
+object ScalaConfig {
```
Qi77Qi, line 7: Personally I don't think this is needed..There're way better config libraries taking care of these things. I don't think this is a custom pattern we should encourage.

PR #253, opened 2019-07-03
Change GCS upload/download methods to rebuild the requests inside the retry

rtitle:
I hope this fixes a situation we saw in Leo auto-tests:

1. We call GCS `storeObject()`, it fails with a 500 "Backend error".
2. GCS call is retried
3. Second try fails on [this line](https://github.com/googleapis/google-api-java-client/blob/master/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpUploader.java#L331) because the `uploadState` is not `NOT_STARTED`.
4. This throws an exception and does not get retried

Apparently for upload/download the GCS request is _mutable_, and carries state. Our code was treating them as immutable objects. This changes it so we rebuild the request for each retry iteration.

The real fix is to have Leo start using `google2` for GCS calls, but that is a bigger change.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,11 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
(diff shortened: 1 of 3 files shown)


MatthewBemis (approved):
Sneaky.

kyuksel (approved):

PR #257, opened 2019-07-11
IA-1046 Support custom storage roles

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -6,7 +6,7 @@ This file documents changes to the `workbench-google2` library, including notes 
 
 Changed
 - Support custom storage IAM roles
-- Remove `project` parameter from `insertBucket` in `GoogleStorageInterpreter`
+- Remove `googleProject` parameter from `insertBucket` in `GoogleStorageInterpreter`
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-TRAVIS-REPLACE-ME"`
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -7,6 +7,7 @@ This file documents changes to the `workbench-google2` library, including notes 
 Changed
 - Support custom storage IAM roles
 - Removed `googleProject` parameter from `insertBucket` in `GoogleStorageInterpreter` as it was not having the desired effect of creating the bucket in the specified project but rather it was configuring requester pays settings for that bucket
+- Add `getIamPolicy`
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-TRAVIS-REPLACE-ME"`
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -6,7 +6,7 @@ This file documents changes to the `workbench-google2` library, including notes 
 
 Changed
 - Support custom storage IAM roles
-- Remove `googleProject` parameter from `insertBucket` in `GoogleStorageInterpreter`
+- Removed `googleProject` parameter from `insertBucket` in `GoogleStorageInterpreter` as it was not having the desired effect of creating the bucket in the specified project but rather it was configuring requester pays settings for that bucket
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-TRAVIS-REPLACE-ME"`
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.6
+
+Changed
+- Support custom storage IAM roles
+- Remove `project` parameter from `insertBucket` in `GoogleStorageInterpreter`
+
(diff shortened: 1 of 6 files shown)


google2/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.6
+
+Changed
+- Support custom storage IAM roles
+- Remove `googleProject` parameter from `insertBucket` in `GoogleStorageInterpreter`
```
Qi77Qi: this is removed because this parameter is not actually doing what we expect it's doing. This project parameter is for requester pays feature

Qi77Qi (approved):

google2/CHANGELOG.md
MatthewBemis: yeah, I guess I should call that out explicitly

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -106,14 +107,14 @@ trait GoogleStorageService[F[_]] {
     * Acl is deprecated. Use setIamPolicy if possible
     */
   @deprecated("Deprecated in favor of insertBucket", "0.5")
-  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]] = None, traceId: Option[TraceId] = None): Stream[F, Unit] = insertBucket(billingProject, bucketName, acl, Map.empty, traceId)
+  def createBucket(billingProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]] = None, traceId: Option[TraceId] = None): Stream[F, Unit] = insertBucket(bucketName, acl, Map.empty, traceId)
```
dvoet: this still has billingProject

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -170,15 +171,15 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
           .build()
     }.getOrElse(bucketInfoBuilder.build())
 
-    val createBucket = blockingF(Async[F].delay(db.create(bucketInfo, BucketTargetOption.userProject(billingProject.value)))).void.handleErrorWith {
+    val createBucket = blockingF(Async[F].delay(db.create(bucketInfo))).void.handleErrorWith {
```
dvoet: so what project is this bucket created in?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
Qi77Qi: if we'll need to remove billingProject from a deprecated method, might as well delete this method?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: it's created in whatever project the google client is initialized in https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala#L238

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
dvoet: you are bumping the version anyway... didn't we say we could remove deprecated functions at that point?

MatthewBemis: yeah i think it should just be deleted...will do

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
dvoet: I think this is a problem as it is counter to how we generally think about and use these services. We think of them as long lived objects that are instantiated once at app start then reused.

Instead of changing the api to remove project and have it passed in when the GoogleStorageService is constructed, I think we need to change the construction to not build the `StorageOptions` immediately. It seems like that is not the google pattern.

Qi77Qi: I don't like how this currently is either. But I don't know if there's a way around it. Even if we can defer `StorageOptions`'s definition, we don't have a way to pass project to https://github.com/broadinstitute/workbench-libs/blob/929a8f7f32c549204b6d8fb82517682bfe7f66cc/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala#L174

dvoet:
Looking at the google client lib it appears as if they have optimized for the case where the user generally operates in the context of a single project so they have take steps so that it does not need to be specified all the time. Unfortunately that is not our use case and we work in lots of projects and our api needs to have it as a mandatory parameter so we are sure to always create buckets in the correct projects.

dvoet (approved):

PR #258, opened 2019-07-12
[CA-387] Refactor implementation of retry functions

helgridly:
`GoogleUtilities.when500orGoogleError` defines the retry behaviour for a lot of our Google calls. The details of the implementation are unclear from the method name, and its ubiquitous usage via `retryWhen500orGoogleError` and `retryWithRecoverWhen500orGoogleError` has caused developers to run into problems.

This PR breaks out the `when500orGoogleError` predicate into a bunch of smaller, composable predicates. It also defines more general `retry` and `retryWithRecover` functions that take a list of predicates, so developers can explicitly define their desired retry behaviour rather than relying on a poorly understood function.

Where this repo used the now-deprecated retry functions, I have replaced them with the equivalent list of explicit predicates in order to preserve functionality. If you think the predicates don't make sense for a given application... well, now you know it's been like that all along!

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -48,6 +48,8 @@ object GoogleUtilities {
     }
 
     def whenNonHttpIOException(throwable: Throwable): Boolean = throwable match {
+      //NOTE Google exceptions are subclasses of IO, so without the two false cases at the top, this would
+      //match on ANY non-2xx Google response.
       case _: GoogleJsonResponseException => false
       case _: GoogleHttpResponseException => false
       case _: IOException => true
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -24,6 +24,7 @@ import scala.util.{Failure, Success, Try}
 trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInstrumented {
   implicit val executionContext: ExecutionContext
 
+  @deprecated(message = "This predicate is complicated and almost certainly doesn't do what you mean. Favor use of retry() and retryWithRecover() with explicitly defined predicates instead.", since = "workbench-google 0.20")
   protected def when500orGoogleError(throwable: Throwable): Boolean = {
     throwable match {
       case t: GoogleJsonResponseException => {
@@ -38,12 +39,49 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
     }
   }
 
+  protected def when500(throwable: Throwable): Boolean = throwable match {
+    case t: GoogleHttpResponseException => t.getStatusCode / 100 == 5
+    case _ => false
+  }
+
+  protected def whenRateLimited(throwable: Throwable): Boolean = throwable match {
+    case t: GoogleJsonResponseException =>
+      (t.getStatusCode == 403 || t.getStatusCode == 429) && t.getDetails.getErrors.asScala.head.getDomain.equalsIgnoreCase("usageLimits")
+    case _ => false
+  }
+
+  protected def when404(throwable: Throwable): Boolean = throwable match {
+    case t: GoogleHttpResponseException => t.getStatusCode == 404
+    case _ => false
+  }
+
+  protected def whenInvalidValueOnBucketCreation(throwable: Throwable): Boolean = throwable match {
+    case t: GoogleJsonResponseException => t.getStatusCode == 400 && t.getDetails.getErrors.asScala.head.getReason.equalsIgnoreCase("invalid")
+    case _ => false
+  }
+
+  @deprecated(message = "This function relies on a complicated predicate that almost certainly doesn't do what you mean. Use retry() with explicitly defined predicates instead.", since = "workbench-google 0.20")
   protected def retryWhen500orGoogleError[T](op: () => T)(implicit histo: Histogram): Future[T] = {
     retryExponentially(when500orGoogleError)(() => Future(blocking(op())))
   }
 
+  private def combine(predicates: Seq[Throwable => Boolean]): (Throwable => Boolean) = { throwable =>
+    predicates.map( _(throwable) ).foldLeft(false)(_ || _)
+  }
+
+  //Retry if any of the predicates return true.
+  protected def retry[T](predicates: (Throwable => Boolean)*)(op: () => T): Future[T] = {
+    retryExponentially(combine(predicates))(() => Future(blocking(op())))
+  }
+
+  @deprecated(message = "This function relies on a complicated predicate that almost certainly doesn't do what you mean. Use retryWithRecover() with explicitly defined predicates instead.", since = "workbench-google 0.20")
   protected def retryWithRecoverWhen500orGoogleError[T](op: () => T)(recover: PartialFunction[Throwable, T])(implicit histo: Histogram): Future[T] = {
     retryExponentially(when500orGoogleError)(() => Future(blocking(op())).recover(recover))
+  }
+
+  //Retry if any of the predicates return true.
+  protected def retryWithRecover[T](predicates: (Throwable => Boolean)*)(op: () => T)(recover: PartialFunction[Throwable, T]) : Future[T] = {
+    retryExponentially(combine(predicates))(() => Future(blocking(op())).recover(recover))
   }
 
   // $COVERAGE-OFF$Can't test Google request code. -hussein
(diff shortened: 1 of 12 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleBigQueryDAO.scala
```diff
@@ -28,7 +28,7 @@ class HttpGoogleBigQueryDAO(appName: String,
   private def submitQuery(projectId: String, job: Job): Future[JobReference] = {
     val queryRequest = bigquery.jobs.insert(projectId, job)
 
-    retryWhen500orGoogleError { () =>
+    retry(when5xx, whenRateLimited, when404, whenInvalidValueOnBucketCreation) { () =>
```
helgridly: If you think this set of predicates is weird... that's what they've been doing all along!

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -21,9 +21,39 @@ import scala.util.{Failure, Success, Try}
...
+        (t.getStatusCode == 403 || t.getStatusCode == 429) && t.getDetails.getErrors.asScala.head.getDomain.equalsIgnoreCase("usageLimits")
+      case _ => false
+    }
+
+    def when404(throwable: Throwable): Boolean = throwable match {
+      case t: GoogleHttpResponseException => t.getStatusCode == 404
+      case _ => false
+    }
+
+    def whenInvalidValueOnBucketCreation(throwable: Throwable): Boolean = throwable match {
```
helgridly, line 45: I had to dig all the way back to https://github.com/broadinstitute/rawls/pull/247 to find the reason this got added. I should probably just call it `whenInvalidValueError` but I wanted to make the original intention clear.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -21,9 +21,46 @@ import scala.util.{Failure, Success, Try}
...
+//These predicates are for use in retries.
+//To use them, import GoogleUtilities.Predicates._
+object GoogleUtilities {
+  object Predicates {
+    def when5xx(throwable: Throwable): Boolean = throwable match {
+      case t: GoogleHttpResponseException => t.getStatusCode / 100 == 5
+      case _ => false
+    }
+
+    def whenRateLimited(throwable: Throwable): Boolean = throwable match {
```
kcibul: I've hit cases where the "domain" piece was not 'usageLimits' and although it was a 429, we weren't retrying.  I would rename this to something that indicates the bizareness of the predicate (which we should keep for backwards compatibility).  Maybe 'when403OrUsageLimitDomainRateLimit'

helgridly: If the domain isn't `usageLimits` then this predicate won't evaluate to true. Do you mentally read HTTP 429 (not necessarily in the Google sense) as "usage limits"? The intention of this predicate is to capture "Google is rate limiting you".

kcibul: right -- but you can also be "rate limited" by receiving a 429 where the domain is "global' (specifically around creating service accounts and keys) and one might naturally a predicate of "whenRateLimited" would retry for that... but because of the extra specificity in this predicate of the "usageLimits" you wouldn't get retried (even though google is rate limiting you).

helgridly: Oh, yikes. TIL, thank you! I'll call it `whenUsageLimited` then, and whoever comes along and wants to retry on global-429s can figure out how they want to compose them.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -21,9 +21,46 @@ import scala.util.{Failure, Success, Try}
 /**
  * Created by mbemis on 5/10/16.
  */
+
+//These predicates are for use in retries.
+//To use them, import GoogleUtilities.Predicates._
+object GoogleUtilities {
+  object Predicates {
```
dvoet: should this be called `RetryPredicates`?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -21,9 +21,46 @@ import scala.util.{Failure, Success, Try}
...
+    }
+
+    def whenInvalidValueOnBucketCreation(throwable: Throwable): Boolean = throwable match {
+      case t: GoogleJsonResponseException => t.getStatusCode == 400 && t.getDetails.getErrors.asScala.head.getReason.equalsIgnoreCase("invalid")
+      case _ => false
+    }
+
+    def whenNonHttpIOException(throwable: Throwable): Boolean = throwable match {
+      case _: GoogleJsonResponseException => false
+      case _: GoogleHttpResponseException => false
```
dvoet, line 54: why do you need the above 2 cases?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
helgridly: sure

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -38,14 +75,30 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
     }
   }
 
+  @deprecated(message = "This function relies on a complicated predicate that almost certainly doesn't do what you mean. Use retry() with explicitly defined predicates instead. There are some useful predicates at the top of GoogleUtilities; try importing GoogleUtilities.Predicates._", since = "workbench-google 0.20")
   protected def retryWhen500orGoogleError[T](op: () => T)(implicit histo: Histogram): Future[T] = {
     retryExponentially(when500orGoogleError)(() => Future(blocking(op())))
   }
 
+  protected def combine(predicates: Seq[Throwable => Boolean]): (Throwable => Boolean) = { throwable =>
+    predicates.map( _(throwable) ).foldLeft(false)(_ || _)
```
rtitle, line 86: This is hard-coded to `or`, but someone might want an `and`? Maybe we could make a more generic `combine` that takes the combination operator.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -38,14 +75,30 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
...
+  }
+
+  //Retry if any of the predicates return true.
+  protected def retry[T](predicates: (Throwable => Boolean)*)(op: () => T)(implicit histo: Histogram): Future[T] = {
```
rtitle, line 90: Minor: the superclass `Retry` declares
```
type Predicate[A] = A => Boolean
```
So instead of `Throwable => Boolean` you could write `Predicate[Throwable]`

project/Settings.scala
```diff
@@ -129,7 +129,7 @@ object Settings {
   val googleSettings = only212 ++ commonSettings ++ List(
     name := "workbench-google",
     libraryDependencies ++= googleDependencies,
-    version := createVersion("0.19"),
+    version := createVersion("0.20"),
```
rtitle, line 132: Reminder to update README.md and CONTRIBUTING.md with the version bump

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
helgridly: `and` seems unlikely to me? since our predicates are generally based around status codes, it wouldn't make sense to construct "retry in the case where the error is 404 AND 500", because that's impossible. people can always roll their own predicates -- i am by no means claiming this is a complete list!

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
helgridly: i noticed this, but i found the explicit return type more readable -- i didn't have to ask myself "what's a predicate again"? i don't mind either way so someone gimme a second opinion

project/Settings.scala
helgridly: yup, working on it :)

dvoet (approved):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
helgridly: Google HTTP exceptions are subclasses of IOException

PR #260, opened 2019-07-19
Add setBucketLabels and fix labels bug in insertBucket

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -160,14 +160,13 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
   }
 
   override def insertBucket(googleProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]] = None, labels: Map[String, String] = Map.empty, traceId: Option[TraceId] = None): Stream[F, Unit] = {
-    val bucketInfoBuilder = BucketInfo.of(bucketName.value).toBuilder
+    val bucketInfoBuilder = BucketInfo.of(bucketName.value).toBuilder.setLabels(labels.asJava)
     val bucketInfo = acl.map{
       aclList =>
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -160,14 +160,13 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
   override def insertBucket(googleProject: GoogleProject, bucketName: GcsBucketName, acl: Option[NonEmptyList[Acl]] = None, labels: Map[String, String] = Map.empty, traceId: Option[TraceId] = None): Stream[F, Unit] = {
-    val bucketInfoBuilder = BucketInfo.of(bucketName.value).toBuilder
+    val bucketInfoBuilder = BucketInfo.of(bucketName.value).toBuilder.setLabels(labels.asJava)
     val bucketInfo = acl.map{
       aclList =>
         val acls = aclList.toList.asJava
         bucketInfoBuilder
           .setAcl(acls)
           .setDefaultAcl(acls)
-          .setLabels(labels.asJava)
```
MatthewBemis, line 170: this would only happen if `acl` was `Some` so i broke it out into the builder above

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -196,6 +195,16 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
     ).void
   }
 
+  override def setBucketLabels(bucketName: GcsBucketName, labels: Map[String, String], traceId: Option[TraceId] = None): Stream[F, Unit] = {
```
MatthewBemis, line 198: i added this because rawls is doing something additional where it is somehow clobbering these later on (i think it might be an incompatibility between google1 and google2 libs). also this seemed like it may be generally useful

Qi77Qi (approved):

PR #264, opened 2019-07-23
Restore non-Stream versions of GoogleKMS* to google1

helgridly:
Previously this code was moved from google1 to google2 and Stream-ified. This meant that in order to integrate minor changes to google1 into Sam, I'd have had to move Sam to use Stream everywhere, which was a lot of work. This PR returns these classes to google1.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,11 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-metric
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.20-9d7f0b6"`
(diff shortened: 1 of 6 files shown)


Qi77Qi:
this looks like straight copy of what's in google2, what's the point of this?

helgridly:
we are currently below the threshold of passing tests that will allow us to release. i have a fix that will reduce test failure rates by around 10%, and it's blocked because i needed to upgrade google1 in sam. suddenly the classes sam was relying on had disappeared, replaced with google2 versions which use streams. it turned out to be a lot of headache to move sam to streams. this PR restores the non-stream version of the classes to google1 so i can get back on track.

Qi77Qi:
I'm still confused..Where does KMS in google use fs2.Stream?

helgridly:
i dunno but i found i was Streaming a LOT of stuff to make it compile

dvoet (approved):

PR #265, opened 2019-07-30
RetryConfig optional parameter on each google storage service function

dvoet:
https://broadworkbench.atlassian.net/browse/QA-761

The retry strategy inadvertently changed when rawls switched to use google2 from creating buckets and setting iam policies. This PR allows retries to be specified per google storage service function call. `RetryPredicates` introduced to make it easier.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.5-3fa06b5"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.5-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -10,7 +10,7 @@ import cats.implicits._
 import com.google.auth.oauth2.ServiceAccountCredentials
 import com.google.cloud.storage.BucketInfo.LifecycleRule
 import com.google.cloud.storage.Storage.{BlobListOption, BlobSourceOption, BlobTargetOption}
-import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, BucketInfo, Storage, StorageException, StorageOptions}
+import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, BucketInfo, Storage, StorageOptions}
 import com.google.cloud.{Identity, Policy, Role}
 import fs2.{Stream, text}
 import io.chrisdavenport.linebacker.Linebacker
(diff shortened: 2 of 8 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -191,87 +177,75 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
       .build()
-    retryStorageF(blockingF(Async[F].delay(db.update(bucketInfo))), traceId, s"com.google.cloud.storage.Storage.update($bucketInfo)").void
+    retryGoogleF(retryConfig)(blockingF(Async[F].delay(db.update(bucketInfo))), traceId, s"com.google.cloud.storage.Storage.update($bucketInfo)").void
   }
 
   private def blockingF[A](fa: F[A]): F[A] = Linebacker[F].blockCS(fa)//ContextShift[F].evalOn(blockingEc)(fa)
 }
 
 object GoogleStorageInterpreter {
-  private val storageRetryableCode = Set(408, 429, 500, 502, 503, 504) //error codes are copied from com.google.cloud.storage.StorageException.RETRYABLE_ERRORS
```
dvoet, line 261: this retry stuff moved to `RetryPredicates` however we don't need this set of codes because the google exception class as an `isRetryable` function that does this check

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -29,23 +29,24 @@ trait GoogleStorageService[F[_]] {
   /**
     * @param traceId uuid for tracing a unique call flow in logging
     */
-  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, isRecursive: Boolean = false, maxPageSize: Long = 1000, traceId: Option[TraceId] = None): Stream[F, GcsObjectName]
+  def listObjectsWithPrefix(bucketName: GcsBucketName, objectNamePrefix: String, isRecursive: Boolean = false, maxPageSize: Long = 1000, traceId: Option[TraceId] = None, retryConfig: RetryConfig = standardRetryConfig): Stream[F, GcsObjectName]
```
dvoet, line 32: every function in this trait as `retryConfig` added

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala
```diff
@@ -0,0 +1,39 @@
...
+    case _ => false
+  }
+
+  def whenStatusCode(code: Int): Throwable => Boolean = {
+    case e: BaseServiceException => e.getCode == code
+    case _ => false
+  }
+
+  def combine(predicates: Seq[Throwable => Boolean]): Throwable => Boolean = { throwable =>
+    predicates.map( _(throwable) ).foldLeft(false)(_ || _)
```
helgridly, line 37: :sunglasses:

helgridly (approved):
ohh, i see. you're allowing each function to use a different `RetryConfig`m instead of the same one defined for the entire storage class

gpolumbo-broad (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
rtitle: Wow I didn't know about `isRetryable` :mindblown:

README.md
```diff
@@ -51,7 +51,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.5-3fa06b5"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.5-TRAVIS-REPLACE-ME"`
```
rtitle, line 54: Should this be a minor version bump? Removing the `RetryConfig` constructor param is backwards-incompatible, right?

dvoet: `GoogleStorageInterpreter` is package private so the constructor is not part of the public api

rtitle: Ah, right

rtitle (approved):
Looks good to me, though Travis is not happy

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreterSpec.scala
```diff
@@ -110,7 +109,8 @@ class GoogleStorageInterpreterSpec extends AsyncFlatSpec with Matchers with Work
     }
   }
 
-  it should "retrieve multiple pages" in ioAssertion {
+  // disabled because the LocalStorageHelper does not seem to support pagination of results
+  ignore should "retrieve multiple pages" in ioAssertion {
```
dvoet, line 113: @rtitle @MatthewBemis In implementing this new retry scheme I fixed [this line of code](https://github.com/broadinstitute/workbench-libs/pull/265/files#diff-945e2ea62787fe3407f871f78faf0ac6R44) to properly pass on values to the nested call which caused this test to start failing. It turns out it was not really doing what it says since the page size was being ignored. But now it seems that the LocalStorageHelper does not do paging anyway (I walked through the google code and I am sure)... so did this ever work? Any insight appreciated.

PR #268, opened 2019-08-09
change traceId to simple String type

Qi77Qi:
Stackdriver generates `X-Cloud-Trace-Context` header, which isn't a UUID. I'd like to use this traceId when it's available.

Example:
```
2019-08-08 20:00:43,295 INFO [ioapp-compute-0] o.h.s.m.Logger [Logger.scala:29] HTTP/1.1 POST /objects/lock Headers(Host: 127.0.0.1:8080, Origin: https://leonardo.dsde-dev.broadinstitute.org, User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.142 Safari/537.36, Accept: */*, Referer: https://leonardo.dsde-dev.broadinstitute.org/notebooks/aou-rw-test-3b102bbe/all-of-us-2278/notebooks/workspaces/qitest/qi-test.ipynb, Accept-Encoding: gzip, deflate, br, Accept-Language: en-US, en;q=0.9, zh-CN;q=0.8, zh;q=0.7, Cookie: <REDACTED>, X-Cloud-Trace-Context: 906e0fc6f601cdad3641d4fff58c6d5f/14103256683195023100, Via: 1.1 google, X-Forwarded-For: 69.173.127.109, 35.241.41.169, 35.191.2.128, 130.211.230.178, X-Forwarded-Proto: https, X-Forwarded-Host: leonardo.dsde-dev.broadinstitute.org, app:8080, X-Forwarded-Server: leonardo.dsde-dev.broadinstitute.org, all-of-us-2278-m.c.aou-rw-test-3b102bbe.internal, Content-Type: text/plain; charset=UTF-8, Connection: close, Content-Length: 47) body="{"localPath":"workspaces/qitest/qi-test.ipynb"}"
```

I think welder is the only app that uses traceId, so I'm inclined to not bump version. But I'm happy to bump it if people prefer.

Blocking https://github.com/DataBiosphere/welder/pull/43

**PR checklist**

In all cases:

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/workbench/model/models.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/workbench/model/models.scala
@@ -2,5 +2,9 @@ package org.broadinstitute.dsde.workbench.model
 
 import java.util.UUID
 
-// uuid for tracing a unique call flow in logging
-final case class TraceId(uuid: UUID) extends AnyVal
+// unique identifier for tracing a unique call flow in logging
+final case class TraceId(asString: String) extends AnyVal
+
+object TraceId {
+  def apply(uuid: UUID): TraceId = new TraceId(uuid.toString)
+}
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" 
 
 Contains generic, externally-facing model classes used across Workbench.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.13-d4e0782"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.13-TRAVIS-REPLACE-ME"`
 
 [Changelog](model/CHANGELOG.md)
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/JsonCodec.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/JsonCodec.scala
@@ -10,5 +10,5 @@ object JsonCodec {
       s"projects/${projectTopicName.getProject}/topics/${projectTopicName.getTopic}"
   )
 
-  implicit val traceIdEncoder: Encoder[TraceId] = Encoder.encodeString.contramap(_.uuid.toString)
+  implicit val traceIdEncoder: Encoder[TraceId] = Encoder.encodeString.contramap(_.asString)
 }
--- a/model/CHANGELOG.md
+++ b/model/CHANGELOG.md
@@ -6,7 +6,7 @@ This file documents changes to the `workbench-model` library, including notes on
 
 - Added generators in `test`
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.13-d4e0782"`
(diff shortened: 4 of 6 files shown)


rtitle (changes requested):
Since it's a non-backwards-compatible change I think I'd feel better about bumping the version (even if we think only welder is using it). Otherwise 👍

notifications/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This file documents changes to the `workbench-notifications` library, including notes on how to upgrade to new versions.
 
+## 0.4
+
+- upgrade `org.broadinstitute.dsde.workbench.model`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.4-TRAVIS-REPLACE-ME"`
```
rtitle: Should say workbench-notifications?

google2/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.6
+
+- upgrade `org.broadinstitute.dsde.workbench.model`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-util" % "0.6-TRAVIS-REPLACE-ME"`
```
rtitle: Should say `workbench-google2`

rtitle (changes requested):
A couple typos in changelogs.. also I did not see CHANGELOG updates for workbench-google and workbench-service-test.

Is it standard practice to bump _all_ modules if making changes to `workbench-model`? I don't think I've done that in the past.

Qi77Qi:
@rtitle in the past, I've always bumped all modules when dependency changed, which I think makes sense, because it breaks those modules bin compatibility(which is why I didn't feel like bumping model in the beginning).

rtitle (approved):

model/src/main/scala/org/broadinstitute/dsde/workbench/model/models.scala
```diff
@@ -1,6 +1,4 @@
 package org.broadinstitute.dsde.workbench.model
 
-import java.util.UUID
-
-// uuid for tracing a unique call flow in logging
-final case class TraceId(uuid: UUID) extends AnyVal
+// unique identifier for tracing a unique call flow in logging
+final case class TraceId(asString: String) extends AnyVal
```
dvoet, line 6: you could provide a companion object with an apply function that takes a UUID and remain backwards compatible.

Qi77Qi: it feels to me too much effort and making code more complicated than necessary given no one else but me has used this field in welder repo

dvoet: This repo is special. It is used by many services. We need to go through extra pains to be backwards compatible. The first rule of making breaking api changes is don't (see https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md).

Qi77Qi: @dvoet , updated. Let me know if you have more comments

Qi77Qi:
@dvoet do you have more comments on this?

dvoet (approved):

PR #273, opened 2019-10-23
Parameterize member type when adding/removing IAM roles

kyuksel:
The changes were motivated by [IA-1364](https://broadworkbench.atlassian.net/browse/IA-1364) where we need to add an IAM role to a (Google) `group` (as opposed to a `user` or a `serviceAccount`). While at it, also refactored some of the methods to make them generic but still backwards-compatible.

**PR checklist**

In all cases:
- [x] Verify all CI tests go green

### DIFF
--- a/google/CHANGELOG.md
+++ b/google/CHANGELOG.md
@@ -2,25 +2,16 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
-## 0.22
+## 0.21
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.22-TRAVIS-REPLACE-ME"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.21-TRAVIS-REPLACE-ME"`
 
 ### Added
 
 - `addIamRoles` and `removeIamRoles` in `GoogleIamDAO`:
-  - These methods now accept a member type of the newly created ADT `MemberType`.
+  - These methods accept a member type of the newly created ADT `MemberType`.
   - The now deprecated `addIamRolesForUser` and `removeIamRolesForUser` call the aforementioned methods
   for backwards compatibility.
-
-### Deprecated
-
-- `addIamRolesForUser` and `removeIamRolesForUser` are deprecated in favour of the more generic versions: 
-`addIamRoles` and `removeIamRoles`. 
-
-## 0.21
-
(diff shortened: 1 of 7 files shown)


google/CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
 
 This file documents changes to the `workbench-google` library, including notes on how to upgrade to new versions.
 
+## 0.22
```
kyuksel: Minor version bump due to the deprecation

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -134,4 +166,24 @@ trait GoogleIamDAO {
...
+    final case object User extends MemberType {
+      override def toString = "user"
+    }
+    final case object Group extends MemberType {
+      override def toString = "group"
+    }
+    final case object ServiceAccount extends MemberType {
+      override def toString = "serviceAccount"
+    }
+    final case class Other(memberType: String) extends MemberType {
```
Qi77Qi: do we actually need `Other`?

kyuksel: We don't. I just wanted to provide flexibility forward looking. It looks like currently the only other valid `MemberType` is `domain`. I'll add that and remove `Other`.

Qi77Qi:
do we actually need to bump the version now that it doesn't really break BC?

google/CHANGELOG.md
dvoet: Deprecation does not require version bump

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -80,7 +82,10 @@ trait GoogleIamDAO {
     * @param rolesToAdd Set of roles to add (example: roles/storage.admin)
     * @return true if the policy was updated; false otherwise.
     */
-  def addIamRolesForUser(iamProject: GoogleProject, email: WorkbenchEmail, rolesToAdd: Set[String]): Future[Boolean]
```
Qi77Qi, line 83: version

kyuksel: thanks for the catch!

Qi77Qi (approved):

dvoet (approved):

PR #280, opened 2020-02-03
[no ticket] add traceId in published messages

Qi77Qi:
1. add `traceId` in published messages in publisher so that we can easily correlate messages between publisher and subcriber
2. extract `traceId` and log it in subscriber
3. Use `StructuredLogger` instead of `Logger`...(this shouldn't cause much issue for caller since `Slf4jLogger.getLogger[IO]` is a `StructuredLogger`, which takes advantage of http://logback.qos.ch/manual/mdc.html..... Benefit of using mdc is we'll be able to query log more easily. We can do things like `request.traceId` for instance (definitely in stackdriver, additional things may needed for kibana). To see mdc in logs, apps may need to update logback config to reflect it. One example is https://github.com/DataBiosphere/welder/blob/master/server/src/main/resources/logback.xml#L9, and add `"net.logstash.logback" % "logstash-logback-encoder" % "6.2" // for structured logging in logback` in dependencies
4. Fix an issue where if there's an exception is thrown in receiver, error not get lost in subscriber.

**PR checklist**

In all cases:

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -19,12 +19,17 @@ Changed
 - Remove `LineBacker` usage
 - Add arguments to `insertBucket`
 - Fix `scala.MatchError` from `handleErrorWith`
+<<<<<<< HEAD
+=======
+- Add `delete` function to `GoogleTopicAdmin` trait and implementation
+- Use `recoverWith` instead of `onError` which doesn't actually recover the error
+>>>>>>> 8956707... fix wrong use of onError
 
 Add
 - Add `GoogleDataproc` and `GoogleDataprocInterpreter`
 - Add `delete` function to `GoogleTopicAdmin` trait and implementation
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-c91d96b"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-TRAVIS-REPLACE-ME"`
 
 ## 0.5
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.7
+Changed
+- Use `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
+
+Add
+- Add `tracedPublish` to `GooglePublisher[F]`
+- Log messages with traceId in `GoogleSubscriberInterpreter`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.7-TRAVIS-REPLACE-ME"`
+
 ## 0.6
 Changed
 - Bump `fs2-io` to `2.0.1`
@@ -13,12 +23,9 @@ Changed
 
 Add
 - Add `GoogleDataproc` and `GoogleDataprocInterpreter`
-- Add `tracedPublish` to `GooglePublisher[F]`
-- Log messages with traceId in `GoogleSubscriberInterpreter`
 - Add `delete` function to `GoogleTopicAdmin` trait and implementation
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-c91d96b"`
-
 
 ## 0.5
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,16 +2,6 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
-## 0.7
-Changed
-- Use `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
-
-Add
-- Add `tracedPublish` to `GooglePublisher[F]`
-- Log messages with traceId in `GoogleSubscriberInterpreter`
-
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.7-TRAVIS-REPLACE-ME"`
-
 ## 0.6
 Changed
 - Bump `fs2-io` to `2.0.1`
@@ -19,15 +9,15 @@ Changed
 - Remove `LineBacker` usage
 - Add arguments to `insertBucket`
 - Fix `scala.MatchError` from `handleErrorWith`
-<<<<<<< HEAD
-=======
 - Add `delete` function to `GoogleTopicAdmin` trait and implementation
 - Use `recoverWith` instead of `onError` which doesn't actually recover the error
->>>>>>> 8956707... fix wrong use of onError
 
 Add
 - Add `GoogleDataproc` and `GoogleDataprocInterpreter`
 - Add `delete` function to `GoogleTopicAdmin` trait and implementation
+- Add `publishNative` to `GooglePublisher[F]` so that user can add attributes easily
+- Log messages with traceId in `GoogleSubscriberInterpreter`
+- Add `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.6-TRAVIS-REPLACE-ME"`
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -19,7 +19,6 @@ Changed
 - Remove `LineBacker` usage
 - Add arguments to `insertBucket`
 - Fix `scala.MatchError` from `handleErrorWith`
-- Use `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
 
 Add
 - Add `GoogleDataproc` and `GoogleDataprocInterpreter`
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -9,10 +9,13 @@ Changed
 - Remove `LineBacker` usage
 - Add arguments to `insertBucket`
 - Fix `scala.MatchError` from `handleErrorWith`
-- Add `delete` function to `GoogleTopicAdmin` trait and implementation
+- Use `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
 
 Add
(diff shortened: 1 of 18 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataproc.scala
```diff
@@ -30,22 +30,22 @@ trait GoogleDataproc[F[_]] {
...
     credentialFile <- org.broadinstitute.dsde.workbench.util2.readFile(pathToCredential)
     client <- fromServiceAccountCrendential(ServiceAccountCredentials.fromStream(credentialFile), blocker, blockerBound, retryConfig)
   } yield client
 
-  def fromServiceAccountCrendential[F[_]: Logger: Async: Timer: ContextShift](serviceAccountCredentials: ServiceAccountCredentials, blocker: Blocker, blockerBound: Semaphore[F], retryConfig: RetryConfig = GoogleTopicAdminInterpreter.defaultRetryConfig): Resource[F, GoogleDataproc[F]] = {
+  def fromServiceAccountCrendential[F[_]: StructuredLogger: Async: Timer: ContextShift](serviceAccountCredentials: ServiceAccountCredentials, blocker: Blocker, blockerBound: Semaphore[F], retryConfig: RetryConfig = GoogleTopicAdminInterpreter.defaultRetryConfig): Resource[F, GoogleDataproc[F]] = {
     val settings = ClusterControllerSettings.newBuilder()
-      .setCredentialsProvider(FixedCredentialsProvider.create(serviceAccountCredentials))
+      .setCredentialsProvider(FixedCredentialsProvider.create(serviceAccountCredentchials))
```
jdcanas: typo

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
```diff
@@ -11,6 +11,11 @@ trait GooglePublisher[F[_]] {
     */
   def publish[MessageType: Encoder]: Pipe[F, MessageType, Unit]
 
+  /**
+    * Watch out message size quota and limitations https://cloud.google.com/pubsub/quotas
+    */
+  def tracedPublish[MessageType: Encoder]: Pipe[F, Message[MessageType], Unit]
```
jdcanas: does this really need to be another method in the interface? I might be missing something, but It seems like this could with an optional parameter, and then it would still be a non-breaking change.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -65,25 +65,36 @@ object GoogleSubscriberInterpreter {
...
-        json <- parse(message.getData.toStringUtf8)
-        message <- json.as[MessageType]
-        _ <- queue.enqueue1(Event(message, consumer)).attempt.toIO.unsafeRunSync()
+        json <- Effect[F].fromEither(parse(message.getData.toStringUtf8))
+        message <- Effect[F].fromEither(json.as[Message[MessageType]])
+        r <- queue.enqueue1(Event(message.traceId, message.msg, consumer)).attempt
+        loggingCtx = Map("traceId" -> message.traceId.map(_.asString).getOrElse(""))
+        _ <- r match {
+          case Left(e) =>
+            StructuredLogger[F].info(loggingCtx)(s"failed to enqueue $message to internal Queue due to $e") >> Effect[F].delay(consumer.nack()) //pubsub will resend the message up to ackDeadlineSeconds (this is configed during subscription creation
```
jdcanas: nit: does this comment reference the correct field in the RetryConfig?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -65,25 +65,36 @@ object GoogleSubscriberInterpreter {
...
+          case Right(_) =>
+            StructuredLogger[F].info(loggingCtx)(s"successfully to enqueue $message")
```
jdcanas: something like 'successfully enqueued $message' reads better imo

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -70,4 +71,4 @@ package object google2 {
 }
 
 final case class RetryConfig(retryInitialDelay: FiniteDuration, retryNextDelay: FiniteDuration => FiniteDuration, maxAttempts: Int, retryable: Throwable => Boolean = scala.util.control.NonFatal.apply)
-final case class LoggableGoogleCall(traceId: Option[TraceId], googleCall: String, duration: FiniteDuration, response: Option[String], result: String)
\ No newline at end of file
+final case class LoggableGoogleCall(response: Option[String], result: String)
```
jdcanas, line 79: Result seems to be a case class with two options, 'success' or 'failure'. I'd either refactor it, or, if you wish to keep the encoder/decoder simple, add a comment here indicating its current usage.

jdcanas (changes requested):
I don't think this compiles due to https://github.com/broadinstitute/workbench-libs/blob/5f6963ecf42d1e1fa6f89361c4af6bdf11dad795/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataproc.scala#L40

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
Qi77Qi: we will need a separate function, because `Pipe[F[_], I, O]` is just type alias of `Stream[F, I] => Stream[F, O]`, what we want is `Stream[F, Message] => Stream[F, Unit]`, not `Message => Stream[F, I] => Stream[F, Unit]`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
Qi77Qi: are you referring to a different line?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
Qi77Qi: I don't think it's needed to make this ADT since it's just for logging and no caller elsewhere will be consuming the info

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
jdcanas: In this comment, `//pubsub will resend the message up to ackDeadlineSeconds` , it seems to be referring to the `maxAttempts` field as opposed to the `ackDeadlineSeconds` field based on the wording.

jdcanas (approved):

google2/CHANGELOG.md
```diff
@@ -2,21 +2,31 @@
...
+
 ## 0.6
 Changed
 - Bump `fs2-io` to `2.0.1`
 - Add optional `blockderBound` to `GoogleStorageService` constructor
 - Remove `LineBacker` usage
 - Add arguments to `insertBucket`
 - Fix `scala.MatchError` from `handleErrorWith`
-- Add `delete` function to `GoogleTopicAdmin` trait and implementation
+- Use `io.chrisdavenport.log4cats.StructuredLogger` instead of `io.chrisdavenport.log4cats.Logger`
```
rtitle: Remove this from 0.6?

rtitle (approved):
Lgtm

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -27,6 +29,20 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: Timer: Logger](
...
 
+  def tracedPublish[MessageType: Encoder]: Pipe[F, DecoratedMessage[MessageType], Unit] = in => {
+    in.flatMap {
+      message =>
+      for {
+        now <- Stream.eval(Timer[F].clock.realTime(TimeUnit.MILLISECONDS))
+        extraInfo = message.traceId.fold[Map[String, String]](Map.empty)(tid => Map ("traceId" -> tid.asString)) ++ Map(
+          "publishedInMillis" -> now.toString
+        )
+        msg = message.msg.asJson.deepMerge(extraInfo.asJson) //decorate the message with extra info
```
dvoet: Embedding the extra information in message itself seems problematic to me because it means the consumer must deal with it. It also eats into the the max message size (granted the max message size is 10MB and this little bit of data probably won't be noticed).

Did you consider user message attributes instead? Also the publish time is automatically included in the message by Google. See https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -65,25 +67,47 @@ object GoogleSubscriberInterpreter {
...
+
       val result = for {
-        json <- parse(message.getData.toStringUtf8)
-        message <- json.as[MessageType]
-        _ <- queue.enqueue1(Event(message, consumer)).attempt.toIO.unsafeRunSync()
+        res <- parseEvent.attempt
+        _ <- res match {
+          case Right(event) =>
+            for {
+              r <- queue.enqueue1(event).attempt
```
rtitle, line 93: So do we think this error handling will resolve the message-not-received issue?

Qi77Qi:
@rtitle , I think previously the message was received. It's just the error was thrown and somehow didn't show in log. I see error log when this happens in my manual test https://github.com/broadinstitute/workbench-libs/pull/280/files#diff-161d25ff3e7b79c4cd2e1c5deca0027aR49 ....But in leo, error was never shown when I publish bad messages to the topic. After this PR, we'll see errors.

kyuksel:
Curious if you found out anything that may explain seeing the issue in dev and not Justin's fiab  🤔

Qi77Qi:
@kyuksel needs a little more testing..but I think https://github.com/DataBiosphere/leonardo/pull/1271 is the real fix why fiab works but not in dev...it's user error

Qi77Qi:
talked to @dvoet this morning...I think he's okay with merging

PR #281, opened 2020-02-05
Add Rawls submission option for delete_intermediate_output_files WA-98

kshakir:
**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/CHANGELOG.md
+++ b/serviceTest/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 This file documents changes to the `workbench-service-test` library, including notes on how to upgrade to new versions.
 
+## 0.17
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.17-TRAVIS-REPLACE-ME"`
+
+###
+
+### Added
+
+- add `deleteIntermedateOutputFiles` parameter to `Rawls.submissions.launchWorkflow`
+
 ## 0.16
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.16-20dceaa"`
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
@@ -21,6 +21,7 @@ trait Rawls extends RestClient with LazyLogging {
     mapper.readValue(response, classOf[List[Map[String, T]]])
   }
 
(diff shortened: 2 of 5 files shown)


kshakir:
Manually run swatomation pipeline:
✅https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/30218/

serviceTest/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 This file documents changes to the `workbench-service-test` library, including notes on how to upgrade to new versions.
 
+## 0.17
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.17-TRAVIS-REPLACE-ME"`
```
aednichols, line 7: Huh? Does Travis replace strings in the CHANGELOG?

aednichols (approved):

serviceTest/CHANGELOG.md
kshakir: Yep. It's even referred to in the pull-request-template message.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -60,6 +61,7 @@ trait Rawls extends RestClient with LazyLogging {
 
   }
 
+  //noinspection RedundantBlock,ScalaUnnecessaryParentheses
```
salonishah11, line 64: What does this do?

kshakir: It shuts off warnings about these issues in the file. It then allows other bugs to be quickly discovered.

- These warnings are enabled by default in IntelliJ IDEA
- This switches them off only for this file
- Doesn't muddle this PR with other fixes for now
- Suppression can be removed easily and issues cleaned up in a later PR
- Unmasks more important inspections that catch bugs like `if (myString == myOption)` which will never match
- https://www.jetbrains.com/help/idea/code-inspection.html

In this case, this class contains extra `{` and extra `(` that don't need to be there.

salonishah11 (approved):
I prefer `delete_intermediate_output_files` over `delete_intermediates` because `delete_intermediate_output_files` suggests that we are **only** deleting the intermediate output files (which is the current implementation) and leaving files like stderr, stdout, rc, transfer scripts, etc. created by Cromwell as it is in the bucket. While `delete_intermediates` suggests that we are deleting everything except final outputs. That being said, I am not married to it being called `delete_intermediate_output_files` and am open to it being renamed :)

PR #287, opened 2020-03-10
[IA-1718][IA-1748] GKE and Kubernetes client 

jdcanas:
This PR
- adds a google service to interact with GKE
   - can create Clusters/Nodepools
- Adds a service to interact with kubernetes API
   - Allows creation of a namespace
   - Allows creation of a service (which allows containers to be exposed externally to the cluster)
   - Allows creation of a Pod with a set of containers on the Pod 
   - Allows mapping between a service and a pod
   - lays the foundation for unit testing the above to via a test object
   - Manages multiple clients and auth for a multi-cluster scenario

TODO list:
- make snapshot


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
```diff
@@ -0,0 +1,100 @@
...
+
+trait GoogleKubernetesService[F[_]] {
+  def createCluster(kubernetesClusterRequest: KubernetesCreateClusterRequest): F[Operation]
+
+  def deleteCluster(clusterId: KubernetesClusterIdentifier): F[Operation]
+
+  def getCluster(clusterId: KubernetesClusterIdentifier): F[Cluster]
+}
+//TODO: migrate to a unit test
+object Test {
```
jdcanas: this is mainly for local testing of these APIs. Should be migrated into unit tests before merging.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,58 @@
...
+  def resource[F[_]: StructuredLogger : Async : Timer : ContextShift](
+                                                                        pathToCredential: String,
+                                                                        //TODO: preferably, this should be set via KUBECONFIG env var, which  See https://cloud.google.com/sdk/gcloud/reference/container/clusters/get-credentials
+                                                                        pathToKubeConfig: String,
+                                                                        clusterIds: Set[KubernetesClusterIdentifier],
+                                                                        retryConfig: RetryConfig = RetryPredicates.standardRetryConfig
+                                                                     ): Resource[F,KubernetesService[F]] =
+    for {
+      _ <- Resource.liftF(performGKEtoKubernetesCLIAuthHandoff(pathToCredential, clusterIds))
+    fileReader <- Resource.make(
```
jdcanas: I really don't like surrounding this as a resource because it will never get released, and it can be closed before the KubernetesInterpreter is even created...

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
Qi77Qi: things for test shouldn't live in main code...you can move this to `test` and it'll still be available to applications if they decide to include it

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,207 @@
...
+import io.kubernetes.client.proto.V1.ResourceRequirements
+import org.broadinstitute.dsde.workbench.model.WorkbenchException
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+
+object KubernetesConstants {
+  //this default namespace is initialized automatically with a kubernetes environment, and we do not create it
+  val DEFAULT_NAMESPACE = KubernetesNamespace(KubernetesNamespaceName("default"))
+  //this is a default service you must create in your environment. This is mostly used for testing, and should possibly be initialized with createCluster if you wish to use it.
+  val DEFAULT_SERVICE_NAME = KubernetesServiceName("default-service")
```
jdcanas: this 'default_service' is used for pods that don't need any user-specific networking, and we just wish to expose to the internet

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,48 @@
...
+import io.chrisdavenport.log4cats.StructuredLogger
+import io.kubernetes.client.apis.CoreV1Api
+import org.broadinstitute.dsde.workbench.RetryConfig
+import cats.implicits._
+import io.kubernetes.client.ApiException
+
+// This uses a kubernetes client library to make calls to the kubernetes API. The client library is autogenerated from the kubernetes API.
+// It is highly recommended to use the kubernetes API docs here https://kubernetes.io/docs/reference/generated/kubernetes-api as opposed to the client library documentation
+
+class KubernetesInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
```
Qi77Qi: shall we abbreviate to `GkeInterpreter`? `KubernetesInterpreter` is really long

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
```diff
@@ -0,0 +1,100 @@
...
+  val project = GoogleProject("broad-dsde-dev")
+  val location =  Location("us-central1")
+  val parent = Parent(project, location)
+  val clusterName = KubernetesClusterName("c2")
+  val nodePoolName = NodePoolName("nodepool1")
+
+  val cluster2Id = KubernetesClusterIdentifier(project, location, clusterName)
+
+  val serviceResource = GoogleKubernetesService.resource("/Users/jcanas/Downloads/kube-broad-dsde-dev-key.json")
+  val kubeService = KubernetesService.resource("/Users/jcanas/Downloads/kube-broad-dsde-dev-key.json", "/Users/jcanas/.kube/config", Set(cluster2Id))
```
jdcanas: if you want to run this locally, you should only have to change the above hardcoded strings, and pick a `clusterName` that is not already created. Ping me if you want the service account key, its a new one I created for this.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,58 @@
...
+  //pods represent a set of containers
+  def createPod(pod: KubernetesPod, namespace: KubernetesNamespace = KubernetesConstants.DEFAULT_NAMESPACE): F[Unit]
+
+  //certain services allow us to expose various containers via a matching selector
+  def createService(service: KubernetesServiceKind, namespace: KubernetesNamespace = KubernetesConstants.DEFAULT_NAMESPACE): F[Unit]
+}
+
+object KubernetesService {
+  def resource[F[_]: StructuredLogger : Async : Timer : ContextShift](
+                                                                        pathToCredential: String,
```
Qi77Qi: I know you copied from previous code, but these paths should really be `java.nio.file.Path`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,58 @@
...
+      Async[F].delay(new FileReader(pathToKubeConfig))
+    )(reader => Async[F].delay(reader.close()))
+    apiClient <- Resource.liftF(Async[F].delay(ClientBuilder.kubeconfig(
+      KubeConfig.loadKubeConfig(fileReader)
+    ).build()))
+    _ = Configuration.setDefaultApiClient(apiClient)
+    kubernetesApi = new CoreV1Api()
+  } yield new KubernetesInterpreter(kubernetesApi, retryConfig)
+
+  private def performGKEtoKubernetesCLIAuthHandoff[F[_] : Async](pathToCredential: String,  clusterIds: Set[KubernetesClusterIdentifier]): F[Unit] =
```
Qi77Qi: what is this function for?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/ReplacedGCPAuthenticator.java
```diff
@@ -0,0 +1,78 @@
...
+import org.slf4j.LoggerFactory;
+
+
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Date;
+import java.util.Map;
+
+// this class implements io.kubernetes.client.util.authenticators.GCPAuthenticator
+// It is taken from https://github.com/kubernetes-client/java/issues/290#issuecomment-480205118
```
jdcanas: this isn't used currently, but its here in case we find it relevant in the future

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,74 @@
...
+import io.chrisdavenport.log4cats.StructuredLogger
+import org.broadinstitute.dsde.workbench.RetryConfig
+import io.kubernetes.client.apis.CoreV1Api
+import io.kubernetes.client.Configuration
+import io.kubernetes.client.util.{ClientBuilder, Config, KubeConfig}
+import org.broadinstitute.dsde.workbench.google2.util.{ReplacedGCPAuthenticator, RetryPredicates}
+
+trait KubernetesService[F[_]] {
+  //namespaces group resources, and allow our list/get/update API calls to be segmented. This can be used on a per-user basis, for example
+  def createNamespace(namespace: KubernetesNamespace): F[Unit]
```
jdcanas: the Kubernetes client is a bit simple and just returns the java object you give it to create, which I found not very useful, so these just return Unit

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+class GoogleKubernetesInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
+                                         clusterManagerClient: ClusterManagerClient,
+                                         blocker: Blocker,
+                                         blockerBound: Semaphore[F],
+                                         retryConfig: RetryConfig
+                                       ) extends GoogleKubernetesService[F] {
+
+   override def createCluster(kubernetesClusterRequest: KubernetesCreateClusterRequest)
```
rtitle: Would it be cleaner for this method to take a `(project: GoogleProject, zone: ZoneName, cluster: com.google.container.v1.Cluster)`?

I feel like we don't need to rebuild all the input fields in `KubernetesCreateClusterRequest` -- the caller can be responsible for creating their own `Cluster` object. This is the pattern that the GCE service/interpreter uses.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
```diff
@@ -0,0 +1,117 @@
...
+import io.chrisdavenport.log4cats.StructuredLogger
+import org.broadinstitute.dsde.workbench.RetryConfig
+import org.broadinstitute.dsde.workbench.google2.KubernetesConstants.DEFAULT_LOADBANCER_PORTS
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates._
+
+trait GoogleKubernetesService[F[_]] {
+  //should clusters be created per project, or per user for billing?
```
rtitle: In GKE clusters are per (project, location), so the wb-libs API should reflect that.

https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters/create

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
rtitle: +1 should move to a unit test

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,84 @@
...
+import io.kubernetes.client.apis.CoreV1Api
+import io.kubernetes.client.Configuration
+import io.kubernetes.client.util.{ClientBuilder, Config, KubeConfig}
+import org.broadinstitute.dsde.workbench.google2.util.{ReplacedGCPAuthenticator, RetryPredicates}
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+trait KubernetesService[F[_]] {
+  //namespaces group resources, and allow our list/get/update API calls to be segmented. This can be used on a per-user basis, for example
+  def createNamespace(namespace: KubernetesNamespace)
+  (implicit ev: ApplicativeAsk[F, TraceId]): F[Unit]
```
rtitle: Might be good to run `scalafmt` on this :)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,59 @@
...
+       _ <- Async[F].delay(kubernetesClient.createNamespace(namespace.getJavaSerialization, null, null, null))
+      } yield ()
+    )
+
+  //TODO: retry once we know what kubernetes codes are applicable
+  private def kubernetesCallWrapper[A](fa: F[A]): F[A] =
+    blockerBound.withPermit(blocker.blockOn(
+      fa
+    ).onError { //we aren't handling it here, it will be bubbled up, but we want to print a more helpful message
+      case e: ApiException => Async[F].delay(println(e.getResponseBody()))
```
rtitle: Maybe add a `Logger` context bound and use that instead of `println`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,59 @@
...
+                                                                                 blocker: Blocker,
+                                                                                 blockerBound: Semaphore[F],
+                                                                                 retryConfig: RetryConfig
+                                                                               ) extends KubernetesService[F] {
+
+  // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podspec-v1-core
+  override def createPod(pod: KubernetesPod, namespace: KubernetesNamespace)
+                        (implicit ev: ApplicativeAsk[F, TraceId]): F[Unit] =
+    kubernetesCallWrapper(
+      for {
```
rtitle: Minor, you can omit this `for`, this is equivalent:
```
Async[F].delay(kubernetesClient.createNamespacedPod(namespace.name.value, pod.getJavaSerialization, null, null, null)).void
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,84 @@
...
+        KubeConfig.loadKubeConfig(fileReader)
+      ).build()))
+      _ = Configuration.setDefaultApiClient(apiClient)
+      kubernetesApi = new CoreV1Api()
+  } yield new KubernetesInterpreter(kubernetesApi, blocker, blockerBound, retryConfig)
+
+  //doesn't work for multi-cluster or multi-project configs
+  //doesn't refresh every hour
+  //use this to get the arg for this function credentials <- GoogleServiceHttpInterpreter.credentialResourceWithScope(pathToCredential)
+  private def getKubernetesApiWithGoogleCreds[F[_] : Async](credentials: GoogleCredentials): F[CoreV1Api] = {
```
rtitle: Is this used anywhere?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/ReplacedGCPAuthenticator.java
rtitle: Probably want to remove this before merge

rtitle:
Neat. There's still some cleanup to do before this is mergeable, but the overall structure looks good. I'm going to try and play around with it myself tonight, in particular the auth piece to try and better understand what's going on

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
jdcanas: I agree, I changed it. I kept some of the nodepool-related objects because they make some method signatures cleaner, but I'm open to discussion on that.

On the other hand, I feel it is still valuable to provide an easy way to get a 'defaultCluster', so I moved some of the logic to a function that takes the minimum necessary user input and creates a `Cluster` with defaults.. Clusters are pretty complex, and it doesn't sound like we have any requirements to allow users to tune their cluster specs, so in an ideal world I would like all the guts of the objects used by in the google package hidden away in wb-libs

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
```diff
@@ -0,0 +1,54 @@
...
+      .setCluster(kubernetesClusterRequest.cluster)
+      .build()
+
+    tracedGoogleRetryWithBlocker(
+      Async[F].delay(clusterManagerClient.createCluster(createClusterRequest)),
+      f"com.google.cloud.container.v1.ClusterManagerClient.createCluster(${kubernetesClusterRequest})"
+    )
+  }
+
+  override def getCluster(clusterId: KubernetesClusterId)(implicit ev: ApplicativeAsk[F, TraceId]): F[Cluster] =
```
rtitle: We often use the pattern of `getXxx` returning an `Option`, so it can return `None` instead of throw 404 when the object doesn't exist. For example here's the GCE `getInstance` implementation:
```
override def getInstance(project: GoogleProject, zone: ZoneName, instanceName: InstanceName)(
    implicit ev: ApplicativeAsk[F, TraceId]
  ): F[Option[Instance]] = {
    val projectZoneInstanceName = ProjectZoneInstanceName.of(instanceName.value, project.value, zone.value)
    retryF(
      recoverF(
        Async[F].delay(instanceClient.getInstance(projectZoneInstanceName)),
        whenStatusCode(404)
      ),
      s"com.google.cloud.compute.v1.InstanceClient.getInstance(${projectZoneInstanceName.toString})"
    )
  }
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.broadinstitute.dsde.workbench.RetryConfig
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates._
+import cats.implicits._
+
+trait GoogleKubernetesService[F[_]] {
+  def createCluster(kubernetesClusterRequest: KubernetesCreateClusterRequest)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[Operation]
```
rtitle: This may be for a future PR, but for GCE I added some `Operation`-polling logic to `GoogleComputeService/Interpreter`. GCE and GKE have separate `Operation` objects, but maybe some of the polling logic could be reused/abstracted and used here too.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesService.scala
```diff
@@ -0,0 +1,49 @@
...
+    retryConfig: RetryConfig = RetryPredicates.retryConfigWithPredicates(whenStatusCode(404))
+  ): Resource[F, GoogleKubernetesService[F]] =
+    for {
+      credential <- credentialResource(pathToCredential)
+      credentialsProvider = FixedCredentialsProvider.create(credential)
+      clusterManagerSettings = ClusterManagerSettings
+        .newBuilder()
+        .setCredentialsProvider(credentialsProvider)
+        .build()
+      clusterManager <- Resource.make(
```
rtitle: I think I added a utility for creating these clients -- could this be 
```
clusterManager <- resourceF(ClusterManagerClient.create(clusterManagerSettings))
```
?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import scala.collection.JavaConverters._
+import com.google.api.services.container.ContainerScopes
+import io.chrisdavenport.log4cats.StructuredLogger
+import org.broadinstitute.dsde.workbench.RetryConfig
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+trait KubernetesService[F[_]] {
+  //namespaces group resources, and allow our list/get/update API calls to be segmented. This can be used on a per-user basis, for example
```
rtitle: for example what? haha don't leave me hanging

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,46 @@
...
+  def createNamespace(clusterId: KubernetesClusterId, namespace: KubernetesNamespace)(
```
rtitle: Eventually will we need getters for namespaces/pods/service?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -0,0 +1,46 @@
...
+    pathToCredential: String,
+    googleKubernetesService: GoogleKubernetesService[F],
+    blocker: Blocker,
+    blockerBound: Semaphore[F],
+    //This is not used anywhere yet, there should be a custom kube one
+    retryConfig: RetryConfig = RetryPredicates.standardRetryConfig
+  ): Resource[F, KubernetesService[F]] =
+    for {
+      credentials <- credentialResource(pathToCredential)
+      scopedCredential = credentials.createScoped(Seq(ContainerScopes.CLOUD_PLATFORM).asJava)
```
rtitle, line 49: Minor - are there more restrictive scopes we could use instead of `CLOUD_PLATFORM`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+  private def getClient(clusterId: KubernetesClusterId)(implicit ev: ApplicativeAsk[F, TraceId]): F[CoreV1Api] = {
+    val clientOpt = clientStore.get(clusterId)
+
+    //we always update the token, even for existing clients, so we don't have to maintain a reference to the last time each client was updated
+    //unfortunately, the kubernetes client does not implement a gcp authenticator, so we must do this ourselves.
+    //See this for details https://github.com/kubernetes-client/java/issues/290
+    credentials.refreshIfExpired()
+    val token = credentials.getAccessToken
+
+    clientOpt match {
```
rtitle: I was wondering if a guava cache with a loading function could be used for this… but this implementation seems fine I think.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+
+class KubernetesInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
+  credentials: GoogleCredentials,
+  googleKubernetesService: GoogleKubernetesService[F],
+  blocker: Blocker,
+  blockerBound: Semaphore[F],
+  retryConfig: RetryConfig
+) extends KubernetesService[F] {
+  //here, we store the apiClient so we do not have to make the google calls necessary to construct it more often than needed
+  private val clientStore = TrieMap[KubernetesClusterId, ApiClient]()
```
rtitle: This is a nice of use of state, and an elegant way of avoiding shelling out to connect with k8s.

I was debating with myself whether it's better to keep the state here in wb-libs code, or make the caller responsible for handling its own state. Generally wb-libs is stateless but in this case I think it's reasonable to have the state here because of the low-level details necessary for creating a k8s client (getting the certificate, token, etc).

So 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,252 @@
...
+          .setMinNodeCount(config.autoscalingConfig.minimumNodes)
+          .setMaxNodeCount(config.autoscalingConfig.maximumNodes)
+      )
+}
+
+// Common kubernetes models //
+
+final case class KubernetesException(message: String) extends WorkbenchException
+
+trait KubernetesNameValidation {
```
rtitle: I don't really like using a trait for this kind of validation.. would prefer something like:
```
final case class KubernetesName(value: String) extends AnyVal
object KubernetesName {
  def fromString(str: String): KubernetesName = /* do regex validation */
}
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,252 @@
...
+  if (!isValidName) {
+    //the throwing of this exception here is assuming that leo will be the one controlling nodepool names, not users, and is to sanitize the data in this case class at construction as opposed to downstream
+    throw KubernetesException(s"The name ${value} must match the regex ${regex}")
+  }
+}
+
+// Google kubernetes client models //
+
+//"us-central1" is an example of a valid location.
+final case class Parent(project: GoogleProject, location: Location) {
```
rtitle: We have `RegionName` as well which I feel is more specific than `Location` -- if this should represent a region then could we use `RegionName`?

rtitle: Enh, I just saw the below lines that GKE expects `../locations/..`. This is probably fine then.

I just wish GCP was consistent with itself ;)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,252 @@
...
+
+//the cluster must have a name, and a com.google.container.v1.NodePool. The NodePool must have an initialNodeCount and a name.
+//see getDefaultCluster for an example of construction with the minimum fields necessary, plus some others you almost certainly want to configure
+final case class KubernetesCreateClusterRequest(project: GoogleProject,
+                                                location: Location,
+                                                cluster: com.google.container.v1.Cluster)
+
+//this is NOT analogous to clusterName in the context of dataproc/GCE. A single cluster can have multiple nodes, pods, services, containers, deployments, etc.
+//clusters should most likely NOT be provisioned per user as they are today. More design/security research is needed
+final case class KubernetesClusterName(value: String) extends KubernetesNameValidation
```
rtitle: Maybe we should rename the current `ClusterName` to `DataprocClusterName` 🤔

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,252 @@
...
+  override def getJavaSerialization: V1Service = {
+    val v1Service = new V1Service()
+    v1Service.setKind(SERVICE_KIND) //may not be necessary
+    v1Service.setMetadata(name.getJavaSerialization)
+
+    v1Service
+  }
+}
+
+final case class KubernetesLoadBalancerService(selector: KubernetesSelector,
```
rtitle: It's often a good practice to put ADT definitions instead the trait's companion object:
```
sealed trait Foo extends Product with Serializable {
  def type: String
  def name: String
}
object Foo {
  final case class Bar(name: String) extends Foo {
    override val type: String = "bar"
  }
  final case class Baz(name: String) extends Foo {
    override val type: String = "baz"
  }
}
```

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import cats.effect.{Blocker, IO}
+import cats.effect.concurrent.Semaphore
+import cats.mtl.ApplicativeAsk
+import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+//TODO: migrate to a unit test
+//TODO: investigate running minikube in a docker for unit/automation tests https://banzaicloud.com/blog/minikube-ci/
```
rtitle, line 18: 👍 .. maybe we could have tickets for these. Still useful to have a manual test here.

rtitle (approved):
See comments/questions above.. but in general this looks great. Thumbing for now, assuming we will iterate on this as time goes on.

Question.. does this require a version bump? Probably not as we're just adding new classes. Please update the `google2/CHANGELOG.md` regardless though to note the addition of the k8s clients.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
Qi77Qi: +1 we shouldn't throw 404 in a library. It should be up to caller what the action is when the cluster doesn't exist

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
```diff
@@ -0,0 +1,54 @@
...
+    tracedGoogleRetryWithBlocker(
+      Async[F].delay(clusterManagerClient.deleteCluster(clusterId.idString)),
+      f"com.google.cloud.container.v1.ClusterManagerClient.deleteCluster(${clusterId.idString})"
+    )
+
+  private def tracedGoogleRetryWithBlocker[A](fa: F[A], action: String)(implicit ev: ApplicativeAsk[F, TraceId]): F[A] =
+    blockerBound.withPermit(
+      blocker.blockOn(
+        tracedRetryGoogleF(retryConfig)(fa, action).compile.lastOrError
+      )
```
Qi77Qi:
```
private def retryF[A](fa: F[A], loggingMsg: String)(implicit ev: ApplicativeAsk[F, TraceId]): F[A] =
   tracedRetryGoogleF(retryConfig)(blockerBound.withPermit(blocker.blockOn(fa)), loggingMsg).compile.lastOrError
```
We shouldn't wrap the whole retry inside `blockOn`, we should give `blockOn` the exact `F[A]` that you want to run on the blocking execution context

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+      Async[F].delay(kubernetesClient.createNamespace(namespace.getJavaSerialization, null, null, null))
+    })
+
+  private def getClient(clusterId: KubernetesClusterId)(implicit ev: ApplicativeAsk[F, TraceId]): F[CoreV1Api] = {
+    val clientOpt = clientStore.get(clusterId)
+
+    //we always update the token, even for existing clients, so we don't have to maintain a reference to the last time each client was updated
+    //unfortunately, the kubernetes client does not implement a gcp authenticator, so we must do this ourselves.
+    //See this for details https://github.com/kubernetes-client/java/issues/290
+    credentials.refreshIfExpired()
```
Qi77Qi: this `refreshIfExpired` is side effecting, hence it should be in `F`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,121 @@
...
+          )
+          .setSslCaCert(certStream)
+      )
+    }
+  }
+
+  //TODO: retry once we know what kubernetes codes are applicable
+  private def blockingClientProvider[A](clusterId: KubernetesClusterId,
+                                        fa: CoreV1Api => F[A])(implicit ev: ApplicativeAsk[F, TraceId]): F[A] =
+    blockerBound.withPermit(
```
Qi77Qi, line 138: use retry

Qi77Qi (changes requested):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
jdcanas: I don't know if we will ever need getters for those, I think we can save the information we used to create them if we need that as state in leo. 

In other words, I don't know an exact use case for getters, so they may be needed but I'd prefer to defer until we find a use case.    

For clusters, there are certain things GKE magically sets up for us (like certs), so we need the getter there.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
jdcanas: Yeah, I agree, I thought a lot about this before putting state in and I think its the best approach. Its 'loose' state too, almost more of a cache, since at any time the program can restart, and the worst that will happen is we will need to perform the getCluster google call again and reconstruct the client. 

We technically don't have to store it, google holds all the information so we could either 1) call google every time we want to interact with the cluster or 2) store it in a DB, but I like doing this and holding it in memory since storing it in a DB would involve saving a certificate, and calling google repeatedly has performance overhead

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: Hmm... the reason I did it like this was because of the mobbing we did where @Qi77Qi mentioned that she assumes that the data in a case class is validated at the time of construction. I kind of like that assumption, as otherwise the validation error is pushed further downstream and a bit harder to reason about. 

To me, it makes more sense that it errors out at instantiation as opposed to when you try to use the data. There could be a more elegant way to implement that than a trait though, curious to hear your thoughts

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
jdcanas: I defer retrying, since I'm not sure what exactly we want to retry in the context of the kubernetes client as opposed to the google client, which has fairly established retry predicates in wb-libs. I leave a comment above the retry config param, since we will need to reconstruct the predicates to handle the relevant kubernetes client codes

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
rtitle: Maybe something like this?
```
scala> :paste
// Entering paste mode (ctrl-D to finish)

val regex = "(?:[a-z](?:[-a-z0-9]{0,38}[a-z0-9])?)".r

final case class KubernetesName private (value: String) extends AnyVal

object KubernetesName {
  def apply(value: String): KubernetesName = 
    if (regex.pattern.matcher(value).matches()) new KubernetesName(value)
    else throw new Exception("regex doesn't match")
}

// Exiting paste mode, now interpreting.

regex: scala.util.matching.Regex = (?:[a-z](?:[-a-z0-9]{0,38}[a-z0-9])?)
defined class KubernetesName
defined object KubernetesName

scala> KubernetesName("foo")
res0: KubernetesName = KubernetesName(foo)

scala> KubernetesName("Foo")
java.lang.Exception: regex doesn't match
  at KubernetesName$.apply(<pastie>:21)
  ... 29 elided
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKubernetesInterpreter.scala
jdcanas: so are you saying 

```
    blockerBound.withPermit(
      blocker.blockOn(
        tracedRetryGoogleF(retryConfig)(fa, action).compile.lastOrError
      )
    )
```

turns into this (pre scalafmt)

```
    tracedRetryGoogleF(retryConfig)(
      blockerBound.withPermit(
        blocker.blockOn(fa)
      ),
      action).compile.lastOrError
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
jdcanas: its removed in a recent revision, I think you reviewed a cached version somehow

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
jdcanas: I don't think there are based on all the code I've seen and this: https://developers.google.com/identity/protocols/oauth2/scopes

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: Hmm.. I think it should be done but in a separate PR as it will make this require a version bump

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: Hmm.. I would be ok with this approach if we could simplify all Names to KubernetesName, but I think there is value in having multiple case classes for each type of Kubernetes name for type safety. This validation is also used for some GKE names as well as kubernetes client names. It seems cumbersome to have a companion object and apply for all names related to these two services.

I found a method `require`, which based on the [documentation](https://www.scala-lang.org/api/current/scala/Predef$.html) seems to be for this exact purpose. I really like the flexibility that the trait gives us over a companion object for many private case classes, since this validation applies to both GKEModels and KubernetesModels.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,140 @@
...
+      new CacheLoader[KubernetesClusterId, ApiClient] {
+        def load(clusterId: KubernetesClusterId): ApiClient = {
+          val res = for {
+            _ <- StructuredLogger[F]
+              .info(s"Determined that there is no cached client for kubernetes cluster ${clusterId}. Creating a client")
+            cluster <- gkeService.getCluster(clusterId)
+            token <- getToken
+            client <- createClient(
+              cluster.getOrElse(
+                throw KubernetesClusterNotFoundException(
```
rtitle: Avoid `throw`, I would do something like this:
```
val res = for {
  _ <- StructuredLogger[F]
    .info(s"Determined that there is no cached client for kubernetes cluster ${clusterId}. Creating a client")
  clusterOpt <- gkeService.getCluster(clusterId)
  cluster <- clusterOpt.fold(
    Async[F].raiseError[Cluster](KubernetesClusterNotFoundException(
      s"Could not create client for cluster ${clusterId} because it does not exist in google")
    ))(Async[F].pure)
  token <- getToken
  client <- createClient(cluster, token)
} yield client
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -82,16 +82,34 @@ package object google2 {
...
+    Resource.make(Sync[F].delay(resource))(c => Sync[F].delay(c.shutdown()) >> Sync[F].delay(c.close()))
+
+  def autoClosableResourceF[F[_]: Sync, A <: AutoCloseable](resource: => A): Resource[F, A] =
     Resource.make(Sync[F].delay(resource))(c => Sync[F].delay(c.close()))
 
   // Recovers a F[A] to an F[Option[A]] depending on predicate
   def recoverF[F[_]: Sync, A](fa: F[A], pred: Throwable => Boolean): F[Option[A]] =
     fa.map(Option(_)).recover { case e if pred(e) => None }
+
+  def streamFUntilDone[F[_]: Timer, A, B <: DoneCheckable](fa: F[A],
```
rtitle: Nice

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
rtitle: Ok, I can buy this

google2/CHANGELOG.md
```diff
@@ -11,12 +11,13 @@ Changed:
 - Bump `http4s` version to `0.21.0`
 - Bump `cats-effect` version to `2.1.2`
 - Bump `scalacheck` version to `1.14.3`
+- Add `GKEInterpreter`, `GKEService`, `KubernetesService`, and `KubernetesInterpreter`
```
rtitle: Can you put this in the `Added` section?

rtitle: Also maybe note the dependencies added/changed

rtitle:
Nice! Just a couple more comments above, this is looking great though

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: 👍 on not throwing...you can do something like this as well
```
Async[F].fromEither(clusterOpt.toRight(KubernetesClusterNotFoundException(
      s"Could not create client for cluster ${clusterId} because it does not exist in google")
    ))
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -0,0 +1,84 @@
...
+    tracedGoogleRetryWithBlocker(
+      Async[F].delay(clusterManagerClient.deleteCluster(clusterId.idString)),
+      f"com.google.cloud.container.v1.ClusterManagerClient.deleteCluster(${clusterId.idString})"
+    )
+
+  //delete and create operations take around ~5mins with simple tests, could be longer for larger clusters
+  override def pollOperation(operationId: KubernetesOperationId,
+                             delay: FiniteDuration,
+                             maxAttempts: Int): Stream[F, GKEPollOperation] = {
+    val opString = operationId.idString
```
Qi77Qi: remove this line

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    val request = GetOperationRequest
+      .newBuilder()
+      .setName(operationId.idString)
+      .build()
+
+    val getOperation = Async[F].delay(clusterManagerClient.getOperation(request))
+
+    streamFUntilDone(getOperation, GKEPollOperation.apply, maxAttempts, delay)
+    //TODO: remove below before merge, just displaying the alternative
```
Qi77Qi: remove commented out code

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -82,16 +82,34 @@ package object google2 {
...
                              retryNextDelay: FiniteDuration => FiniteDuration,
                              maxAttempts: Int,
                              retryable: Throwable => Boolean = scala.util.control.NonFatal.apply)
 final case class LoggableGoogleCall(response: Option[String], result: String)
+
+// trait used in conjunction with streamFUntilDone
+// implement isDone on a class extending this to poll if F[A] is done
+trait DoneCheckable {
+  def isDone: Boolean
+}
```
Qi77Qi, line 128: I would do typeclass pattern like this:

```
trait DoneCheckable[A] {
  def isDone(a: A): Boolean
}

object DoneCheckableInstances {
  implicit val containerDoneCheckable = new DoneCheckable[com.google.container.v1.Operation]{
    def isDone(op: com.google.container.v1.Operation): Boolean = op.getStatus == "DONE"
  }
  implicit val computeDoneCheckable = new DoneCheckable[com.google.cloud.compute.v1.Operation]{
    def isDone(op: com.google.cloud.compute.v1.Operation): Boolean = op.getStatus == "DONE"
  }
}

final case class DoneCheckableOps[A: DoneCheckable](a: A) {
  def isDone: Boolean = implicitly[DoneCheckable].isDone(a)

}

object DoneCheckableSyntax {
  implicit def doneCheckableSyntax[A](a: A): DoneCheckableOps[A] = DoneCheckableOps(a)
}
```

This way, you won't need `GKEPollOperation`  and `PollOperation`, and at call site, once you `import DoneCheckableSyntax._`, you'll be able to call `isDone` on regular `com.google.cloud.compute.v1.Operation` and `com.google.container.v1.Operation`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
Qi77Qi: I like what Rob suggested as well, but with slight change
```
final case class KubernetesName(value: String) extends AnyVal
object KubernetesName {
  def fromString(str: String): Either[Throwable, KubernetesName] = /* do regex validation */
}
```
notice it should return `Either`...Caller should know that converting string to 
`KubernetesName` is partial, and potentially do something about it (for example return back request)

also `require` throws exception

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,273 @@
...
+  
+  require(isValidName, s"The name ${value} must match the regex ${regex}")
+}
+
+// Google kubernetes client models //
+object GKEModels {
+
+  //"us-central1" is an example of a valid location.
+  final case class Parent(project: GoogleProject, location: Location) {
+    def parentString: String = s"projects/${project.value}/locations/${location.value}"
```
Qi77Qi: use `val`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,273 @@
...
+                                  name: NodePoolName,
+                                  autoscalingConfig: ClusterNodePoolAutoscalingConfig =
+                                    KubernetesConstants.DEFAULT_NODEPOOL_AUTOSCALING)
+
+  final case class KubernetesClusterId(project: GoogleProject, location: Location, clusterName: KubernetesClusterName) {
+    def idString: String = s"projects/${project.value}/locations/${location.value}/clusters/${clusterName.value}"
+  }
+
+  final case class KubernetesOperationId(project: GoogleProject, location: Location, operation: Operation) {
+    def idString: String = s"projects/${project.value}/locations/${location.value}/operations/${operation.getName}"
```
Qi77Qi: use `val`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+}
+
+// Kubernetes client models and traits //
+sealed trait KubernetesSerializable extends Product with Serializable {
```
Qi77Qi: why is this needed?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -0,0 +1,273 @@
...
+  final case class KubernetesSelector(labels: Map[String, String])
+
+  final protected case class KubernetesServiceType(value: String)
+
+  final case class KubernetesMasterIP(value: String) {
+    val url = s"https://${value}"
+  }
+
+  final case class KubernetesClusterCaCert(value: String) {
+    val base64Cert = Base64.decodeBase64(value)
```
Qi77Qi: `val base64Cert = Either.catchNonFatal(Base64.decodeBase64(value)`

Qi77Qi (changes requested):

Qi77Qi:
overall looks good 👍 ..a few more questions/comments

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: It defines the interface for all Kubernetes ADTs to get the java representation for use with the client library. Its useful because a lot of the Kubernetes objects have nested components, and this allows them serialize any subcomponents with a common interface. 

Additionally, there are scenarios where subclasses use polymorphism to get more specific. For example, `KubernetesServiceKind` is a trait that extends this trait, and there are a few type of services in Kubernetes that are concrete implementations

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
jdcanas: I see the value in removing the map step and making a doneWrapper with implicits, but I think the non-type classed trait with case classes approach provides more flexibility and readability. 

1) the flexibility of being able to specify any number of parameters to a case class allows for more complex computations of 'done' as long as the result is a boolean, while type classing does not allow for a variable number of parameters as easily
2) in terms of readability, I feel the syntax of ` def streamFUntilDone[F[_]: Timer, A, B <: DoneCheckable]` requires a bit of unwrapping, and the use of implicits further obfuscates the intent

Qi77Qi: let's jump on a call

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -0,0 +1,151 @@
...
+              endpoint.url,
+              token.getTokenValue
+            )
+            .setSslCaCert(certStream)
+        )
+      }
+    } yield (apiClient)
+  }
+
+  //TODO: retry once we know what kubernetes codes are applicable
```
Qi77Qi, line 136: remove todo

Qi77Qi (approved):
👍 lgtm

google2/CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.8
```
rtitle, line 5: If you're bumping to 0.8 you need to update `Settings.scala`, I didn't see that in the PR

rtitle (approved):

dvoet (approved):

PR #288, opened 2020-03-18
[IA-1754] Add VPC and Firewall methods to GoogleComputeService

rtitle:
This is to support https://broadworkbench.atlassian.net/browse/IA-1754. For a user project, Leo will do the following:

1. Delete `default-allow-rdp` and `default-allow-ssh` firewall rules in the default network
2. Add `default-broad-allow-ssh` firewall rule in the default network, restricted to Broad internal IPs
3. Create a dedicated network and subnet for Leonardo in the us-central1 region
4. Create the Leonardo tcp:443 firewall rule in the Leo subnet

These steps are taking [these guidelines](https://docs.google.com/document/d/1adV0LC2f_GIpl3A1AeoQuNiwcP59NToIt6VYT3xRCkU/edit) into account.

No version bump because I'm only adding APIs.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
@@ -280,7 +280,7 @@ private[google2] class GoogleComputeInterpreter[F[_]: Async: StructuredLogger: T
   ): Stream[F, PollOperation] =
     (Stream.eval(getOperation(project, operation)) ++ Stream.sleep_(delay))
       .repeatN(maxAttempts)
-      .map(op => PollOperation(op, op.getStatus == "DONE"))
+      .map(PollOperation.fromOperation)
       .takeThrough(!_.isDone)
 
   private def buildMachineTypeUri(zone: ZoneName, machineTypeName: MachineTypeName): String =
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
@@ -277,10 +277,11 @@ private[google2] class GoogleComputeInterpreter[F[_]: Async: StructuredLogger: T
 
   override def pollOperation(project: GoogleProject, operation: Operation, delay: FiniteDuration, maxAttempts: Int)(
     implicit ev: ApplicativeAsk[F, TraceId]
-  ): Stream[F, Operation] =
+  ): Stream[F, PollOperation] =
     (Stream.eval(getOperation(project, operation)) ++ Stream.sleep_(delay))
       .repeatN(maxAttempts)
-      .takeThrough(_.getStatus == "DONE")
+      .map(op => PollOperation(op, op.getStatus == "DONE"))
+      .takeThrough(!_.isDone)
 
   private def buildMachineTypeUri(zone: ZoneName, machineTypeName: MachineTypeName): String =
     s"zones/${zone.value}/machineTypes/${machineTypeName.value}"
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
@@ -241,47 +241,65 @@ private[google2] class GoogleComputeInterpreter[F[_]: Async: StructuredLogger: T
     )
   }
 
-  override def getOperation(project: GoogleProject,
-                            operation: Operation)(implicit ev: ApplicativeAsk[F, TraceId]): F[Operation] =
-    (getZoneName(operation.getZone), getRegionName(operation.getRegion)) match {
-      case (Some(zone), _) =>
-        val request = ProjectZoneOperationName
-          .newBuilder()
-          .setProject(project.value)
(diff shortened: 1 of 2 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
```diff
@@ -19,6 +19,8 @@ private[google2] class GoogleComputeInterpreter[F[_]: Async: StructuredLogger: T
   diskClient: DiskClient,
   zoneClient: ZoneClient,
   machineTypeClient: MachineTypeClient,
+  networkClient: NetworkClient,
```
rtitle, line 24: Should we break out some of these Google clients to their own service/interpreter? 🤔

kyuksel: Yea, it seems like there's enough security logic to justify breaking it out 👍

kyuksel (approved):
LGTM 👍 Breaking out the security logic is a good idea, either as part of this or a separate PR.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -135,3 +193,10 @@ final case class ZoneName(value: String) extends AnyVal
 final case class FirewallRuleName(value: String) extends AnyVal
 final case class MachineTypeName(value: String) extends AnyVal
 final case class RegionName(value: String) extends AnyVal
+final case class NetworkName(value: String) extends AnyVal
+final case class SubnetworkName(value: String) extends AnyVal
+final case class PollOperation(op: Operation, isDone: Boolean)
```
Qi77Qi: maybe
```
final case class PollOperation(op: Operation) {
def isDone: Boolean = op.getStatus == "DONE"
}
```
?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -111,18 +139,48 @@ object GoogleComputeService {
...
+      .setCredentialsProvider(credentialsProvider)
+      .build()
 
     for {
       instanceClient <- resourceF(InstanceClient.create(instanceSettings))
       firewallClient <- resourceF(FirewallClient.create(firewallSettings))
       diskClient <- resourceF(DiskClient.create(diskSettings))
       zoneClient <- resourceF(ZoneClient.create(zoneSettings))
       machineTypeClient <- resourceF(MachineTypeClient.create(machineTypeSettings))
+      networkClient <- resourceF(NetworkClient.create(networkSettings))
```
Qi77Qi, line 177: none of these clients has any `shutDown` function on them?

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -68,6 +74,28 @@ trait GoogleComputeService[F[_]] {
...
+
+  def getSubnetwork(project: GoogleProject, region: RegionName, subnetwork: SubnetworkName)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[Option[Subnetwork]]
+
+  def createSubnetwork(project: GoogleProject, region: RegionName, subnetwork: Subnetwork)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[Operation]
+
+  def getOperation(project: GoogleProject, operation: Operation)(
```
Qi77Qi: we're giving this function an `Operation`, and returning an `Operation`, this seems a bit suspicious. Can we pass in what we actually need? or return just the `status` of the operation?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
rtitle: On the other hand all these clients are under `com.google.cloud.compute.v1._` which is why I left them all in `GoogleComputeService`.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
rtitle: Yeah.. 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
rtitle: I think `resourceF` handles the `shutdown`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
rtitle: yeah maybe will pass `OperationName` instead. We're currently just using the `status` but `Operation` has a lot more fields that might be useful.. I like returning the whole object.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
Qi77Qi: ah cool... 👍

PR #291, opened 2020-03-26
IA-1675  add openTelemetry module

Qi77Qi:
1. Add `openTelemetry` module. This is not actually using `openTelemetry` api yet since it doesn't have stackdriver exporter yet. Hence using `openCencus` for now (opencensus and openTracing are merged into openTelemetry)

2. Bump scalatest version across all modules. This will bring in new scalatest for apps that don't specify scalatest version; otherwise, no effect on apps.

3. majority of the file changes are from `scalafmt`...didn't realize non-google2 modules have never been formatted, which created tons of diff

**PR checklist**

In all cases:

### DIFF
--- a/openTelemetry/CHANGELOG.md
+++ b/openTelemetry/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+This file documents changes to the `workbench-openTelemetry` library, including notes on how to upgrade to new versions.
+
+## 0.1
+
+### Added
+- Add `OpenTelemetryMetrics`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-openTelemetry" % "0.1-TRAVIS-REPLACE-ME"`
--- a/openTelemetry/CHANGELOG.md
+++ b/openTelemetry/CHANGELOG.md
@@ -7,4 +7,7 @@ This file documents changes to the `workbench-openTelemetry` library, including 
 ### Added
 - Add `OpenTelemetryMetrics`
 
+### Changed
+- Bump `scalatest` to `3.1.1`
+
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-openTelemetry" % "0.1-TRAVIS-REPLACE-ME"`
--- a/openTelemetry/README.md
+++ b/openTelemetry/README.md
@@ -1,0 +1,33 @@
+#Testing Locally in sbt console
+
+This is an example on how to test Google2 methods locally. In this example, we are testing the 
+GoogleStorageInterpreter.getObjectMetadata method. Before this, we have downloaded a credential JSON 
+file with which we can create a GoogleStorageService object.
+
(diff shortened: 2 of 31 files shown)


openTelemetry/README.md
```diff
@@ -0,0 +1,33 @@
...
+val resource = OpenTelemetryMetrics.resource[IO](Paths.get("<your google service account path>"), "test_app")
+val res = resource.use {
+  openTelemetry => 
+    openTelemetry.incrementCounter("qi_counter", 1)
+}
+
+res.unsafeRunSync()
+```
+
+`scala> res0.unsafeRunSync()`
```
kyuksel, line 33: Thank you for adding this!

project/Dependencies.scala
```diff
@@ -75,6 +75,10 @@ object Dependencies {
   val fs2Io: ModuleID = "co.fs2" %% "fs2-io" % "2.0.1"
   val rawlsModel: ModuleID = "org.broadinstitute.dsde" %% "rawls-model" % "0.1-0d02c8ce-SNAP" exclude("com.typesafe.scala-logging", "scala-logging_2.11") exclude("com.typesafe.akka", "akka-stream_2.11")
   val newRelic: ModuleID = "com.newrelic.agent.java" % "newrelic-api" % "5.0.0"
+//  val openTelemetry: ModuleID = "io.opentelemetry" % "opentelemetry-sdk" % "0.2.4"
```
kyuksel: Do we need ^^ ?

[Here](https://github.com/open-telemetry/opentelemetry-java/blob/master/QUICKSTART.md) it says 

> Libraries that want to export telemetry data using OpenTelemetry MUST only depend on the opentelemetry-api package and should never configure or depend on the OpenTelemetry SDK.

openTelemetry/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+This file documents changes to the `workbench-openTelemetry` library, including notes on how to upgrade to new versions.
```
kyuksel, line 3: Mention `scalaVersion` change too?

kyuksel (approved):
LGTM 👍 Minor comments inline.

project/Dependencies.scala
Qi77Qi: yea, this is not needed...will remove

openTelemetry/CHANGELOG.md
Qi77Qi: it's a minor bump for all published libraries...I'm not sure if it belongs in a particular module's changelog, but it also doesn't matter to user since it's a scala minor version bump

rtitle (changes requested):
If we're scalafmt-ing the whole repo I think it should probably be done in its own PR. Could format just the new module for now.

Qi77Qi:
@rtitle I think I'll leave the scalafmt commit in since I might have other stuff amended into that commit

PR #295, opened 2020-04-06
add deleteBucket

Qi77Qi:
**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -10,13 +10,14 @@ import cats.effect.concurrent.Semaphore
 import cats.implicits._
 import com.google.auth.oauth2.ServiceAccountCredentials
 import com.google.cloud.storage.BucketInfo.LifecycleRule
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -275,6 +276,19 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
     )
   }
 
+  override def deleteBucket(googleProject: GoogleProject,
+                   bucketName: GcsBucketName,
+                   bucketSourceOptions: List[BucketSourceOption] = List.empty,
+                   traceId: Option[TraceId] = None,
+                   retryConfig: RetryConfig = standardRetryConfig): Stream[F, Unit] = {
+    val deleteBucket = Async[F].delay(db.delete(bucketName.value, bucketSourceOptions: _*))
+    retryGoogleF(retryConfig)(
```
rtitle, line 285: Should we recover 404s? E.g. if you delete a bucket that doesn't exist, should it throw or return `Unit`?

Qi77Qi: good point...`db.delete` returns false if not found, I just updated the signature to `Stream[F, Boolean]` as well...

rtitle (approved):

PR #296, opened 2020-04-07
Jc k8 security

jdcanas:
**PR checklist**
- No version bump

In all cases:

### DIFF
--- a/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala
+++ b/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala
@@ -56,6 +56,7 @@ object Test {
       .setSubnetwork(KubernetesSubNetwork(project, RegionName("us-central1"), SubnetworkName("kube-test")).idString)
       .setNetworkPolicy(KubernetesConstants.getDefaultNetworkPolicy()) //needed for security
       .build()
+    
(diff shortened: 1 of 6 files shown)


google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/KubernetesManualTest.scala
```diff
@@ -48,8 +48,16 @@ object Test {
     }
   }
 
-  def callCreateCluster(clusterId: KubernetesClusterId = clusterId) = createCluster(KubernetesCreateClusterRequest(project, location,
-    KubernetesConstants.getDefaultCluster(nodePoolName.right.get, clusterName.right.get)))
+  def callCreateCluster(clusterId: KubernetesClusterId = clusterId) =  {
```
Qi77Qi: what is this funciton returning?

jdcanas: `IO[Operation]`. I left it blank because I don't think its very valuable to annotate/fix the returns of the manual testing functions

Qi77Qi: Unless you don't expect others to read/use this code, I think it's better to specify return type...

Qi77Qi (approved):

PR #297, opened 2020-04-08
Add deleting objects recursively for deleteBucket

Qi77Qi:
1. deleteBucket will fail if there're still objects in the bucket...hence, provide a `isRecursive` flag like `google` lib
2. Add `credentials` option to `getBlob`

**PR checklist**

In all cases:

### DIFF
--- a/google2/README.md
+++ b/google2/README.md
@@ -24,11 +24,11 @@ import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 import cats.effect.Blocker
 implicit val cs = IO.contextShift(global)
 implicit val t = IO.timer(global)
 implicit def logger = Slf4jLogger.getLogger[IO]
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
 val blocker = Blocker.liftExecutionContext(global)
 ```
 
--- a/google2/README.md
+++ b/google2/README.md
@@ -24,9 +24,12 @@ import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import cats.effect.Blocker
 implicit val cs = IO.contextShift(global)
 implicit val t = IO.timer(global)
-implicit def logger = Slf4jLogger.getLogger[IO]  
+implicit def logger = Slf4jLogger.getLogger[IO]
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+val blocker = Blocker.liftExecutionContext(global)
 ```
 
 `scala> GoogleStorageService.resource[IO]("credentials.json")`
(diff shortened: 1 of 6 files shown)


google2/README.md
```diff
@@ -24,9 +24,12 @@ import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
 import cats.effect.IO
 import org.broadinstitute.dsde.workbench.model.google.GcsBucketName
 import org.broadinstitute.dsde.workbench.google2.GcsBlobName
+import cats.effect.Blocker
 implicit val cs = IO.contextShift(global)
 implicit val t = IO.timer(global)
-implicit def logger = Slf4jLogger.getLogger[IO]  
+implicit def logger = Slf4jLogger.getLogger[IO]
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
```
jdcanas: nit: should be with the other imports. Not needed, but it would make me feel better

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -278,15 +278,45 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
+        Option(currentPage).traverse { p =>
+          val fetchNext = retryGoogleF(retryConfig)(
+            blockingF(Async[F].delay(p.getNextPage)),
+            traceId,
+            s"com.google.api.gax.paging.Page.getNextPage"
+          )
+          fetchNext.compile.lastOrError.map(next => (p, next))
+        }
+      }
+      blobId <- Stream.fromIterator[F](page.getValues.iterator().asScala).map(b => Option(b)).unNone.map(_.getBlobId)
```
jdcanas, line 309: `blobIds` maybe?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -278,15 +278,45 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer: Async
...
+    } yield blobId
+
+    for {
+      _ <- if (isRecursive) {
+        val r = for {
```
jdcanas, line 314: There are three different `r`s in the scope of this for comprehension, it gets a bit hard to keep track. It'd be helpful if they had distinct names.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
jdcanas, line 309: Huh, I didn't know you could use Option like this to get rid of nulls. TIL

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -212,6 +212,7 @@ trait GoogleStorageService[F[_]] {
    */
   def deleteBucket(googleProject: GoogleProject,
                    bucketName: GcsBucketName,
+                   isRecursive: Boolean = false,
```
jdcanas, line 217: Is there any reason we wouldn't want it to be recursive by default besides to prevent this from being considered a breaking change? 

Just curious, since I don't think it counts as a breaking change, and it seems like its desirable to have it recursive by default

jdcanas (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: no..this is one BlobId...on the right is a stream of BlobId, `<-` is getting inside the `Stream`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
Qi77Qi: google api throws exception if the bucket is not empty when trying to delete. I think that's desired default behavior to prevent people from accidentally losing data

PR #299, opened 2020-04-17
[No ticket]add clusterip

jdcanas:
clusterIP is correct, nodePort should not be used if the desired use case is to make an application inaccesible from outside the cluster
**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
@@ -119,7 +119,7 @@ object GKEModels {
   final case class NodePoolConfig(initialNodes: Int,
                                   name: NodePoolName,
                                   autoscalingConfig: ClusterNodePoolAutoscalingConfig =
-                                    KubernetesConstants.DEFAULT_NODEPOOL_AUTOSCALING)
+                                  KubernetesConstants.DEFAULT_NODEPOOL_AUTOSCALING)
 
   final case class KubernetesClusterId(project: GoogleProject, location: Location, clusterName: KubernetesClusterName) {
     val idString: String = s"projects/${project.value}/locations/${location.value}/clusters/${clusterName.value}"
@@ -301,6 +301,7 @@ object KubernetesModels {
   sealed trait KubernetesServiceKind {
     val SERVICE_TYPE_NODEPORT = "NodePort"
     val SERVICE_TYPE_LOADBALANCER = "LoadBalancer"
+    val SERVICE_TYPE_CLUSTERIP = "ClusterIP"
 
     def serviceType: KubernetesServiceType
     def name: KubernetesServiceName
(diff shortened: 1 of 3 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -301,6 +301,7 @@ object KubernetesModels {
   sealed trait KubernetesServiceKind {
```
kyuksel: May want to have this `extend Product with Serializable` so [types look nicer](https://stackoverflow.com/questions/36525804/scala-case-class-extending-product-with-serializable) 😊

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -301,6 +301,7 @@ object KubernetesModels {
...
     val SERVICE_TYPE_NODEPORT = "NodePort"
     val SERVICE_TYPE_LOADBALANCER = "LoadBalancer"
+    val SERVICE_TYPE_CLUSTERIP = "ClusterIP"
```
kyuksel: Might make sense to have a `ServiceType` ADT for ^^ too.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -323,6 +324,13 @@ object KubernetesModels {
       val serviceType = KubernetesServiceType(SERVICE_TYPE_NODEPORT)
     }
 
+    final case class KubernetesClusterIPService(selector: KubernetesSelector,
```
kyuksel, line 327: Nit: I wonder if it'd be sufficient to remove `KubernetesCluster` from `case class` names since they're already within `KubernetesServiceKind` namespace. It may also be good to use the name `Type` instead of `Kind` for consistency.

kyuksel (commented):
Good call on `ClusterIP` 👍 A few comments inline.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: I changed it to `Kind` everywhere that is applicable. 'Type' and 'Kind' are distinct Kubernetes concepts within the .yaml manifests

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
jdcanas: We do, but I had it inverted such that the service types were defined separate from the strings. I fixed.

Qi77Qi (approved):

kyuksel (approved):

PR #305, opened 2020-05-14
IA-Add google2 code for disk management

gpcarr:
This PR adds Disk algebra to `workbenchGoogle2`. Algebra methods include `createDisk`, `deletedDisk` and `listDisks` (and moved `resizeDisk` from `GoogleComputeService` to new `GoogleDiskService`) 

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
```diff
@@ -0,0 +1,66 @@
...
+    val projectZoneDiskName = ProjectZoneDiskName.of(diskName.value, project.value, zone.value)
+    retryF(
+      Async[F].delay(diskClient.deleteDisk(projectZoneDiskName)),
+      s"com.google.cloud.compute.v1.DiskClient.deleteDisk(${projectZoneDiskName.toString})"
+    )
+  }
+
+  override def listDisks(project: GoogleProject, zone: ZoneName)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[List[Disk]] = {
```
gpcarr: For this method, `diskClient.listDisks` returns object `ListDisksPagedResponse`. I'm trying to return object `List[Disk]`. The code compiles but the types do not match, my initial thought was that I have to do some processing to the return value of `diskClient.listDisks` but was thrown off when it compiled as is.

Qi77Qi: For your question, is `ListDisksPagedResponse` a subtype of `Disk`? It says pagedResponse, so I'm guessing you'll need to do something similar to https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala#L405

Another thing, I would return `Stream[F, Disk]` in this case. Your `retryF` will return `Stream[F, A]` if you get rid of `.compile.lastOrError` in the end

gpcarr: Ah yes, okay I'll work on these changes. But if I get rid of `.compile.lastOrError` won't this affect the other algebra methods that don't return a list?

Qi77Qi: yea. you'll just do `.compile.lastOrError` in individual function that expects `F[A]`

Qi77Qi:
also don't forget to
```
 Replace the appropriate version hashes in README.md and the CHANGELOG.md for any libs you changed with TRAVIS-REPLACE-ME to auto-update the version string
``` 
btw

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
gpcarr: makes sense, thank you!

README.md
```diff
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.9-e66171c"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.9-TRAVIS-REPLACE-ME"`
```
jdcanas: Huh... why is this 0.9, despite the fact that the changelog is already beyond 1.0 ._.

I don't think the previous rename deserves a major version bump, but I'd say just bumping this to 1.1-[blah] is fine

Qi77Qi: mm yea..this is suspicious...last PR didn't need major bump I think...but now that we've bumped, I guess we'll just leave it....you'll need to update https://github.com/broadinstitute/workbench-libs/blob/develop/project/Settings.scala#L195 for bumping version

gpcarr: Yeah I was thinking the same thing, but wouldn't I change this to `1.0-blah` since the previous version should be 1.0 and not 0.9? 

@Qi77Qi I updated that line, it's at the bottom of the PR for some reason :)

rtitle: If we're sticking with `1.1`, update this README too so it's consistent.

google2/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+## 1.1
+Changed:
+- Move `resizeDisk` from `GoogleComputeService` to `GoogleDiskService`
+
+Added:
+- Add `GoogleDiskService` and `GoogleDiskInterpreter`
+- Add `{create,delete}Disk` and `listDisks` to `GoogleDiskService`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "1.0-TRAVIS-REPLACE-ME"`
```
rtitle: Should be `1.1-TRAVIS-REPLACE-ME`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
rtitle: We paired on this and got `Stream.unfold` for paged responses working :-D

Qi77Qi (approved):
looks great 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
```diff
@@ -0,0 +1,73 @@
...
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[Operation] = {
+    val projectZoneDiskName = ProjectZoneDiskName.of(diskName.value, project.value, zone.value)
+    retryF(
+      Async[F].delay(diskClient.deleteDisk(projectZoneDiskName)),
+      s"com.google.cloud.compute.v1.DiskClient.deleteDisk(${projectZoneDiskName.toString})"
+    ).compile.lastOrError
+  }
+
+  override def listDisks(project: GoogleProject, zone: ZoneName)(
```
gpcarr, line 42: when testing this method, I changed the api call to google and set a maximum results per page of 2 and made sure that my code still returns all the disks under `callisto-dev`
Screen Shot 2020-05-18 at 5 27 58 PM

google2/README.md
```diff
@@ -25,10 +25,14 @@ import cats.effect.IO
...
 import org.broadinstitute.dsde.workbench.google2.GcsBlobName
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import cats.mtl.ApplicativeAsk
+import java.util.UUID
+import org.broadinstitute.dsde.workbench.model.TraceId
 import cats.effect.Blocker
 implicit val cs = IO.contextShift(global)
 implicit val t = IO.timer(global)
 implicit def logger = Slf4jLogger.getLogger[IO]
+implicit val traceId = ApplicativeAsk.const[IO, TraceId](TraceId(UUID.randomUUID()))
```
gpcarr, line 35: added `traceId` to this README because most methods now include implicit param `traceId`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
```diff
@@ -0,0 +1,73 @@
...
+    val projectZone = ProjectZoneName.of(project.value, zone.value)
+    for {
+      firstPageResults <- retryF(
+        Async[F].delay(diskClient.listDisks(projectZone)),
+        s"com.google.cloud.compute.v1.DiskClient.listDisks(${projectZone.toString})"
+      )
+
+      page <- Stream.fromIterator[F](firstPageResults.iteratePages().iterator().asScala)
+
+      res <- Stream.fromIterator[F](page.iterateAll().iterator().asScala)
```
rtitle: Hm, from javadoc:
```
  /**
   * Returns an iterable that traverses all of the elements of the underlying data source. The data
   * is fetched lazily page by page, where each page may contain multiple elements. A new page is
   * fetched whenever the elements of any particular page are exhausted.
   */
  Iterable<ResourceT> iterateAll();
```

Based on this, do we need to call `iteratePages` at all? Sounds like `iterateAll` will just fetch everything.

rtitle:
One comment above, rest looks good to me.

Seems there might be a compilation error though. Also can you make this a non-draft PR?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
Qi77Qi: the line above is calling `iteratePages`

rtitle: Sorry wasn't clear -- I know we're calling it, I'm asking if we _don't_ need to call `iteratePages` because `iterateAll` fetches everything across all pages anyway?

Qi77Qi: I didn't see duplicate items yesterday during pairing, but I guess it's worth double checking explicitly

gpcarr: I'll double check and test without `iteratePages` and let you know!

gpcarr:
> One comment above, rest looks good to me.
> 
> Seems there might be a compilation error though. Also can you make this a non-draft PR?

Hm yeah, looks like I have some unused imports. I'll update :)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
gpcarr: @rtitle good find! looks like you were right and we only have to call `.iterateAll`, I tested with and without a max results per page limit

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -303,7 +303,7 @@ object KubernetesModels {
 
   final case class KubernetesSelector(labels: Map[String, String])
 
-  final protected case class KubernetesServiceKindName(value: String)
+  final case class KubernetesServiceKindName(value: String)
```
gpcarr, line 306: removed `protected` as per @jdcanas request :)

rtitle (approved):
:shipit:

PR #306, opened 2020-05-18
[CA-841] Add bigquery service/interpreter using cloud client library to google2

marctalbott:
Ticket: [CA-841](https://broadworkbench.atlassian.net/browse/CA-841)

The cloud client version of BigQuery takes care of paginated responses for us, which will be helpful for working with Data Repo snapshots in Rawls. We think this is the functionality we'll need, but we'll hold off on merging this PR until we have a use case for it in Rawls (coming soon with [this epic](https://broadworkbench.atlassian.net/browse/CA-827)) and we can test it out against BigQuery for real and know if this 1. works and 2. does everything we need it to.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.10-93b852f"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.10-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
@@ -5,9 +5,8 @@ import com.google.cloud.bigquery.{BigQuery, JobId, QueryJobConfiguration, TableR
 import io.chrisdavenport.log4cats.StructuredLogger
 
 
-private[google2] class GoogleBigQueryInterpreter[F[_]: Sync: ContextShift: Timer](client: BigQuery,
-                                                                                  blocker: Blocker)
-                                                                                 (implicit logger: StructuredLogger[F])
+private[google2] class GoogleBigQueryInterpreter[F[_]: Sync: ContextShift: Timer: StructuredLogger](client: BigQuery,
+                                                                                                    blocker: Blocker)
   extends GoogleBigQueryService[F] {
 
   override def query(queryJobConfiguration: QueryJobConfiguration, options: BigQuery.JobOption*): F[TableResult] = {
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.9-e66171c"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "1.0-TRAVIS-REPLACE-ME"`
```
rtitle: FYI - I think the bump from `0.9` to `1.0` was a mistake. We discussed changing it to `0.10` in @gpcarr's PR: https://github.com/broadinstitute/workbench-libs/pull/305. I don't think that affects this PR because there's no version bump needed, just calling it out.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import cats.effect.{Blocker, ContextShift, Sync}
+import com.google.cloud.bigquery.{BigQuery, JobId, QueryJobConfiguration, TableResult}
+
+
+private[google2] class GoogleBigQueryInterpreter[F[_]: Sync: ContextShift](client: BigQuery,
+                                                                           blocker: Blocker) extends GoogleBigQueryService[F] {
+
+  override def query(queryJobConfiguration: QueryJobConfiguration, options: BigQuery.JobOption*): F[TableResult] = {
+    blockingF(Sync[F].delay[TableResult] {
```
Qi77Qi: do you care about retry? if so, you can use https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala#L30... it also logs the call for you

dvoet: @Qi77Qi as an aside, we discovered that com.google.cloud libraries have retry built in and it is blocking. It is probably not great to have both, we should figure out which one to keep.

dvoet (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
Qi77Qi: hmm I think we talked about this back when we added retry...I think the reason you gave back then was visibility. Personally I'm happy to rely on google's retry. In reality, I haven't seen a call being retried yet.

We do bake logging in `retryGoogleF` currently.  Shall we at least have that?

Qi77Qi:
btw, you can test this out against google API in console following https://github.com/broadinstitute/workbench-libs/blob/develop/google2/README.md

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
marctalbott: Since the plan is to rely on the com.google.cloud libraries' built in retry as we start using them in the cloud resource library, I'm going to also rely on it here at the cost of potentially losing some retry visibility. I will add in logging though, thanks!

gpolumbo-broad (approved):

dvoet (approved):

rtitle (approved):

PR #311, opened 2020-06-05
IA-1969 fix a bug where we aren't getting all instances

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1969

Tested in console:
```
scala> res.unsafeRunSync
res0: Map[org.broadinstitute.dsde.workbench.google2.DataprocRole,Set[org.broadinstitute.dsde.workbench.google2.InstanceName]] = Map(Master -> Set(InstanceName(qi-dp-0-m)), Worker -> Set(InstanceName(qi-dp-0-w-0), InstanceName(qi-dp-0-w-1), InstanceName(qi-dp-0-w-2)), SecondaryWorker -> Set())
```

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -153,22 +153,27 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Async: StructuredLogger: 
 
 object GoogleDataprocInterpreter {
   def getAllInstanceNames(cluster: Cluster): Map[DataprocRole, Set[InstanceName]] = {
-    def getFromGroup(role: DataprocRole)(group: InstanceGroupConfig): Map[DataprocRole, Set[InstanceName]] =
-      group.getInstanceNamesList
+    def getFromGroup(role: DataprocRole, group: InstanceGroupConfig): Map[DataprocRole, Set[InstanceName]] = {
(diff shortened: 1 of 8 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -152,23 +152,31 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Async: StructuredLogger:
 }
 
 object GoogleDataprocInterpreter {
+  // WARNING: Be very careful refactoring this function and make sure you test this out in console.
```
rtitle, line 155: I'm trying to understand the bug -- why didn't the old version work? The code seems pretty equivalent to me.

Qi77Qi: so in https://github.com/DataBiosphere/leonardo/pull/1395/files#diff-97adeeac0a7ef9b09ea7f9a42e224796L396...that last `.map` is in the outer `Option`, the original code here is the `map` on 161 is on `List`, so when it does `toMap` in the end, it's removing duplicate keys, vs the correct code should return a `Map` with one key value pair in it, value being a set.

rtitle: Ohh thanks, that makes sense.

rtitle (approved):

PR #315, opened 2020-06-16
make genDiskName non empty

Qi77Qi:
disk name can't be empty. Hence postfix a fixed string

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/Generators.scala
+++ b/google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/Generators.scala
@@ -44,7 +44,7 @@ object Generators {
     topic <- genTopicName
     filters <- genFilters
   } yield NotificationRequest(topic, "JSON_API_V1", filters.eventTypes, filters.objectNamePrefix)
-  val genDiskName = Gen.alphaLowerStr.map(x => DiskName(x.take(30)))
+  val genDiskName = Gen.alphaLowerStr.map(x => DiskName(x.take(30) + "disk"))
 
   implicit val arbProjectTopicName: Arbitrary[TopicName] = Arbitrary(genTopicName)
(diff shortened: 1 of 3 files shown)


rtitle (approved):

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/Generators.scala
```diff
@@ -44,7 +44,7 @@ object Generators {
     topic <- genTopicName
     filters <- genFilters
   } yield NotificationRequest(topic, "JSON_API_V1", filters.eventTypes, filters.objectNamePrefix)
-  val genDiskName = Gen.alphaLowerStr.map(x => DiskName(x.take(30)))
+  val genDiskName = Gen.alphaLowerStr.map(x => DiskName(x.take(30) + "disk"))
```
kyuksel: If the previous version was creating empty strings on occasion, will the new version create `DiskName`s of `disk` on occasion, too, and if so, would it be undesirable to generate the same name?

Could do the following to generate fixed-length strings without postfixing if you'd like:

```

scala> def alphaLowerStrOfLength(n: Int): Gen[String] = Gen.listOfN(n, Gen.alphaChar).map(_.mkString) 
alphaLowerStrOfLength: (n: Int)org.scalacheck.Gen[String]

scala> alphaLowerStrOfLength(5).sample
res27: Option[String] = Some(zlpho)

scala> alphaLowerStrOfLength(5).sample
res28: Option[String] = Some(aimcS)

scala> alphaLowerStrOfLength(5).sample
res29: Option[String] = Some(pgnjv)

scala> alphaLowerStrOfLength(5).sample
res30: Option[String] = Some(okyaw)
```

Qi77Qi: I think it'd be fine within the same test run....what you have fix the length, but I guess it's probably fine 🤔  ...

PR #320, opened 2020-06-23
[IA-1794] APIs to create ServiceAccount, Role and RoleBinding

kyuksel:
[JIRA ticket
](https://broadworkbench.atlassian.net/browse/IA-1794) plus adding logging for allo `KubernetesInterpreter` API methods

For usage examples, see `KubernetesManualTest` methods starting with [callCreateServiceAccount()](https://github.com/broadinstitute/workbench-libs/pull/320/files#diff-8ca7c4ebdf9dda798794c144ff2afedeR118).

**PR checklist**

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -68,40 +68,98 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
 
   // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#podspec-v1-core
   override def createPod(clusterId: KubernetesClusterId, pod: KubernetesPod, namespace: KubernetesNamespace): F[Unit] =
-    blockingClientProvider(clusterId, { kubernetesClient =>
-      Async[F].delay(
-        kubernetesClient.createNamespacedPod(namespace.name.value, pod.getJavaSerialization, null, null, null)
-      )
-    })
+    blockingClientProvider[CoreV1Api, Unit](
```
kyuksel: `blockingClientProvider` and `getClient()` assumed `CoreV1Api` would be used. `Role` and `RoleBinding` methods are in another API class (`RbacAuthorizationV1Api`) so I made this method and `getClient()` method more generic to work with both.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -68,40 +68,98 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        )
+      }
+    )
 
   //DO NOT QUERY THE CACHE DIRECTLY
   //There is a wrapper method that is necessary to ensure the token is refreshed
   //we never make the entry stale, because we always need to refresh the token (see comment above getToken)
   //if we did stale the entry we would have to unnecessarily re-do the google call
-  private def getClient(clusterId: KubernetesClusterId): F[CoreV1Api] =
+  private def getClient[A](clusterId: KubernetesClusterId, fa: ApiClient => A): F[A] =
```
kyuksel, line 194: see [this comment](https://github.com/broadinstitute/workbench-libs/pull/320/files#r444179355)

Qi77Qi, line 194: this fa is basically just an additional `.map` call...seems a bit unnecessary to bake it into `getClient` 🤔 ....it also makes the naming not matching what the function returns

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -325,4 +417,30 @@ object KubernetesModels {
     val base64Cert = Either.catchNonFatal(Base64.decodeBase64(value))
   }
 
+  final case class KubernetesApiGroup(name: ApiGroupName)
+  final case class KubernetesResource(name: ResourceName)
+  final case class KubernetesVerb(name: VerbName)
+  final case class KubernetesPolicyRule(apiGroups: List[KubernetesApiGroup],
```
Qi77Qi: shall we use `Set` here to avoid duplicates?

project/plugins.sbt
```diff
@@ -11,4 +11,4 @@ addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
 
 addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.3.2")
 
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.0") 
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.4")
```
Qi77Qi, line 14: 👍 thanks for upgrading :)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: why not do something like `def getClient(clusterId: KubernetesClusterId): F[ApiClient]`, and `def blockingClientProvider[A, B](clusterId: KubernetesClusterId, fa: ApiClient => F[A]): F[B]`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -133,15 +188,16 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
         .blockOn(
           for {
-            kubernetesClient <- getClient(clusterId)
-            clientCallResult <- fa(kubernetesClient)
-              .onError { //we aren't handling any errors here, they will be bubbled up, but we want to print a more helpful message that is otherwise obfuscated
+            kubernetesClient <- getClient(clusterId, fa)
+            clientCallResult <- fb(kubernetesClient)
+              .onError { // We aren't handling any errors here, they will be bubbled up, but we want to print a more helpful message that is otherwise obfuscated
+                // TODO: e.getResponseBody() doesn't seem to get printed out. Find out why and fix.
                 case e: ApiException => Async[F].delay(StructuredLogger[F].info(e.getResponseBody()))
```
Qi77Qi: not your code, but this is incorrect...`StructuredLogger[F].info(e.getResponseBody()` already returns an `F[Unit]`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: also for `blockingClientProvider`, it's probably better to add something like 
```
  private def blockingF[A](fa: F[A]): F[A] =
    blockerBound.withPermit(
      blocker
        .blockOn(fa)
    )
```
Instead of wrapping several blocking IOs in `blocker.blockOn`...

So I might just get rid of `blockingClientProvider`, and do this
```
      override def createRole(clusterId: KubernetesClusterId,
                          role: KubernetesRole,
                          namespace: KubernetesNamespace): F[Unit] =
    for {
      client <- blockingF(getClient(clusterId))
      apiClient = new RbacAuthorizationV1Api(client)
      _ <- blockingF(
        Async[F].delay(
          apiClient.createNamespacedRole(namespace.name.value, role.getJavaSerialization, null, "true", null)
        )
      )
    } yield ()
```

or with you version of `getClient`, 
```
      override def createRole(clusterId: KubernetesClusterId,
                          role: KubernetesRole,
                          namespace: KubernetesNamespace): F[Unit] =
    for {
      client <- blockingF(getClient(clusterId, x => new RbacAuthorizationV1Api(x)))
      _ <- blockingF(
        Async[F].delay(
          client.createNamespacedRole(namespace.name.value, role.getJavaSerialization, null, "true", null)
        )
      )
    } yield ()
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
jdcanas: ohh.. I thought this wasn't printing due to the fact StructuredLogger doesn't work in console

kyuksel: Removed the `Async[F].delay()` wrapper. Still doesn't print on the console, though.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
kyuksel: 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: we should also add logging for all function (not sure how this slipped through in previous PR reviews)...but similar to https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala#L12

kyuksel: Thanks. Done.

kyuksel:
@Qi77Qi Thank you for your feedback. I believe I've addressed all of your comments.

Qi77Qi (approved):
lgtm 👍  thanks for addressing comments!

PR #324, opened 2020-06-29
IA-2000 add dettach

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2000

1. Add `ConsoleLogger` in util2. This is used in cleanup script. It's also handy for console testing
2. Add `detachDisk ` in google2
3. Move `pollOperation` to its own class so that we don't have to bring in `GoogleComputeSerivce` for polling operation (now that GoogleDiskService is in a separate class). Evolved `pollOperation` a bit to support interruption.

**PR checklist**

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
```diff
@@ -299,18 +314,23 @@ private[google2] class GoogleComputeInterpreter[F[_]: Async: Parallel: Structure
     )
   }
 
-  override def pollOperation(project: GoogleProject, operation: Operation, delay: FiniteDuration, maxAttempts: Int)(
+  override def pollOperation(project: GoogleProject,
```
Qi77Qi: if we already have an `Operation` returned from google API call, we can directly use that `Operation` for `streamFUntilDone` instead of going `getOp`

rtitle (approved):
LGTM

PR #330, opened 2020-07-10
[IA-1911] Jc applicative patch

jdcanas:
Pesky java function we override prevents us from using Applicative implicits like we'd like

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
@@ -150,7 +150,7 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
     delete.handleErrorWith {
       case _: com.google.gson.JsonSyntaxException =>
         Async[F].unit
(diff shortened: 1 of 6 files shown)


Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -125,6 +127,33 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        s"io.kubernetes.client.apis.CoreV1Api.deleteNamespace(${namespace.name.value}, null, null, null, null, null, null)"
+      )
+    } yield ()
+
+    // There is a known bug with the client lib json decoding.  `com.google.gson.JsonSyntaxException` occurs every time.
+    // See https://github.com/kubernetes-client/java/issues/86
+    delete.handleErrorWith {
+      case _: com.google.gson.JsonSyntaxException =>
+        Async[F].unit
+      case e: Exception => Async[F].raiseError(e)
```
Qi77Qi: `handleErrorWith` deals with `Throwable`,  and `Exception` is a subtype..so I would remove `Exception` on line 153 here

PR #334, opened 2020-07-27
log only row count for BQ results

dvoet:
I noticed that `GoogleBigQueryInterpreter` logs the first 1024 chars of the response from BQ. This is problematic due to size and compliance (we want to keep our logs as free as possible from user data and there may be locality constraints on this data).

At the same time we do want to know what queries we run and response times. 

This PR introduces a way for each `Interpreter` to specify how to log a response and provides a default which preserves the current state. For BQ `TableResult` this is overridden to include only the result row count.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.workbench
 import java.util.concurrent.TimeUnit
 
 import DoneCheckableSyntax._
+import cats.Show
 import cats.implicits._
 import cats.effect.{Resource, Sync, Timer}
 import cats.mtl.ApplicativeAsk
@@ -39,7 +40,11 @@ package object google2 {
                        retryConfig.retryable)
   }
 
-  def withLogging[F[_]: Sync: Timer, A](fa: F[A], traceId: Option[TraceId], action: String)(
+  def withLogging[F[_]: Sync: Timer, A](fa: F[A],
+                                        traceId: Option[TraceId],
+                                        action: String,
+                                        resultFormatter: Show[A] =
+                                          Show.show(a => if (a == null) "null" else a.toString.take(1024)))(
     implicit logger: StructuredLogger[F]
   ): F[A] =
     for {
@@ -59,8 +64,8 @@ package object google2 {
           // Duplicate MDC context in regular logging until log formats can be changed in apps
           logger.error(loggingCtx, e)(msg.noSpaces)
         case Right(r) =>
-          val response = if (r == null) "null" else r.toString.take(1024)
-          val loggableGoogleCall = LoggableGoogleCall(Some(response), "Succeeded")
+          val response = Option(Show(resultFormatter).show(r))
+          val loggableGoogleCall = LoggableGoogleCall(response, "Succeeded")
           val msg = loggingCtx.asJson.deepMerge(loggableGoogleCall.asJson)
           logger.info(loggingCtx)(msg.noSpaces)
       }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
@@ -44,7 +44,7 @@ package object google2 {
                                         traceId: Option[TraceId],
                                         action: String,
                                         resultFormatter: Show[A] =
-                                          Show.show(a => if (a == null) "null" else a.toString.take(1024)))(
+                                          Show.show((a: A) => if (a == null) "null" else a.toString.take(1024)))(
     implicit logger: StructuredLogger[F]
   ): F[A] =
     for {
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
@@ -44,7 +44,7 @@ package object google2 {
                                         traceId: Option[TraceId],
                                         action: String,
                                         resultFormatter: Show[A] =
-                                          Show.show((a: A) => if (a == null) "null" else a.toString.take(1024)))(
+                                          Show.show[A](a => if (a == null) "null" else a.toString.take(1024)))(
     implicit logger: StructuredLogger[F]
   ): F[A] =
     for {
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.11-0d87698"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.11-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -11,11 +11,12 @@ Changed:
 - add deleteNamespace to `KubernetesService`
 - added mocks for `GKEService` and `KubernetesService`
 - optimized implementation of `GoogleStorageInterpreter.getBlobBody` to fully use streams
+- log only result row count for BigQuery queries
 
 Added:
 - Add `detachDisk`
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.11-0d87698"`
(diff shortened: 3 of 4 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -129,6 +132,10 @@ trait DoneCheckable[A] {
   def isDone(a: A): Boolean
 }
 
+trait ResultSerializerForLogging[A] {
```
Qi77Qi: this is essentially https://typelevel.org/cats/typeclasses/show.html ..so I would use `Show` instead

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -129,6 +132,10 @@ trait DoneCheckable[A] {
...
+  def toString(a: A): Option[String]
```
kyuksel: The implementation of this method seems to always be a `Some[String]` so could the result type just be a `String` instead of `Option[String]`?

kyuksel (commented):
LGTM overall. One question inline.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
dvoet: the current implementation does but I wanted to reserve the option for an Interpreter to not print any result

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
dvoet: I think this should return an Option and I like the purposeful name

rtitle (commented):
Looks fine to me but CI failed (I think it needs a `scalafmt`) and needs changelog bump

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
Qi77Qi: can we just put `""` when we don't want to print anything, that's what will happen essentially anyway

kyuksel: okay

dvoet: the purist in me says "" != null

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
Qi77Qi: for naming preference, I think it's very subjective...When I see `Show`, I immediately know what it could be used for (cuz the only thing you can do with it is get some string representation, vs if I see `ResultSerializerForLogging`, I have to go to its definition to see what it does.

also by requiring this in `withLogging`, this will be needed at all call sites, not just `GoogleBigQueryInterpreter.scala `

rtitle (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
dvoet: I would prefer a name that indicates what it is used for, not what it does (I think it is pretty clear what it does). And not something generic that may have other uses as well.

The problem with using `Show` is that there may be random `Show` implementations floating around that are not purpose built for log messages and the implicit system makes it very difficult to know if the wrong one is being used. For example, 6 months from now someone may think, "The code I am currently working on is a great place to use `Show`." Then, suddenly the logging messages change and the chance of someone noticing this is very slim.

I am happy for a better name suggestion as I agree that good names are subjective.

Is there a better place than `withLogging` to include this?

kyuksel (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
Qi77Qi: seems ci is all passing...so I guess `withLogging` isn't used much elsewhere...

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -22,12 +22,15 @@ package object google2 {
 
   implicit val finiteDurationEncoder: Encoder[FiniteDuration] = Encoder.encodeString.contramap(_.toString())
 
+  implicit def defaultResultSerializerForLogging[A <: Any]: ResultSerializerForLogging[A] =
```
dvoet: @Qi77Qi it compiles because of this

Qi77Qi: hmm.....does `bqTableResultSerializerForLogging` work with this? I think things get tricky when we have multiple implicit instances in scope...

Probably better to make another `withExplicitResponseLogging` function in addition to `withLogging` and leave `withLogging` as is?

dvoet: this was definitely something I wanted to see if there is feedback on. IntelliJ seems to pick the right one.

Qi77Qi: I never looked into how implicits are prioritized much, http://eed3si9n.com/implicit-parameter-precedence-again may help....For this PR, I would still recommend using `withExplicitResponseLogging` if caller chooses to. I think having 
 `implicit def defaultResultSerializerForLogging[A <: Any]` in package object (which makes this available to all `google2 ` namespace) will make it harder to migrate `google2` to future versions of scala as well.

dvoet: first problem I have run into exploring the `withExplicitResponseLogging` approach is that there are 2 functions that build from `withLogging`: `tracedLogging` and `retryGoogleF`. For completeness these would also need an explicit response flavor. Implicits take care of that plumbing for us.

This would be the 3rd or 4th package level implicit. Does it really make migration harder?

Qi77Qi: I think we can leave the retry related functions out now that google retries anyway.


> This would be the 3rd or 4th package level implicit. Does it really make migration harder?

I think there's a good chance. First thing is they're removing package object in scala 3 I think. There's a good amount of rework in scala3 around implicits.

dvoet: don't get me wrong... I hate implicits and would love a different solution. However this code is already riddled with implicits so using another one seems to fit in.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -39,7 +40,11 @@ package object google2 {
                        retryConfig.retryable)
   }
 
-  def withLogging[F[_]: Sync: Timer, A](fa: F[A], traceId: Option[TraceId], action: String)(
+  def withLogging[F[_]: Sync: Timer, A](fa: F[A],
+                                        traceId: Option[TraceId],
+                                        action: String,
+                                        resultFormatter: Show[A] =
```
dvoet, line 46: @Qi77Qi I changed it to add a parameter with a default arg

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -59,8 +64,8 @@ package object google2 {
           // Duplicate MDC context in regular logging until log formats can be changed in apps
           logger.error(loggingCtx, e)(msg.noSpaces)
         case Right(r) =>
-          val response = if (r == null) "null" else r.toString.take(1024)
-          val loggableGoogleCall = LoggableGoogleCall(Some(response), "Succeeded")
+          val response = Option(Show(resultFormatter).show(r))
```
Qi77Qi, line 67: fyi, you can also do
```
implicit val showA = resultFormatter
val response = r.show
```

Also I don't think you need `Option`

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
dvoet: `LoggableGoogleCall` requires an `Option`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -44,7 +44,7 @@ package object google2 {
                                         traceId: Option[TraceId],
                                         action: String,
                                         resultFormatter: Show[A] =
-                                          Show.show(a => if (a == null) "null" else a.toString.take(1024)))(
+                                          Show.show((a: A) => if (a == null) "null" else a.toString.take(1024)))(
```
Qi77Qi: you can do `Show.show[A](a =>???)`

PR #338, opened 2020-08-07
[IA-2117] Transition app from creating to ready or error

gpcarr:
Testing: I ran the `listPodStatus` method on an existing cluster and got the following output:
`List(KubernetesPodStatus(PodName(csi-cvmfsplugin-attacher-0),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-c2rgq),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-h56lx),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-j5fdr),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-nhj22),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-p5ss8),Ready), KubernetesPodStatus(PodName(csi-cvmfsplugin-provisioner-0),Ready), KubernetesPodStatus(PodName(galaxy1-cvmfs-cache-preload-0-vshq2),Ready), KubernetesPodStatus(PodName(galaxy1-galaxy-postgres-0),Ready), KubernetesPodStatus(PodName(galaxy1-job-0-f6fddd6fc-md5rz),Ready), KubernetesPodStatus(PodName(galaxy1-nfs-0),Ready), KubernetesPodStatus(PodName(galaxy1-web-fc99d8d98-p8bl5),Ready), KubernetesPodStatus(PodName(galaxy1-workflow-647d6d5898-26k94),Ready))`

I compared to running `kubectl get pods` with the same cluster and namespace:
image

as we see, all the pods in this cluster are either running or completed(maps to succeeded for pod phase) therefore we return that all the pods are ready in wb-libs\

**PR checklist**

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +91,49 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+      )
+      //TODO: add status to KubernetesPod
+      //Kubernetes pod phase is what we'll use to map to KubernetesPodStatus - phases include Pending, Running, Succeeded, Failed, Unknown (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)
+      // Pending will map to Creating, Running/Succeeded maps to Ready, Failed maps to Error
+      // an app is Ready when all pods return Ready status
+      listPods = response.getItems.asScala.toList.map(v1Pod => convertToKubernetesPod(v1Pod))
+      _ = println(listPods)
+    } yield (listPods)
+
+  override def getPodStatus(clusterId: KubernetesClusterId, name: PodName, namespace: KubernetesNamespace)(
```
rtitle: Will this method be needed by Leo, or can we use `listPods` to get the information we need?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -275,6 +322,22 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
     } yield (apiClient) // appending here a .setDebugging(true) prints out useful API request/response info for development
   }
 
+  private def convertToKubernetesPod(pod: V1Pod): KubernetesPod = {
+    // Open question - not sure if this is the right approach. all other wb-lib services return the google object and don't convert it to our leo data model. figured it
```
rtitle: This approach is generally fine -- it's true that we directly return Google model objects for a lot of the Google methods, but our _kubernetes_ client code tends to use our own models so this fits the pattern.

One thing to think about is whether we want `listPods` to return the full `KubernetesPod` object (in which case we need to convert containers/ports/etc) -- or rename the method to `listPodStatus` and return something like:
```
case class KubernetesPodStatus(name: PodName, status: PodStatus)
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
gpcarr: we can use `listPods`, I was testing if this k8s method made more sense to use but seems like it returns the full pod object. I'll remove

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
gpcarr: I think then for now, we should rename to `listPodStatus` and if we do need to return the full `KubernetesPod` object in the future then we can implement that then. I was leaning towards this but wanted to double check first

README.md
```diff
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.11-cac8af4"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.12-TRAVIS-REPLACE-ME"`
```
Qi77Qi: I don't think we need to bump version since this PR doesn't break any API

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +90,38 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        Some(traceId),
+        s"io.kubernetes.client.apis.CoreV1Api.listNamespacedPod(${namespace.name.value}, null, true, null, null, null, null, null, null, null)"
+      )
+
+      //Kubernetes pod phase is what we'll use to map to KubernetesPodStatus - phases include Pending, Running, Succeeded, Failed, Unknown (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)
+      // Pending will map to Creating, Running/Succeeded maps to Ready, Failed maps to Error
+      // an app is Ready when all pods return Ready status
+      listPodStatus = response.getItems.asScala.toList
+        .map(v1Pod => KubernetesPodStatus(PodName(v1Pod.getMetadata.getName), getPodStatus(v1Pod.getStatus.getPhase)))
+      _ = println(listPodStatus)
```
Qi77Qi: remove println

README.md
rtitle: I don't think this needs a version bump since it's just adding a new method. E.g. no compatibility issues.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +90,38 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        )
+      )
+      response <- withLogging(
+        call,
+        Some(traceId),
+        s"io.kubernetes.client.apis.CoreV1Api.listNamespacedPod(${namespace.name.value}, null, true, null, null, null, null, null, null, null)"
+      )
+
+      //Kubernetes pod phase is what we'll use to map to KubernetesPodStatus - phases include Pending, Running, Succeeded, Failed, Unknown (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)
+      // Pending will map to Creating, Running/Succeeded maps to Ready, Failed maps to Error
```
rtitle: This might be a good comment to put on the `PodStatus` ADT itself

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
rtitle: Remove println

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +90,38 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+      // an app is Ready when all pods return Ready status
+      listPodStatus = response.getItems.asScala.toList
```
Qi77Qi: we need to be careful with null pointer with java libraries. I would do `Option(response.getItems).map(l => l.asScala.toList).getOrElse(Nil)`

README.md
gpcarr: ah good point. i'll remove the version bump

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
gpcarr: will update

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +90,38 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        .map(v1Pod => KubernetesPodStatus(PodName(v1Pod.getMetadata.getName), getPodStatus(v1Pod.getStatus.getPhase)))
+      _ = println(listPodStatus)
+    } yield listPodStatus
+
+  private def getPodStatus(podPhase: String): PodStatus =
```
rtitle: Maybe could put this in `PodStatus` companion object (see below comment)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -467,6 +469,21 @@ object KubernetesModels {
                                          roleRef: KubernetesRoleRef,
                                          subjects: List[KubernetesSubject])
 
+  sealed trait PodStatus extends Product with Serializable
```
rtitle: I'd probably do something like this here: 
https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/samModels.scala

E.g.:
* define `asString` in the trait
* use `sealerate` to create a `Map[String, PodStatus]` in the companion object
* at the call site (`KubernetesInterpreter`) use the Map to lookup strings (like `PodStatus.stringToPodStatus.get(str)`)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
gpcarr: good point, will update

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
gpcarr: will update!

gpcarr: to double check, that means that the actual strings in the companion object will be the strings we get from kubernetes? ie Pending, Running, etc

kyuksel:
@gpcarr Have you tested the case where some of the pods hadn't become ready when you ran the `wb-libs` code as well?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
rtitle: Yes

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/MockKubernetesService.scala
```diff
@@ -28,6 +29,10 @@ class MockKubernetesService extends org.broadinstitute.dsde.workbench.google2.Ku
     namespace: KubernetesModels.KubernetesNamespace
   )(implicit ev: ApplicativeAsk[IO, TraceId]): IO[Unit] = IO.unit
 
+  override def listPodStatus(clusterId: GKEModels.KubernetesClusterId, namespace: KubernetesModels.KubernetesNamespace)(
```
Qi77Qi, line 33: I wonder if we should just make this function into `isAllPodsReady(clusterId: GKEModels.KubernetesClusterId, namespace: KubernetesModels.KubernetesNamespace): F[Boolean]` 🤔

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
gpcarr: just updated

gpcarr:
> @gpcarr Have you tested the case where some of the pods hadn't become ready when you ran the `wb-libs` code as well?

just finished addressing all comments, will test that case now!

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/MockKubernetesService.scala
gpcarr: hmmm then leo would have no way of knowing what the cluster's pod's statuses are in google and I don't see how we can transition an app to error in the case that a pod fails to start up. or am i missing something?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -87,6 +90,39 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+        )
+      )
+      response <- withLogging(
+        call,
+        Some(traceId),
+        s"io.kubernetes.client.apis.CoreV1Api.listNamespacedPod(${namespace.name.value}, null, true, null, null, null, null, null, null, null)"
+      )
+
+      listPodStatus = Option(response.getItems)
+        .map { l =>
```
rtitle, line 111: Was wondering if this can be changed to `foldMap` and get rid of the `combineAll` :)

rtitle (approved):
Code looks good, I think you need to update the CHANGELOG and hashes to `TRAVIS-REPLACE-ME` in README.md and CHANGELOG.md

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/MockKubernetesService.scala
Qi77Qi: for transition the app, leo only need to know if all pods are up, it doesn't really need to know each individual pod statuses

rtitle: There are 3 cases we care about though: all pods ready, some pods not ready, and some pods errored. So `F[Boolean]` probably isn't enough. I kind of like returning `List[PodStatus]` and Leo can handle that however it wants (including logging/alerting if some particular pods failed).

gpcarr: when transitioning to `Error` though, polling an endpoint that returns a Boolean probably won't suffice unless we just have a certain timeout and say that an app has error'd after a certain amount of time that this endpoint keeps returning false

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
gpcarr: I can give it a try!

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/MockKubernetesService.scala
Qi77Qi: In reality, it will be a bit more complicated than `Boolean`(maybe `PodsStatus.AllReady, PodsStatus.Error, PodsStatus.ZeroPods`)...my point being shall we consider not returning a `List[PodStatus]`, because the logic for translating the result is not as straightforward, but probably applicable to most all clients of the lib if the client only cares about whether all pods are ready or not.

Qi77Qi (approved):

PR #339, opened 2020-08-10
[No ticket] Add max retries to pubsub subscriber

jdcanas:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -241,5 +246,7 @@ final case class FlowControlSettingsConfig(maxOutstandingElementCount: Long, max
 final case class SubscriberConfig(pathToCredentialJson: String,
                                   topicName: TopicName,
                                   ackDeadLine: FiniteDuration,
+                                  maxRetries: MaxRetries,
                                   flowControlSettingsConfig: Option[FlowControlSettingsConfig])
+final case class MaxRetries(value: Int)
```
Qi77Qi: extends AnyVal

Qi77Qi (approved):

PR #340, opened 2020-08-10
[no ticket] remove unused code

Qi77Qi:
**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
@@ -139,7 +139,7 @@ object GoogleComputeService {
       interpreter <- fromCredential(scopedCredential, blocker, blockerBound, retryConfig)
     } yield interpreter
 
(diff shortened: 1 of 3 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -139,7 +139,7 @@ object GoogleComputeService {
       interpreter <- fromCredential(scopedCredential, blocker, blockerBound, retryConfig)
     } yield interpreter
 
-  private def fromCredential[F[_]: StructuredLogger: Async: Parallel: Timer: ContextShift](
+  def fromCredential[F[_]: StructuredLogger: Async: Parallel: Timer: ContextShift](
```
kyuksel, line 142: Looks like this could remain `private`?

Qi77Qi: no..I intentionally want to expose this funciton...because in many cases, we already have a `GoogleCredentials` object that can be re-used

kyuksel (approved):
LGTM. I'd keep `fromCredential()` `private` unless there's a use case for `public` that I'm missing.

PR #342, opened 2020-08-11
[QA-981] Delete partially created groups that return 5xx when creating before retrying

dvoet:
**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
@@ -9,14 +9,15 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.HttpResponseException
 import com.google.api.services.admin.directory.model.{Group, Member, Members}
 import com.google.api.services.admin.directory.{Directory, DirectoryScopes}
-import com.google.api.services.groupssettings.{Groupssettings, GroupssettingsScopes}
 import com.google.api.services.groupssettings.model.{Groups => GroupSettings}
+import com.google.api.services.groupssettings.{Groupssettings, GroupssettingsScopes}
 import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes._
 import org.broadinstitute.dsde.workbench.google.GoogleUtilities.RetryPredicates._
 import org.broadinstitute.dsde.workbench.metrics.GoogleInstrumentedService
 import org.broadinstitute.dsde.workbench.model._
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 /**
  * Created by mbemis on 8/17/17.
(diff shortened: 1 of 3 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -107,9 +108,21 @@ class HttpGoogleDirectoryDAO(appName: String,
...
     for {
-      _ <- retry(when5xx, whenUsageLimited, when404, whenInvalidValueOnBucketCreation, whenNonHttpIOException)(() => {
+      _ <- retryWithRecover(when5xx,
+                            whenUsageLimited,
+                            when404,
+                            whenInvalidValueOnBucketCreation,
+                            whenNonHttpIOException)(() => {
         executeGoogleRequest(inserter)
-      })
+      }) {
```
gpolumbo-broad: nit of nits....would you mind indenting/formatting this block to align the parens and braces

gpolumbo-broad (approved):
Just a formatting nit, but functionally 👍

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
marctalbott: @gpolumbo-broad I'm not sure I understand which parens and braces you want to align. I formatted this using scalafmt, so this should be consistent with how we have formatted the rest of workbench-libs

marctalbott: I guess the stuff down on line 191-193 is similar, is that the indentation you'd like to see?

gpolumbo-broad: I just find it really hard to read.  the way it's formatted at the moment, I had to manually go one by one through the braces and parens to match them up to understand what block of code goes with what.  I appreciate that this is the way scalafmt wanted to format it, but with the currying on `retryWithRecover` compounded with the fact that we're already in a `for` comp, I could make sense of it, but I had to kind of parse this code in my head first just to be able to read it.  Like I said, I 👍 the PR, formatting isn't something I want to tie up PRs usually, but when indentation and formatting is done properly, you can sort of ignore braces and parens most of the time and it makes the code easier to read.  I guess my only request going forward is to use scalafmt to get us started, but if you think reformatting manually will make the code more readable and understandable, that's more important to me.

gpolumbo-broad: I just find it really hard to read.  the way it's formatted at the moment, I had to manually go one by one through the braces and parens to match them up to understand what block of code goes with what.  I appreciate that this is the way scalafmt wanted to format it, but with the currying on `retryWithRecover` compounded with the fact that we're already in a `for` comp, I could make sense of it, but I had to kind of parse this code in my head first just to be able to read it.  Like I said, I 👍 the PR, formatting isn't something I want to tie up PRs usually, but when indentation and formatting is done properly, you can sort of ignore braces and parens most of the time and it makes the code easier to read.  I guess my only request going forward is to use scalafmt to get us started, but if you think reformatting manually will make the code more readable and understandable, that's more important to me.

marctalbott: Sure, I'm happy to adjust the formatting and indentation to improve readability as it's an easy fix that can be really helpful in the future, I just wasn't sure what you had in mind. I switched it up a little, let me know if you think it's better.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -107,9 +108,21 @@ class HttpGoogleDirectoryDAO(appName: String,
...
+        case t: Throwable if when5xx(t) =>
```
andy7i, line 118: took me a bit to realize this is used within `retryWithRecover` (and not called after). oops

andy7i (approved):
LGTM 👍

PR #346, opened 2020-08-18
[IA-2072] Add KubernetesService.getServiceExternalIp method

rtitle:
Used by Leo (https://github.com/DataBiosphere/leonardo/pull/1565) to poll whether the nginx ingress controller is up.

**PR checklist**

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -1,4 +1,5 @@
 package org.broadinstitute.dsde.workbench.google2
```
rtitle, line 1: The models in this file scare me a little

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -144,6 +144,39 @@ class KubernetesInterpreter[F[_]: Async: StructuredLogger: Effect: Timer: Contex
...
+                                    namespace: KubernetesNamespace,
+                                    serviceName: ServiceName)(
+    implicit ev: ApplicativeAsk[F, TraceId]
+  ): F[Option[ServiceExternalIp]] =
+    for {
+      traceId <- ev.ask
+      client <- blockingF(getClient(clusterId, new CoreV1Api(_)))
+      call = blockingF(
+        Async[F].delay(
+          client.listNamespacedService(namespace.name.value, null, "true", null, null, null, null, null, null, null)
```
Qi77Qi, line 156: should we recover from 404 here as well and return `None`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -428,6 +428,8 @@ object KubernetesModels {
   final case class PortName(value: String) extends AnyVal
   final case class Protocol(value: String) extends AnyVal
 
+  final case class ServiceExternalIp(value: String) extends AnyVal
```
Qi77Qi: what does this IP look like? if it's just regular IP, can we just use an `IP` data type?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
rtitle: Yeah I think that makes sense

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
rtitle: It's just a regular IP -- good point maybe can just use `IP` data type. I think Leo has an `IP` class, maybe can move that to wb-libs.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
kyuksel: me too 😨

kyuksel (approved):
LGTM 👍

PR #348, opened 2020-08-21
[IA-2145] Add `addWorkloadIdentityUserRoleForUser` to wb-libs to support workload identity in back leo

gpcarr:
This PR adds method `addWorkloadIdentityUserRoleForUser` to project `workbenchGoogle`. The purpose of this method is to add the role of `workloadIdentityUser` to a google service account. The equivalent gcloud command to this method is:
`gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
  --member "serviceAccount:$PROJECT.svc.id.goog[$NS/$KSA]" \
  $GSA@$PROJECT.iam.gserviceaccount.com`

Google does some magic to associate the GSA to the given NS/KSA. The call to this wb-libs method in Leo will look like this:
`googleDao.addWorkloadIdentityUserRoleForUser([google_project], GSA, "serviceAccount:$PROJECT.svc.id.goog[$NS/$KSA]")`

I tested this by creating a cluster, GSA, and KSA manually (through gcloud and kubectl commands). I then created a pod with the KSA that runs the cloud-sdk container image, and connected to it with an interactive session. Before using the new wb-libs method, I tried running `gcloud auth list` from within the pod and saw: 
Screen Shot 2020-08-20 at 11 16 05 PM
Which makes sense, because I annotaed the KSA with the GSA information. But if then try to hit a gcloud api by running for example `gcloud projects list` I got the following:
Screen Shot 2020-08-20 at 11 17 30 PM
Because the binding has not happened on the google side yet. 

I then used the new wb-libs method as mentioned above (`googleDao.addWorkloadIdentityUserRoleForUser([google_project], GSA, "serviceAccount:$PROJECT.svc.id.goog[$NS/$KSA]")`), restarted the pod and was able to successfully run `gcloud list projects`:
Screen Shot 2020-08-20 at 11 19 32 PM

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -140,15 +140,15 @@ trait GoogleIamDAO {
 
   /**
    *
-   * @param serviceAccountProject the google projecting which we're adding the IAM roles
-   * @param serviceAccountEmail the service account on which to add the roles to (i.e. the IAM resource).
-   * @param memberEmail the user email address for which to add the roles to
-   * @param rolesToAdd Set of roles to add to the serviceAccountEmail
+   * @param serviceAccountProject the google project where serviceAccount lives
+   * @param serviceAccount the service account (i.e. the IAM resource) to which to add the policy binding
+   * @param member the user email address for which to add the binding to
+   * @param rolesToAdd Set of roles to add to the serviceAccount
    * @return
    */
   def addIamPolicyBindingOnServiceAccount(serviceAccountProject: GoogleProject,
-                                          serviceAccountEmail: WorkbenchEmail,
-                                          memberEmail: WorkbenchEmail,
+                                          serviceAccount: WorkbenchEmail,
+                                          member: WorkbenchEmail,
                                           rolesToAdd: Set[String]): Future[Unit]
 
   /**
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -139,6 +139,19 @@ trait GoogleIamDAO {
                      rolesToRemove: Set[String]): Future[Boolean]
 
   /**
+   *
+   * @param serviceAccountProject the google projecting which we're adding the IAM roles
+   * @param serviceAccountEmail the service account on which to add the roles to (i.e. the IAM resource).
+   * @param memberEmail the user email address for which to add the roles to
+   * @param rolesToAdd Set of roles to add to the serviceAccountEmail
+   * @return
+   */
+  def addIamPolicyBindingOnServiceAccount(serviceAccountProject: GoogleProject,
+                                          serviceAccountEmail: WorkbenchEmail,
+                                          memberEmail: WorkbenchEmail,
+                                          rolesToAdd: Set[String]): Future[Unit]
+
+  /**
    * Adds the Service Account User role for the given users on the given service account.
    * This allows the users to impersonate as the service account.
    * @param serviceAccountProject the project in which to add the roles
@@ -149,18 +162,6 @@ trait GoogleIamDAO {
   def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject,
                                        serviceAccountEmail: WorkbenchEmail,
                                        email: WorkbenchEmail): Future[Unit]
-
-  /**
-   * Adds the Workload Identity User role for the given users on the given service account.
-   * This supports making Google Api calls from a Kubernetes Cluster.
-   * @param serviceAccountProject the project in which to add the roles
-   * @param serviceAccountEmail the service account on which to add the Service Account User role
-   *                               (i.e. the IAM resource).
-   * @param email the user email address for which to add Workload Identity to
-   */
-  def addWorkloadIdentityUserRoleForUser(serviceAccountProject: GoogleProject,
-                                         serviceAccountEmail: WorkbenchEmail,
-                                         email: WorkbenchEmail): Future[Unit]
 
   /**
    * Creates a user-managed key for the given service account.
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -142,8 +142,8 @@ trait GoogleIamDAO {
    *
    * @param serviceAccountProject the google project where serviceAccount lives
    * @param serviceAccount the service account (i.e. the IAM resource) to which to add the policy binding
-   * @param member the user email address for which to add the binding to
-   * @param rolesToAdd Set of roles to add to the serviceAccount
+   * @param member the user email address for which to add the roles to
+   * @param rolesToAdd set of roles to add to the member
    * @return
    */
   def addIamPolicyBindingOnServiceAccount(serviceAccountProject: GoogleProject,
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -151,6 +151,18 @@ trait GoogleIamDAO {
                                        email: WorkbenchEmail): Future[Unit]
 
   /**
+   * Adds the Workload Identity User role for the given users on the given service account.
+   * This supports making Google Api calls from a Kubernetes Cluster.
+   * @param serviceAccountProject the project in which to add the roles
+   * @param serviceAccountEmail the service account on which to add the Service Account User role
+   *                               (i.e. the IAM resource).
+   * @param email the user email address for which to add Workload Identity to
+   */
+  def addWorkloadIdentityUserRoleForUser(serviceAccountProject: GoogleProject,
+                                         serviceAccountEmail: WorkbenchEmail,
+                                         email: WorkbenchEmail): Future[Unit]
+
+  /**
    * Creates a user-managed key for the given service account.
    * @param serviceAccountProject the google project the service account resides in
    * @param serviceAccountEmail the service account email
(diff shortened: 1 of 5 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -243,6 +243,23 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
       }.void
     }
 
+  override def addWorkloadIdentityUserRoleForUser(serviceAccountProject: GoogleProject,
```
Qi77Qi: will https://github.com/broadinstitute/workbench-libs/blob/93d8de40c1dec58529b244f248b93683b0bce06d/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala#L121 suffice your use case?

What's the difference between SA and email in your input?

gpcarr: it won't unfortunately. SA is the user's pet SA while `email` is the unique string that google needs to make the binding between the KSA and pet SA. some more docs here: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_on_new_cluster
Google needs that unique string (`project-id.svc.id.goog[k8s-namespace/ksa-name]`) along with the pet SA.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -243,6 +243,23 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
       }.void
     }
```
kyuksel, line 246: It looks like the only difference between `addWorkloadIdentityUserRoleForUser()` and `addServiceAccountUserRoleForUser()` is that the former hardcodes the role to `roles/iam.serviceAccountUser` and the other to `roles/iam.workloadIdentityUser`. Can we factor out the common code to another method like:

```scala
def addIamPolicyBindingOnServiceAccount(serviceAccountProject: GoogleProject,
                                                                            serviceAccountEmail: WorkbenchEmail,
                                                                            memberEmail: WorkbenchEmail,
                                                                            rolesToAdd: Set[String]): Future[Unit] =
    // Note the project here is the one in which we're adding the IAM roles.
    // In this case the accountEmail acts as a resource, not an identity. Therefore the serviceAccountEmail
    // should live in the provided serviceAccountProject. For more information on service account permissions, see:
    // - https://cloud.google.com/iam/docs/service-accounts#service_account_permissions
    // - https://cloud.google.com/iam/docs/service-accounts#the_service_account_user_role
    // Also a helpful SO answer can be viewed at https://stackoverflow.com/a/61878052/2851999
    getServiceAccountPolicy(serviceAccountProject, serviceAccountEmail).flatMap { policy =>
      val updatedPolicy =
        updatePolicy(policy, userEmail, MemberType.ServiceAccount, rolesToAdd, Set.empty)
      val policyRequest = new ServiceAccountSetIamPolicyRequest().setPolicy(updatedPolicy)
      val request = iam
        .projects()
        .serviceAccounts()
        .setIamPolicy(s"projects/${serviceAccountProject.value}/serviceAccounts/${serviceAccountEmail.value}",
                      policyRequest)
      retry(when5xx, whenUsageLimited, when404, whenInvalidValueOnBucketCreation, whenNonHttpIOException) { () =>
        executeGoogleRequest(request)
      }.void
    }
```

and then Leo can call that method with `rolesToAdd = Set("roles/iam.workloadIdentityUser")`.

We can keep `addServiceAccountUserRoleForUser()` for backwards compatibility but it just call the new method with
`rolesToAdd = Set("roles/iam.serviceAccountUser")`.

Does that make sense?

kyuksel: Potentially helpful reference info:
https://cloud.google.com/iam/docs/reference/rest/v1/Policy
https://stackoverflow.com/questions/61875357/gcloud-confusion-around-add-iam-policy-binding

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
kyuksel: Yea, `addIamRoles()` adds policy binding (which adds roles to members) for a **project** whereas what we want to do is add policy binding for a **serviceaccount**. In the latter, the service account is a Google Cloud **resource** and not **identity**, which is I think what complicates it a lot. [This thread](https://stackoverflow.com/questions/61875357/gcloud-confusion-around-add-iam-policy-binding) may be helpful in clarifying.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
gpcarr: makes sense, will update now!

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -138,6 +138,19 @@ trait GoogleIamDAO {
                      memberType: MemberType,
                      rolesToRemove: Set[String]): Future[Boolean]
 
+  /**
+   *
+   * @param serviceAccountProject the google projecting which we're adding the IAM roles
```
kyuksel: It'd be good to make it really clear that we're not adding roles to the Google project.

```suggestion
   * @param serviceAccountProject the google project where `serviceAccount` lives
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -138,6 +138,19 @@ trait GoogleIamDAO {
...
+   * @param serviceAccountEmail the service account on which to add the roles to (i.e. the IAM resource).
```
kyuksel: Nit: Would it be more concise to remove `Email` from the parameter names without losing info?

kyuksel:
```suggestion
   * @param serviceAccount the service account to which to add the policy binding.
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -221,17 +221,13 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
...
 
-  override def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject,
-                                                serviceAccountEmail: WorkbenchEmail,
-                                                userEmail: WorkbenchEmail): Future[Unit] =
-    // Note the project here is the one in which we're adding the IAM roles.
-    // In this case the serviceAccountEmail acts as a resource, not an identity. Therefore the serviceAccountEmail
-    // should live in the provided serviceAccountProject. For more information on service account permissions, see:
-    // - https://cloud.google.com/iam/docs/service-accounts#service_account_permissions
-    // - https://cloud.google.com/iam/docs/service-accounts#the_service_account_user_role
```
kyuksel, line 231: Would be good to have the helpful comment on top of the new method

kyuksel (changes requested):
Looks good functionally. As we chatted about offline, inline are just some comments to make the methods easier to parse for future users.

gpcarr:
> Looks good functionally. As we chatted about offline, inline are just some comments to make the methods easier to parse for future users.

just updated with everything we talked about, lmk if i'm missing anything!

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -140,15 +140,15 @@ trait GoogleIamDAO {
 
   /**
    *
-   * @param serviceAccountProject the google projecting which we're adding the IAM roles
-   * @param serviceAccountEmail the service account on which to add the roles to (i.e. the IAM resource).
-   * @param memberEmail the user email address for which to add the roles to
-   * @param rolesToAdd Set of roles to add to the serviceAccountEmail
+   * @param serviceAccountProject the google project where serviceAccount lives
+   * @param serviceAccount the service account (i.e. the IAM resource) to which to add the policy binding
+   * @param member the user email address for which to add the binding to
```
kyuksel:
```suggestion
   * @param member the user email address for which to add the roles to
```

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -140,15 +140,15 @@ trait GoogleIamDAO {
...
+   * @param rolesToAdd Set of roles to add to the serviceAccount
```
kyuksel:
```suggestion
   * @param rolesToAdd set of roles to add to member
```

kyuksel (changes requested):
Super close...just a couple of wording changes 🙏

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -221,28 +221,42 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
...
+        .setIamPolicy(s"projects/${serviceAccountProject.value}/serviceAccounts/${serviceAccount.value}", policyRequest)
       retry(when5xx, whenUsageLimited, when404, whenInvalidValueOnBucketCreation, whenNonHttpIOException) { () =>
         executeGoogleRequest(request)
       }.void
     }
 
+  override def addServiceAccountUserRoleForUser(serviceAccountProject: GoogleProject,
+                                                serviceAccountEmail: WorkbenchEmail,
+                                                userEmail: WorkbenchEmail): Future[Unit] =
+    // Note the project here is the one in which we're adding the IAM roles.
```
rtitle: Not sure we need this comment in 2 places - above the `addIamPolicyBindingOnServiceAccount` is fine

rtitle (approved):
Looks good

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
gpcarr: will update!

kyuksel (approved):
Thanks for sticking with my endless rounds of comments 🙏

PR #350, opened 2020-09-03
[IA-2189] google2 uses the com.api.services GKE client

rtitle:
Because of limitations in the newer `com.google.cloud.v1` client for GKE, we need to fall back to using the old `com.google.api.services` client for cluster creation only. This PR is WIP and untested, but shows how things may look.

Bumped a minor version because it changes some GKE interfaces.

[GC] Couple of comments 
- the Travis CI Pull Request build is actually linking to a build off of `develop` - i kicked off a build off of this branch --> https://travis-ci.com/github/broadinstitute/workbench-libs/builds/184951360
- the Travis CI Branch build is failing because `GKEService.scala` isn't formatted correctly. However, I have explicitly configured IntelliJ to `scalafmt` on save and also formatted the file using Cmd + Shift + L on the file and nothing changes. Anyone have advice here?
- **Testing**: I published wb-libs and pulled the `SNAP` versions into Leonardo. See this [PR](https://github.com/DataBiosphere/leonardo/pull/1575) for how everything was tested.
- Testing: I also tested that `GkeService.getCluster` and `deleteCluster` work with no changes to them:
Screen Shot 2020-09-17 at 12 50 58 PM
In the above screenshot, I created a cluster with the old client lib but then used the new client lib to get the cluster and we see that the object returned is from the newer client lib (ie `com.google.container.v1.Cluster`)
I then tested `deleteCluster`
Screen Shot 2020-09-17 at 12 52 57 PM
the above screenshot is the result of the `deleteCluster` call and even tho the cluster was created with the old client lib, in the google console we can see that the cluster is still deleted:
Screen Shot 2020-09-17 at 12 49 04 PM

More comments inline. [GC] inline comments have been addressed

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
@@ -37,7 +37,7 @@ final class GKEInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
 
     tracedGoogleRetryWithBlocker(
       Async[F].delay(legacyClient.projects().locations().clusters().create(parent, googleRequest).execute()),
-      f"com.google.api.services.container.Projects.Locations.Cluster(${request})"
+      s"com.google.api.services.container.Projects.Locations.Cluster(${request})"
     )
   }
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
@@ -4,15 +4,7 @@ import cats.effect.concurrent.Semaphore
 import cats.effect.{Async, Blocker, ContextShift, Timer}
 import cats.mtl.ApplicativeAsk
 import com.google.cloud.container.v1.ClusterManagerClient
-import com.google.container.v1.{
-  Cluster,
-  CreateClusterRequest,
-  CreateNodePoolRequest,
-  GetOperationRequest,
-  IPAllocationPolicy,
-  NodePool,
-  Operation
-}
+import com.google.container.v1.{Cluster, CreateNodePoolRequest, GetOperationRequest, NodePool, Operation}
 import fs2.Stream
 import io.chrisdavenport.log4cats.StructuredLogger
 import org.broadinstitute.dsde.workbench.{DoneCheckable, RetryConfig}
@@ -25,6 +17,7 @@ import scala.concurrent.duration.FiniteDuration
 
 final class GKEInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
   clusterManagerClient: ClusterManagerClient,
+  legacyClient: com.google.api.services.container.Container,
   blocker: Blocker,
   blockerBound: Semaphore[F],
   retryConfig: RetryConfig
@@ -32,25 +25,19 @@ final class GKEInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
 
   override def createCluster(
     request: KubernetesCreateClusterRequest
-  )(implicit ev: ApplicativeAsk[F, TraceId]): F[Operation] = {
-    val parent = Parent(request.project, request.location)
+  )(implicit ev: ApplicativeAsk[F, TraceId]): F[com.google.api.services.container.model.Operation] = {
+    val parent = Parent(request.project, request.location).toString
 
-    val createClusterRequest: CreateClusterRequest = CreateClusterRequest
-      .newBuilder()
-      .setParent(parent.toString)
-      .setCluster(
-        request.cluster.toBuilder
-          .setIpAllocationPolicy( //otherwise it uses the legacy one, which is insecure. See https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips
-            IPAllocationPolicy
-              .newBuilder()
-              .setUseIpAliases(true)
-          )
-      )
-      .build()
+    // Note createCluster uses the legacy com.google.api.services.container client rather than
+    // the newer com.google.container.v1 client because certain options like Workload Identity
+    // are only available in the old client.
+
+    val googleRequest = new com.google.api.services.container.model.CreateClusterRequest()
+      .setCluster(request.cluster)
 
     tracedGoogleRetryWithBlocker(
-      Async[F].delay(clusterManagerClient.createCluster(createClusterRequest)),
-      f"com.google.cloud.container.v1.ClusterManagerClient.createCluster(${request})"
+      Async[F].delay(legacyClient.projects().locations().clusters().create(parent, googleRequest).execute()),
+      f"com.google.api.services.container.Projects.Locations.Cluster(${request})"
     )
   }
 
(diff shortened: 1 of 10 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -25,32 +17,27 @@ import scala.concurrent.duration.FiniteDuration
...
-    val parent = Parent(request.project, request.location)
+  )(implicit ev: ApplicativeAsk[F, TraceId]): F[com.google.api.services.container.model.Operation] = {
+    val parent = Parent(request.project, request.location).toString
 
-    val createClusterRequest: CreateClusterRequest = CreateClusterRequest
-      .newBuilder()
-      .setParent(parent.toString)
-      .setCluster(
-        request.cluster.toBuilder
-          .setIpAllocationPolicy( //otherwise it uses the legacy one, which is insecure. See https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips
```
rtitle, line 43: Previously wb-libs was appending `setIpAllocationPolicy` to the `com.google.container.v1.Cluster` object passed by Leo. Changed wb-libs to just use the `Cluster` object as-is: Leo will be in charge of building the object.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -108,6 +109,13 @@ package object google2 {
       credential <- Resource.liftF(Sync[F].delay(UserCredentials.fromStream(credentialFile)))
     } yield credential
 
+  // returns legacy GoogleCredential object which is only used for the legacy com.google.api.services client
+  def googleCredential[F[_]: Sync](pathToCredential: String): Resource[F, GoogleCredential] =
+    for {
+      credentialFile <- org.broadinstitute.dsde.workbench.util2.readFile(pathToCredential)
+      credential <- Resource.liftF(Sync[F].delay(GoogleCredential.fromStream(credentialFile)))
```
rtitle: TODO maybe need `createScoped`

project/Dependencies.scala
```diff
@@ -65,7 +65,7 @@ object Dependencies {
   val kubernetesClient: ModuleID = "io.kubernetes" % "client-java" % "5.0.0" % "compile"
   val googleBigQueryNew: ModuleID = "com.google.cloud" % "google-cloud-bigquery" % "1.114.0"
   //the below v1 module is a dependency for v2 because it contains the OAuth scopes necessary to created scoped credentials
-  val googleContainerV1: ModuleID = "com.google.apis" % "google-api-services-container" % s"v1-rev74-$googleV"
+  val googleContainerV1: ModuleID = "com.google.apis" % "google-api-services-container" % "v1-rev20200805-1.30.10"
```
rtitle, line 68: The previous version didn't have `Cluster.setWorkloadIdentityConfig`. This is the latest version, and it has support for WI.

Also I was surprised that `google2` already had the old lib as a dependency. It looks like it's just to get access to the `com.google.api.services.container.ContainerScopes` class. So hopefully this change doesn't increase the dependency footprint of `google2`. Alternatively, we could try to remove this dependency from `google2` entirely and add this functionality to `google1` (which may be more work). Not sure I have a strong preference either way.

project/Settings.scala
```diff
@@ -49,7 +49,7 @@ object Settings {
         "-language:implicitConversions", // Allow definition of implicit functions called views
         "-unchecked", // Enable additional warnings where generated code depends on assumptions.
         "-Xcheckinit", // Wrap field accessors to throw an exception on uninitialized access.
-        "-Xfatal-warnings", // Fail the compilation if there are any warnings.
+//        "-Xfatal-warnings", // Fail the compilation if there are any warnings.
```
rtitle, line 49: Unfortunately `GoogleCredential` is deprecated and generates a warning -- AFAICT we need it to instantiate the old client. 

I looked at this plugin to silence specific warnings but didn't add it: https://github.com/ghik/silencer

Looks like Scala 2.13 has built-in support for this with `@nowarn`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -25,32 +17,27 @@ import scala.concurrent.duration.FiniteDuration
...
+    // are only available in the old client.
+
+    val googleRequest = new com.google.api.services.container.model.CreateClusterRequest()
+      .setCluster(request.cluster)
 
     tracedGoogleRetryWithBlocker(
-      Async[F].delay(clusterManagerClient.createCluster(createClusterRequest)),
-      f"com.google.cloud.container.v1.ClusterManagerClient.createCluster(${request})"
+      Async[F].delay(legacyClient.projects().locations().clusters().create(parent, googleRequest).execute()),
+      f"com.google.api.services.container.Projects.Locations.Cluster(${request})"
```
kyuksel:
```suggestion
      s"com.google.api.services.container.Projects.Locations.Cluster(${request})"
```
since there's no string formatting involved

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -108,6 +110,15 @@ package object google2 {
       credential <- Resource.liftF(Sync[F].delay(UserCredentials.fromStream(credentialFile)))
     } yield credential
 
+  // returns legacy GoogleCredential object which is only used for the legacy com.google.api.services client
+  def googleCredential[F[_]: Sync](pathToCredential: String): Resource[F, GoogleCredential] =
```
kyuksel: similarly to `legacyClient`, I wonder if we should be explicit in naming here too with `legacyGoogleCredential` 🤔

rtitle:
Looks good.. for scalafmt, try doing it on the command line maybe?

`sbt`
`project workbenchGoogle2`
`scalafmt`
`test:scalafmt`

Maybe that'll get Travis to pass (the current failure does indeed seem to be due to scalafmt).

I'd approve but can't since I opened the PR. :) :shipit: once tests pass.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
gpcarr: can update!

gpcarr:
> Looks good.. for scalafmt, try doing it on the command line maybe?
> 
> `sbt`
> `project workbenchGoogle2`
> `scalafmt`
> `test:scalafmt`
> 
> Maybe that'll get Travis to pass (the current failure does indeed seem to be due to scalafmt).
> 
> I'd approve but can't since I opened the PR. :) :shipit: once tests pass.

just committed the scalafmt change, will wait for tests to pass and then merge!

gpcarr:
@rtitle going to go ahead and merge this. for some reason the Travis CI - Pull Request links out to a build off of develop. Here is the link for the build off of my branch with the latest commit https://travis-ci.com/github/broadinstitute/workbench-libs/builds/185051896 
It passed so I'm going to merge!

rtitle:
Sounds good, thanks

gpcarr (approved):

PR #352, opened 2020-09-09
CA-1005 

gpolumbo-broad:
Add method `GoogleDirectoryDAO.addServiceAccountToGroup` to be able to add SAs using their `uniqueId` as an alternative to using their email address.  This is to address a race condition seen when creating a new Service Account in GCP and then immediately trying to add that SA to a Group in GSuite.  Google informed us that by adding an SA using its `uniqueId` instead, we may avoid the race condition, so we shall see.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.workbench.google.GoogleCredentialModes._
 import org.broadinstitute.dsde.workbench.google.GoogleUtilities.RetryPredicates._
 import org.broadinstitute.dsde.workbench.metrics.GoogleInstrumentedService
 import org.broadinstitute.dsde.workbench.model._
+import org.broadinstitute.dsde.workbench.model.google.ServiceAccountSubjectId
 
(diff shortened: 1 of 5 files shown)


marctalbott (approved):
Looks good, reminder to update the hash in the readme and changelog to `TRAVIS-REPLACE-ME`

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -153,6 +154,15 @@ class HttpGoogleDirectoryDAO(appName: String,
 
   override def addMemberToGroup(groupEmail: WorkbenchEmail, memberEmail: WorkbenchEmail): Future[Unit] = {
     val member = new Member().setEmail(memberEmail.value).setRole(groupMemberRole)
+    addMemberToGroup(groupEmail, member)
+  }
+
+  override def addServiceAccountToGroup(groupEmail: WorkbenchEmail, serviceAccount: ServiceAccount): Future[Unit] = {
+    val member = new Member().setId(serviceAccount.subjectId.value).setRole(groupMemberRole)
```
dvoet: I would push the `.setRole(groupMemberRole)` into the private `addMemberToGroup`

dvoet (approved):
one DRY comment then 👍

Qi77Qi (approved):

PR #356, opened 2020-10-15
IA-2242 - returning Poll error for better exception handling in Leo

xuscrbw:
This ticket is the first part to address ticket: https://broadworkbench.atlassian.net/browse/IA-2242

Since we are running into some issues with detaching disks while deleting runtimes, we have added a new PollError class to give us the option to handle errors in clients, specifically Leo. The next step once this work is merged would be to make changes to Leonardo, which would allow us to properly handle and recover from detachDisk issues when deleting runtimes

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.13-39c1b35"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.14-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.14-TRAVIS-REPLACE-ME"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.13-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -13,7 +13,7 @@ Changed:
 Add:
 - `GoogleStorageService.getBucket`
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.13-39c1b35"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.13-TRAVIS-REPLACE-ME"`
 
 ## 0.12
 Changed:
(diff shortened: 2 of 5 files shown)


Qi77Qi:
why don't we use more `handleErrorWith` on the result of the call? just trying to understand the motivation a bit better

xuscrbw:
I wasn't quite familiar with the handleErrorWith function, so I didn't think about using it.
It seemed like pollHelper would be a good place to handle the result of the operation, so it made sense to handle errors within the function.

rtitle:
If I'm understanding correctly, I think the issue was that previously `pollHelper` always did this in case of errors:
```
F.raiseError(new RuntimeException("Operation failed due to: " + op.getError.toString))
```
Leo wants to handle certain errors differently, but we don't want to parse the message of the `RuntimeException`. So this PR adds custom error handling behavior to `pollHelper` (of type `Error => F[A]`).

Qi77Qi:
if we rethrow `op.getError` in https://github.com/broadinstitute/workbench-libs/pull/356/files#diff-6bfffbaa05789c1fdefdaa277aa0f9e8bcb414fbc7702b9e052f65a85b863187R151, that should solve the problem right?

I'm hesitant to add `whenError` because I think it's better to treat the `F` returned by `pollHelper` the usual way. Cuz `handleErrorWith` exists no matter what we do here.

xuscrbw:
Do you have an example of handleErrorWith being used? Just want to see how it is used

Qi77Qi:
@xuscrbw [here](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/LeoPubsubMessageSubscriber.scala#L126)...so I think if all we want is to recover on 404, we can 
1. convert this PR to update [this](https://github.com/broadinstitute/workbench-libs/pull/356/files#diff-6bfffbaa05789c1fdefdaa277aa0f9e8bcb414fbc7702b9e052f65a85b863187R151)
2. in leo, we do `.recoverWith{case e: Error => ???}` (think in this case, you want `recoverWith` becauase `handleErrorWith` tries to handle all type of errors).

project/Settings.scala
```diff
@@ -178,7 +178,7 @@ object Settings {
   val google2Settings = cross212and213 ++ commonSettings ++ List(
     name := "workbench-google2",
     libraryDependencies ++= google2Dependencies,
-    version := createVersion("0.13")
+    version := createVersion("0.14")
```
Qi77Qi: oh..I dont think this PR needs a version bump

README.md
```diff
@@ -60,7 +60,7 @@ To depend on the `MockGoogle*` classes, additionally depend on:
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.13-39c1b35"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.14-TRAVIS-REPLACE-ME"`
```
Qi77Qi: version should be 0.13

google2/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ Changed:
 - `GKEService.createCluster` now uses legacy `com.google.api.services.container` client and model objects
 - `KubernetesModels.KubernetesOperationId` now takes `(operationName: String)` instead of `(operation: Operation)`
 - Made `GoogleComputeService.getDisk` recover on 404s and return `F[Option[Disk]]`
+- `ComputePollOperation.pollHelper` now returns a Poll Error type when the operation fails
+- `ComputePollOperation.PollError` added. Takes an operation and returns the HTTP error message
```
Qi77Qi, line 11: bottom of this sesection, there's a hash needs to be `TRAVIS-REPLACE-ME` as well

Qi77Qi (approved):

PR #358, opened 2020-10-21
[IA-2224] Jc idempotency

jdcanas:
This does the following:

- Makes all methods in GKEInterpreter and KubernetesInterpeter that create/delete kubernetes and GKE resources return none on conflict/404 respectively, making the methods idempotent
- Does a bit of refactoring in places we were using `Async[F]` over `F`

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
@@ -60,7 +60,7 @@ final class GKEInterpreter[F[_]: StructuredLogger: Timer: ContextShift](
     tracedGoogleRetryWithBlocker(
       recoverF(
         F.delay(clusterManagerClient.deleteCluster(clusterId.toString)),
-        whenStatusCode(409)
+        whenStatusCode(404)
       ),
       f"com.google.cloud.container.v1.ClusterManagerClient.deleteCluster(${clusterId.toString})"
     )
@@ -98,7 +98,7 @@ final class GKEInterpreter[F[_]: StructuredLogger: Timer: ContextShift](
     tracedGoogleRetryWithBlocker(
       recoverF(
         F.delay(clusterManagerClient.deleteNodePool(nodepoolId.toString)),
-        whenStatusCode(409)
+        whenStatusCode(404)
       ),
       f"com.google.cloud.container.v1.ClusterManagerClient.deleteNodepool(${nodepoolId.toString})"
     )
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
@@ -4,7 +4,7 @@ import cats.effect.concurrent.Semaphore
 import cats.effect.{Async, Blocker, ContextShift, Timer}
 import cats.mtl.ApplicativeAsk
 import com.google.cloud.container.v1.ClusterManagerClient
-import com.google.container.v1.{Cluster, CreateNodePoolRequest, GetOperationRequest, NodePool, Operation}
+import com.google.container.v1.{Cluster, GetOperationRequest, NodePool, Operation}
 import fs2.Stream
 import io.chrisdavenport.log4cats.StructuredLogger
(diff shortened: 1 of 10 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala
```diff
@@ -30,9 +30,11 @@ object RetryPredicates {
...
 
   def whenStatusCode(code: Int): Throwable => Boolean = {
-    case e: BaseServiceException => e.getCode == code
-    case e: ApiException         => e.getStatusCode.getCode.getHttpStatusCode == code
-    case _                       => false
+    case e: BaseServiceException                                              => e.getCode == code
+    case e: ApiException                                                      => e.getStatusCode.getCode.getHttpStatusCode == code
+    case e: com.google.api.client.googleapis.json.GoogleJsonResponseException => e.getDetails.getCode == code
+    case e: io.kubernetes.client.ApiException                                 => e.getCode == code
+    case _                                                                    => false
```
jdcanas, line 37: this is the main logic I needed to add to get this additional functionality

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -15,17 +15,18 @@ import cats.implicits._
...
   blockerBound: Semaphore[F],
   retryConfig: RetryConfig
-) extends GKEService[F] {
+)(implicit F: Async[F])
+    extends GKEService[F] {
 
   override def createCluster(
     request: KubernetesCreateClusterRequest
-  )(implicit ev: ApplicativeAsk[F, TraceId]): F[com.google.api.services.container.model.Operation] = {
+  )(implicit ev: ApplicativeAsk[F, TraceId]): F[Option[com.google.api.services.container.model.Operation]] = {
```
Qi77Qi, line 29: We did something similar for delete before, where `None` case means the resource doesn't exist in Google, hence htere's no operation returned.

what does it mean to return None here for createCluster?

jdcanas: Oops, typo in the PR description. Returning None here means that the resource already exists

Qi77Qi: ok I think that makes sense...can you add a comment for this.

Also for delete, shall we add 404 as well, which means if the nodepool doesn't exist...so basically for delete, None can either mean there's already an operation or the resource doesn't exist

jdcanas: Ah, yeah it should be 404 > 409 for delete idempotency

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEService.scala
```diff
@@ -22,22 +22,27 @@ import scala.concurrent.duration.FiniteDuration
 
 trait GKEService[F[_]] {
 
+  // These methods return Option[Operation] so as to return None on 409 in the case of create and 404 in the case of delete
```
kyuksel, line 25: Can we also add a line regarding the meaning of returning `None`?

kyuksel (approved):
LGTM 👍  with a minor comment inline

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEInterpreter.scala
```diff
@@ -36,53 +37,69 @@ final class GKEInterpreter[F[_]: Async: StructuredLogger: Timer: ContextShift](
...
 
-  override def deleteCluster(clusterId: KubernetesClusterId)(implicit ev: ApplicativeAsk[F, TraceId]): F[Operation] =
+  override def deleteCluster(
+    clusterId: KubernetesClusterId
+  )(implicit ev: ApplicativeAsk[F, TraceId]): F[Option[Operation]] =
     tracedGoogleRetryWithBlocker(
-      Async[F].delay(clusterManagerClient.deleteCluster(clusterId.toString)),
+      recoverF(
+        F.delay(clusterManagerClient.deleteCluster(clusterId.toString)),
+        whenStatusCode(404)
```
Qi77Qi, line 63: doesn't this need to recover from 409 as well?

PR #360, opened 2020-10-30
[IA-2224] Jc disk idempotency

jdcanas:
- fix travis issue
- make create disk return Option[Operation], which returns None in the case the disk already exists

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
@@ -17,6 +17,7 @@ import io.chrisdavenport.log4cats.StructuredLogger
 import org.broadinstitute.dsde.workbench.RetryConfig
 import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates._
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDiskInterpreter.scala
```diff
@@ -64,13 +59,7 @@ private[google2] class GoogleDiskInterpreter[F[_]: StructuredLogger: Timer: Cont
...
   ): F[Option[Operation]] = {
     val projectZoneDiskName = ProjectZoneDiskName.of(diskName.value, project.value, zone.value)
-    val fa = F
-      .delay(diskClient.deleteDisk(projectZoneDiskName))
-      .map(Option(_))
-      .handleErrorWith {
-        case _: com.google.api.gax.rpc.NotFoundException => F.pure(none[Operation])
-        case e                                           => F.raiseError[Option[Operation]](e)
-      }
+    val fa = recoverF(F.delay(diskClient.deleteDisk(projectZoneDiskName)), whenStatusCode(404))
```
Qi77Qi, line 62: did you test the code? I know the original code works, but jsut want to make sure `whenStatusCode` is actually recovering from the right exception

jdcanas: Yeah, i had trouble testing it in wb-libs but I tested in Leo since its a requirement for idempotency. It's also guaranteed by type since [here](https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala#L36) we are checking for the type of the exception that was here

Qi77Qi: I don't think the type you're referencing is correct tho...this is disk, not kubernetes...what I'm worried is it's a different exception type that's thrown here from what you have in `whenStatusCode`

jdcanas: Sorry, I linked the wrong line, this is the relevant one: https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala#L34.

Notice the import here https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala#L5

jdcanas: I also tested it in leo by resubmitted a pubsub message that includes creating/deleting an app with a disk (with create/delete disk specified). The way I found this needed to be done was actually encountering this error in the create case.

Qi77Qi (approved):

PR #362, opened 2020-11-03
IA-2337 try use github action for travis replace me

Qi77Qi:
1. Use github workflow for publishing jars
2. Move away from using @jmthibault79 's personal token
3. Use github workflow for running unit tests

**PR checklist**

In all cases:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,41 @@
...
+          # Default: author of the commit that triggered the run
+          author_name: broadbot
+
+          # The email of the user that will be displayed as the author of the commit
+          # Default: author of the commit that triggered the run
+          author_email: broadbot@broadinstitute.org
+
+          # The message for the commit
+          # Default: 'Commit from GitHub Actions (name of the workflow)'
+#          message: 'Auto update hash in ${f} to ${HASH}'
```
rtitle, line 39: Should this be uncommented? Also does this actually push the replaced hash back to `develop`?

rtitle (approved):

.github/workflows/publish.yml
Qi77Qi: there will be a default message published by the action that adds all changed files in one commit. So it'll be different from what it was before.

commit will be pushed to whichever branch that triggered the workflow. Since this workflow is configed only triggerd when there's push to `develop`, so that's when the commit will be pushed

I'll remove the message in another PR

PR #372, opened 2020-11-16
[IA-2355] support maxRetry

Qi77Qi:
support optional subscriber name and optional `deadLetterPolicy`

Tested in console:
```
qi | INFO: Subscriber Successfully received data: "\"ooo\""
attributes {
  key: "googclient_deliveryattempt"
  value: "5"
}
message_id: "501898938879014"
publish_time {
  seconds: 1605568147
  nanos: 653000000
}
.
```

After 5 retries, msg is forwarded to dead letter topic that I configed, I see the nacked message as expected.

**PR checklist**

In all cases:

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.16
+
+Changed:
+- Add `subscriptionName: Option[ProjectSubscriptionName]` and `deadLetterPolicy: Option[SubscriberDeadLetterPolicy]` to `SubscriberConfig`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.16-TRAVIS-REPLACE-ME"`
+
 ## 0.15
 
 Added:
@@ -10,9 +17,8 @@ Added:
 
 Changed:
 - Upgrade `cats-mtl` to `1.0.0`
-- Fix a bug where `maxRetry` in SubscriberConfig not respected
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.15-TRAVIS-REPLACE-ME"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.15-426a0c2"`
 
 ## 0.14
 Changed:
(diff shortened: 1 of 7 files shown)


google2/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.16
+
+Changed:
+- Add `subscriptionName: Option[ProjectSubscriptionName]` and `deadLetterPolicy: Option[SubscriberDeadLetterPolicy]` to `SubscriberConfig`
```
rtitle, line 8: Just curious - since these are optional could it be a non-breaking change if they default to `None`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -61,25 +62,23 @@ object GoogleSubscriberInterpreter {
...
-            Effect[F].raiseError[MessageType](new Exception(s"${message.getData.toStringUtf8} is not a valid Json"))
+            F.raiseError[MessageType](new Exception(s"${message.getData.toStringUtf8} is not a valid Json"))
           case Right(json) =>
-            Effect[F].fromEither(json.as[MessageType])
+            F.fromEither(json.as[MessageType])
         }
         traceId = Option(message.getAttributesMap.get("traceId")).map(s => TraceId(s))
       } yield Event(msg, traceId, message.getPublishTime, consumer)
 
-      val loggingCtx = Map(
```
rtitle, line 79: Is this context not useful to log, or is it logged in another way?

Qi77Qi: when we log the `PubsubMessage` in the message, these context are already being logged. So it's a bit duplicate

google2/CHANGELOG.md
Qi77Qi: ah yea, I guess (I'm not entirely sure by providing default is backward compatible since I vaguely remember I had to fix things after a PR that added optional param with default...)...I tend to not giving default just so caller have to think about these options. But I'm happy to give default here..also [This](https://blog.ssanj.net/posts/2019-05-01-why-are-default-parameter-values-considered-bad-in-scala.html) has more points on why providing defaults can be bad

rtitle: Yeah I don't love defaults either, was just curious. I don't really have a strong preference, will thumb as is :)

rtitle (approved):

PR #373, opened 2020-11-17
[IA-2350] [google2] Ability to resize and stop Dataproc clusters

kyuksel:
- Addresses [IA-2350](https://broadworkbench.atlassian.net/browse/IA-2350)
- Required by [IA-2229](https://broadworkbench.atlassian.net/browse/IA-2229)
- Punting on adding `startCluster` in favour of getting in sooner the Resource Validator functionality, which doesn't require `startCluster`
- Console-tested resizing and stopping of various types of clusters

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -84,6 +87,122 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+      else F.pure(none[ClusterOperationMetadata])
+
+      // Then, stop each instance individually
+      _ <- instances.toList.parTraverse { instance =>
+        if (instance.dataprocRole == Master && metadata.nonEmpty) {
+          googleComputeService.addInstanceMetadata(
+            instance.project,
+            instance.zone,
+            instance.name,
+            metadata.get // TODO: Refactor
```
Qi77Qi: 1. metadata.traverse {m => ???}


2. if statement and `googleComputeService.stopInstance` needs to be in a for comprehension. `addInstanceMetadata` will never be run like this

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -84,6 +87,122 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+   * Strictly speaking, it is not possible to 'stop' a Dataproc cluster altogether.
+   * Instead, we approximate by:
+   *   1. removing pre-emptible instances (if any) by resizing the cluster, since they would not be possible to restart
+   *   2. stopping underlying nodes individually
+   */
+  override def stopCluster(project: GoogleProject,
+                           region: RegionName,
+                           clusterName: DataprocClusterName,
+                           instances: Set[DataprocInstance],
+                           numWorkers: Option[Int] = None,
```
Qi77Qi: I'd prefer to not have defaults, so that caller has to think about whether they want to pass in workers, preemptibles, etc

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
```diff
@@ -86,12 +120,19 @@ object GoogleDataprocService {
...
       client <- backgroundResourceF(ClusterControllerClient.create(settings))
-    } yield new GoogleDataprocInterpreter[F](client, blocker, blockerBound, retryConfig)
+    } yield new GoogleDataprocInterpreter[F](client, googleComputeService, blocker, blockerBound, retryConfig)
   }
 }
 
 final case class DataprocClusterName(value: String) extends AnyVal
-final case class ClusterErrorDetails(code: Int, message: Option[String])
+
+case class DataprocInstance(name: InstanceName,
```
Qi77Qi: add `final`, especially for data types defined in library

Qi77Qi:
I would also manually test it out in console

```
import org.broadinstitute.dsde.workbench.google2._

import scala.concurrent.ExecutionContext.global
import io.chrisdavenport.log4cats.slf4j.Slf4jLogger
import cats.effect.{Blocker, IO, Resource}
import cats.mtl.Ask
import java.util.UUID

import cats.implicits._
import org.broadinstitute.dsde.workbench.model.TraceId
import cats.effect.concurrent.Semaphore
import com.google.cloud.Timestamp
import com.google.cloud.compute.v1.Disk
import com.google.pubsub.v1.{ProjectSubscriptionName, TopicName}
import fs2.{Pipe, Stream}
import fs2.concurrent.InspectableQueue
import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject}
import org.broadinstitute.dsde.workbench.google2.GoogleDataprocInterpreter
import org.broadinstitute.dsde.workbench.DoneCheckableInstances.computeDoneCheckable
import org.broadinstitute.dsde.workbench.util2.{ConsoleLogger, LogLevel}

implicit def logger = new ConsoleLogger("qi", LogLevel(true, true, true, true))
implicit val traceId = Ask.const[IO, TraceId](TraceId(UUID.randomUUID()))
val blocker = Blocker.liftExecutionContext(global)
val blockerBound = Semaphore[IO](10).unsafeRunSync

val pathToCredentialStorage = ???
val regionName = RegionName("us-central1")
val zoneName = ZoneName("us-central1-a")

val googleProject = GoogleProject(???)

GoogleDataprocService.resource(pathToCredentialStorage, blocker, blockerBound, regionName).use {s => ???}
```

.gitignore
```diff
@@ -1,4 +1,3 @@
-qi-worksheet.sc
```
Qi77Qi: can you add this back :)

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-0e2d8cc"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.18-TRAVIS-REPLACE-ME"`
```
jdcanas, line 75: don't forget to bump the version in `Settings.scala` since you're incrementing it! The CI won't build it otherwise.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -197,10 +197,18 @@ object GoogleComputeService {
...
 final case class FirewallRuleName(value: String) extends AnyVal
 final case class MachineTypeName(value: String) extends AnyVal
-final case class RegionName(value: String) extends AnyVal
 final case class NetworkName(value: String) extends AnyVal
 final case class SubnetworkName(value: String) extends AnyVal
 final case class OperationName(value: String) extends AnyVal
+
+final case class RegionName(value: String) extends AnyVal
+object RegionName {
+  def fromUri(uri: String): Option[RegionName] = Option(uri).flatMap(_.split("/").lastOption).map(RegionName(_))
```
Qi77Qi: this looks to me should be called `fromString`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
```diff
@@ -197,10 +197,18 @@ object GoogleComputeService {
...
+}
+
+final case class ZoneName(value: String) extends AnyVal
+object ZoneName {
+  def fromUri(uri: String): Option[ZoneName] = Option(uri).flatMap(_.split("/").lastOption).map(ZoneName(_))
```
Qi77Qi: same as above

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -83,6 +89,141 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+  )(implicit
+    ev: Ask[F, TraceId]
+  ): F[List[Operation]] =
+    for {
+      clusterInstances <- getClusterInstances(project, region, clusterName)
+
+      // First, remove preemptible instances (if any) and wait until the removal is done
+      remainingClusterInstances <-
+        if (containsPreemptibles(clusterInstances))
+          resizeCluster(project, region, clusterName, numWorkers = None, numPreemptibles = Some(0)) >> streamFUntilDone(
```
Qi77Qi: are `numWorkers` and `numPreemptibles` supposed to be hardcoded?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
jdcanas: I don't think you need `RegionName(_)`, it can just be `RegionName` I believe. Underscores look messy 😄

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
jdcanas: hmm... I think we're running into the issue here that `streamFUntilDone` doesn't actually error if the done condition is not met when it exits. You have to check the return to ensure your condition is met. It may be a good time to introduce a version of `streamFUntilDone` that explicitly does `raiseError` if it is exiting without its condition being met here. Not too sure what a good name for it would be.... Almost seems like the existing method should be called `streamFUntilDoneOrTimeout`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -83,6 +89,141 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+            getClusterInstances(project, region, clusterName),
+            15,
+            6 seconds
+          ).compile.lastOrError
```
Qi77Qi: `streamFUntilDone` doesn't guarantee success, so you'd still need to check if the result is "Done"

We talked about adding `streamUntilSuccess` at one point, maybe now is a good time to add that :)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -83,6 +89,141 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+                             numWorkers: Option[Int],
+                             numPreemptibles: Option[Int]
+  )(implicit
+    ev: Ask[F, TraceId]
+  ): F[Option[ClusterOperationMetadata]] = {
+    val workerMask = "config.worker_config.num_instances"
+    val preemptibleMask = "config.secondary_worker_config.num_instances"
+
+    val configAndMask = (numWorkers, numPreemptibles) match {
+      case (Some(nw), Some(np)) =>
```
jdcanas, line 168: it may be cleaner to have a separate match for each `numWorkers` and `numPreemptiples` instead of one match with 4 cases and bundling both resultant items in a tuple. Less cognitive load than one big match, I don't feel too strongly though.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
```diff
@@ -111,9 +150,12 @@ sealed trait DataprocRole extends Product with Serializable
 object DataprocRole {
   case object Master extends DataprocRole
   case object Worker extends DataprocRole
-  case object SecondaryWorker extends DataprocRole // alias to Preemptible Worker
+  case object SecondaryWorker extends DataprocRole
 
   val stringToDataprocRole = sealerate.values[DataprocRole].map(p => (p.toString, p)).toMap
 }
 
+final case class DataprocRoleZonePreemptibility(role: DataprocRole, zone: ZoneName, isPreemptible: Boolean)
```
rtitle, line 158: Isn't `isPreemptible` duplicative since we already have `DataprocRole.SecondaryWorker`?

Qi77Qi: Kaan explained to me this..it looks like SecondaryWorker is just the role, and it can be either preemptible or non preemptible (you can config in console)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
rtitle: +1 good catch, that has bit us before.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -156,6 +159,16 @@ object DoneCheckableInstances {
   implicit val computeDoneCheckable = new DoneCheckable[com.google.cloud.compute.v1.Operation] {
     def isDone(op: com.google.cloud.compute.v1.Operation): Boolean = op.getStatus == "DONE"
   }
+
+  implicit val resizingDoneCheckable = new DoneCheckable[Map[DataprocRoleZonePreemptibility, Set[InstanceName]]] {
```
rtitle: These are pretty implementation-specific, wondering if they should be defined in `DataprocInterpreter` and not `package.scala`…

.gitignore
kyuksel: okay <sigh> 😛

README.md
kyuksel: thanks!

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeService.scala
kyuksel: Agreed. I had also tried without the underscore first but the compiler seems to infer it to be a `RegionName.type` instead of `RegionName` for some reason and fails. Not sure if I'm missing something.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: For the cluster stopping case, we pass in `numWorkers = None` and `numPreemptibles = Some(0)` because that tells `resizeCluster()` to not change the `worker` config and to size down the number of `secondary worker`s to `0`, which is what we want to do. Is that what you were asking?

kyuksel: Thanks for the catch. I'd forgotten about this 🤦

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: Maybe..we'll see 😛

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: Hmmm...I could see an argument both ways. I may try the version you suggested and keep it if it looks easier to follow 😊

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
kyuksel: Yea, our implementation had been equating `secondary worker` to `preemptible secondary worker`, which is not accurate (more info [here](https://cloud.google.com/dataproc/docs/concepts/compute/secondary-vms#preemptible_and_non-preemptible_secondary_workers)). This PR doesn't correct that but I figured I'd at least not contribute to the misuse further, and if we ever want to expose to users the option of `non-preemptible secondary worker`s, we could make the other necessary changes then 🤷‍♂️

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
kyuksel: Sure. I was on the fence myself and thought perhaps it'd be better organized if I put it together with the previously existing ones. I can move it back.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
rtitle: :mindblown: 🤯 

I always thought they were the same. Good to know! Our code certainly assumes `SecondaryWorker` == preemptible.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: I played with another version that didn't help with my cognitive load so I decided to keep it as is  🤷‍♂️

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -162,6 +182,8 @@ package object google2 {
   )
 }
 
+case object StreamTimeoutError extends NoStackTrace
```
Qi77Qi: How about 
```
final case class StreamTimeoutError(streamAction: String)
```
?

This way, you don't need to `adaptError` every time just to tell what action it timed out

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -149,11 +150,31 @@ package object google2 {
...
   def streamFUntilDone[F[_]: Timer, A: DoneCheckable](fa: F[A], maxAttempts: Int, delay: FiniteDuration): Stream[F, A] =
     (Stream.sleep_(delay) ++ Stream.eval(fa))
       .repeatN(maxAttempts)
       .takeThrough(!_.isDone)
 
+  // Distinctly from the method streamFUntilDone(), this method raises a StreamTimeoutError if the Done condition is not met
+  // by the time maxAttempts are exhausted. Therefore callers may want to use this method instead of streamFUntilDone()
+  // if they want to ascertain the Done condition was satisfied and take action otherwise.
+  // See org.broadinstitute.dsde.workbench.google2.GoogleDataprocInterpreter.stopCluster() for examples.
+  def streamUntilDoneOrTimeout[F[_]: Sync: Timer, A: DoneCheckable](fa: F[A],
```
rtitle, line 165: Cool, glad we have this -- I think other places in Leo could use it too

rtitle:
Late to this party but these changes LGTM 👍

PR #377, opened 2020-11-23
IA-1790 add list to GoogleTopicAdmin

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1790

Add steward config to control PR frequency

**PR checklist**
- [ ] For each library you've modified here, decide whether it requires a major, minor, or no version bump. (Click [here](/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md) for guidance)

If you're doing a **major** or **minor** version bump to any libraries:
- [ ] Bump the version in `project/Settings.scala` `createVersion()`
- [ ] Update `CHANGELOG.md` for those libraries
- [ ] I promise I used `@deprecated` instead of deleting code where possible

In all cases:
- [ ] Replace the appropriate version hashes in `README.md` and the `CHANGELOG.md` for any libs you changed with `TRAVIS-REPLACE-ME` to auto-update the version string
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green (CI _and_ coverage tests)
- [ ] Squash commits and **merge to develop**
- [ ] Delete branch after merge

PR #417, opened 2020-11-25
[no ticket] upgrade cats-core to 2.2.0

Qi77Qi:
1. Bump cats-core version
2. Bump http4s version to fix a regression https://github.com/http4s/http4s/releases/tag/v0.21.13

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-0e2d8cc"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -26,7 +26,7 @@ Update guava to 30.0-jre (#390)
 Update `io.kubernetes client-java` to `5.0.0` to `10.0.0` (This has some breaking changes)
 ```
       
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-0e2d8cc"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-TRAVIS-REPLACE-ME"`
 
 ## 0.16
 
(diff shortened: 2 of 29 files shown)


google2/CHANGELOG.md
```diff
@@ -23,10 +23,10 @@ Update metrics4-scala to 4.1.14 (#413)
 Update http4s-blaze-client, http4s-circe, ... to 0.21.12 (#415)
 Update mockito-core to 3.6.28 (#414)
 Update guava to 30.0-jre (#390)
-Update `io.kubernetes client-java` to `5.0.0` to `10.0.0` (This has some breaking changes)
+Update `io.kubernetes client-java` to `5.0.0` to `10.0.0` (This has some breaking changes if you're using the library's API directly)
```
kyuksel: Is there a typo?
```suggestion
Update `io.kubernetes client-java` from `5.0.0` to `10.0.0` (This has some breaking changes if you're using the library's API directly)
```

kyuksel: > using the library's API directly

We _are_ using it directly (e.g. `io.kubernetes.client.openapi.apis.CoreV1Api`), correct?

Qi77Qi: nope

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-0e2d8cc"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-TRAVIS-REPLACE-ME"`
```
rtitle, line 75: Should this be a version bump, since it'll force dependent code to update their imports?

Qi77Qi: I didn't bump it since it was pretty close to the https://github.com/broadinstitute/workbench-libs/pull/377 which bumped things to 0.17....the kubernetes library bump only affects us and we don't actually reference these classes directly...

oh, also, it doesn't force ppl to update cats imports if that's what you're referring to

Happy to bump the version tho

rtitle: Oh okay I was thinking it forced clients to do `import cats.syntax.all._`. If that's not the case then no need for a version bump I think.

rtitle (approved):

README.md
Qi77Qi: yea, cats 2.2.0 is binary compatible with previous versions ([release notes](https://github.com/typelevel/cats/releases/tag/v2.2.0))

I think the benefit is with `import cats.syntax.all._` in latest cats version, you won't need to import those implicits instances (cats.implicits._ mixes in both syntax implicits and instance implicits)

PR #418, opened 2020-11-30
Set location for bucket during workspace creation [BW-427][BW-443]

salonishah11:
**PR checklist**

In all cases:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,15 @@ We run coverage tests on this repo. Please try to avoid merging PRs that fail co
 
 Defining your mocks in `src/test` rather than `src/main` will help greatly.
 
+## Linting
+
+The project uses [Scalafmt](https://scalameta.org/scalafmt/docs/installation.html) for linting. If there is a formatting error 
+during `Compile and check scalafmt` step during the `Unit tests` check for a PR, the below command will make the required
+formatting changes.
+```
+sbt -Denv.type=test "test:compile" scalafmtAll
+```
+
 ## Publishing
 
 Travis automatically builds Scala 2.12 and 2.13 versions of these libraries and publishes them to [Artifactory](https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-release-local/org/broadinstitute/dsde/workbench/). Commits to any other branch than develop will be labelled `<version>-<githash>-SNAP`; commits to develop will be labelled `<version>-<githash>`. In both cases, the `<githash>` is the first 7 characters of the commit hash. You probably shouldn't be using `-SNAP` versions in downstream code.
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.18-200cb94"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.18-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
@@ -106,7 +106,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains common classes and utilities for writing tests against Workbench REST web services.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.18-d7bebc7" % "test" classifier "tests"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "0.18-TRAVIS-REPLACE-ME" % "test" classifier "tests"`
 
 [Changelog](serviceTest/CHANGELOG.md)
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -316,7 +316,7 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
       bucketInfoBuilder.setLogging(logging)
     }
 
-    // set the location is passed else the bucket will default to `us multi-region`
+    // set the location if passed, else the bucket will default to `us multi-region`
     location.foreach(bucketInfoBuilder.setLocation)
 
     val bucketInfo = acl
@@ -328,8 +328,6 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
           .build()
       }
       .getOrElse(bucketInfoBuilder.build())
-
-    println(s"**** FIND ME: BUCKET NAME- ${bucketInfo.getName}\t BUCKET LOCATION- ${bucketInfo.getLocation}")
 
     val dbForProject = db.getOptions.toBuilder.setProjectId(googleProject.value).build().getService
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -292,7 +292,8 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
                             traceId: Option[TraceId] = None,
                             bucketPolicyOnlyEnabled: Boolean = false,
                             logBucket: Option[GcsBucketName] = None,
-                            retryConfig: RetryConfig
+                            retryConfig: RetryConfig,
+                            location: Option[String] = None
   ): Stream[F, Unit] = {
 
     if (acl.isDefined && bucketPolicyOnlyEnabled) {
@@ -315,6 +316,9 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
       bucketInfoBuilder.setLogging(logging)
     }
 
+    // set the location is passed else the bucket will default to `us multi-region`
+    location.foreach(bucketInfoBuilder.setLocation)
+
     val bucketInfo = acl
       .map { aclList =>
         val acls = aclList.toList.asJava
@@ -324,6 +328,8 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
           .build()
       }
       .getOrElse(bucketInfoBuilder.build())
+
+    println(s"**** FIND ME: BUCKET LOCATION- ${bucketInfo.getLocation}")
 
     val dbForProject = db.getOptions.toBuilder.setProjectId(googleProject.value).build().getService
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -329,7 +329,7 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
       }
       .getOrElse(bucketInfoBuilder.build())
 
-    println(s"**** FIND ME: BUCKET LOCATION- ${bucketInfo.getLocation}")
+    println(s"**** FIND ME: BUCKET NAME- ${bucketInfo.getName}\t BUCKET LOCATION- ${bucketInfo.getLocation}")
 
     val dbForProject = db.getOptions.toBuilder.setProjectId(googleProject.value).build().getService
 
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
@@ -32,10 +32,11 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
                     authDomain: Set[String] = Set.empty,
                     aclEntries: List[AclEntry] = List(),
                     attributes: Option[Map[String, Any]] = None,
-                    cleanUp: Boolean = true
+                    cleanUp: Boolean = true,
+                    workspaceRegion: Option[String] = None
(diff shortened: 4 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -61,6 +61,15 @@ We run coverage tests on this repo. Please try to avoid merging PRs that fail co
 
 Defining your mocks in `src/test` rather than `src/main` will help greatly.
 
+## Linting
+
+The project uses [Scalafmt](https://scalameta.org/scalafmt/docs/installation.html) for linting. If there is a formatting error 
+during `Compile and check scalafmt` step during the `Unit tests` check for a PR, the below command will make the required
+formatting changes.
+```
+sbt -Denv.type=test "test:compile" scalafmtAll
```
cjllanwarne, line 70: Nice. Bravo! 💯

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-1e1f697"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.17-TRAVIS-REPLACE-ME"`
```
cjllanwarne: Does `TRAVIS-REPLACE-ME` belong in the README.md?

cjllanwarne: (and several times below in the PR)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -292,7 +292,8 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
                             traceId: Option[TraceId] = None,
                             bucketPolicyOnlyEnabled: Boolean = false,
                             logBucket: Option[GcsBucketName] = None,
-                            retryConfig: RetryConfig
+                            retryConfig: RetryConfig,
+                            location: Option[String] = None
```
cjllanwarne, line 296: TOL: If you're feeling adventurous, this `String` could become an ADT or Refined data type to avoid the possibility of someone passing a wrong/unexpected value into this function in the first place.

If you're not, it still might be worth considering putting the "support location validation" logic here rather than in Rawls?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -53,11 +39,65 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
...
+   * test code is run. The workspace name will contain a random and highly
+   * likely unique series of characters.
+   *
+   * @param namespace the namespace for the test workspace
+   * @param namePrefix optional prefix for the workspace name
+   * @param authDomain optional auth domain for the test workspace
+   * @param testCode test code to run
+   * @param token auth token for service API calls
+   * @param bucketLocation optional region where the bucket associated with workspace should be created
+   */
```
rsasch: please update to reflect the actual `withWorkspace` method and its parameters

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -53,11 +39,65 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
...
+  def withClonedWorkspace(
```
cjllanwarne, line 86: TOL: my brain parses this as "with a clone of an existing workspace" instead of "with a clone of a new workspace". I wonder whether you could name it `withClonedNewWorkspace` to make that more obvious? If that feels clunky to you, disregard this comment 😄

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -53,11 +39,65 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
...
+   * up after the test code is run. The workspace names will contain a random and highly
+   * likely unique series of characters.
+   *
+   * @param namespace the namespace for the test workspace
+   * @param namePrefix optional prefix for the workspace name
+   * @param authDomain optional auth domain for the test workspace
+   * @param testCode test code to run
+   * @param token auth token for service API calls
+   * @param bucketLocation optional region where the bucket associated with workspace should be created
+   */
```
rsasch: please update to reflect the actual `withClonedWorkspace` method and its parameters

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -16,26 +16,12 @@ import scala.util.Try
...
-   * @param testCode test code to run
-   * @param token auth token for service API calls
-   */
-  def withWorkspace(namespace: String,
-                    namePrefix: String,
-                    authDomain: Set[String] = Set.empty,
-                    aclEntries: List[AclEntry] = List(),
-                    attributes: Option[Map[String, Any]] = None,
-                    cleanUp: Boolean = true
+  private def setupWorkspace(namespace: String,
```
cjllanwarne: I know it's not public but a comment might still help to describe what this is doing (set some workspace attributes and ACLs?)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -53,11 +39,65 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
...
+    val workspaceName = uuidWithPrefix(namePrefix, " ")
+    Orchestration.workspaces.create(namespace, workspaceName, authDomain, bucketLocation)
+
+    setupWorkspace(namespace, workspaceName, aclEntries, attributes, cleanUp) { _ =>
+      // clone the newly created workspace
+      val cloneNamePrefix = appendUnderscore(workspaceName) + "clone"
+      val clonedWorkspaceName = uuidWithPrefix(cloneNamePrefix, " ")
+
+      Orchestration.workspaces.clone(namespace, workspaceName, namespace, clonedWorkspaceName, authDomain)
+      setupWorkspace(namespace, clonedWorkspaceName, aclEntries, attributes, cleanUp)(testCode)
```
cjllanwarne: Kind of surprising to me that the cloned workspace doesn't automatically inherit the ACLs and attributes?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -153,20 +153,24 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
 
-    def create(namespace: String, name: String, authDomain: Set[String] = Set.empty)(implicit
-      token: AuthToken
-    ): Unit = {
+    def create(namespace: String,
+               name: String,
+               authDomain: Set[String] = Set.empty,
+               bucketLocation: Option[String] = None
+    )(implicit token: AuthToken): Unit = {
       logger.info(s"Creating workspace: $namespace/$name authDomain: $authDomain")
```
rsasch: TOL — as long as we're logging, maybe add the optional bucket location?

rsasch (approved):
just some small peripheral comments, otherwise 👍

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
cjllanwarne: I think this should be called `withSetUpWorkspace`

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
```diff
@@ -153,20 +153,24 @@ trait Orchestration extends RestClient with LazyLogging with SprayJsonSupport wi
...
-                        "attributes" -> Map.empty,
-                        "authorizationDomain" -> authDomainGroups
+      val request: Map[String, Object] = Map("namespace" -> namespace,
+                                             "name" -> name,
+                                             "attributes" -> Map.empty,
+                                             "authorizationDomain" -> authDomainGroups
       )
 
-      postRequest(apiUrl(s"api/workspaces"), request)
+      val requestWithRegion: Map[String, Object] = request ++ bucketLocation.map("bucketLocation" -> _)
```
cjllanwarne: TOL: could you inline the `++ bucketLocation.map("bucketLocation" -> _)` onto the end of `val request = ...` line above?

cjllanwarne:
I see @kcibul 's PR includes `storageClass: Option[StorageClass]` alongside the `location: Option[String]` parameter. Do we know why that was necessary for the hackathon project but not for us today?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Orchestration.scala
cjllanwarne: Yes please!

salonishah11: I don't see any harm in adding it 👍

README.md
salonishah11: Yeah so the way it works is that once the PR is merged Github Actions will run a job and publish the jars to Artifactory and replace `TRAVIS-REPLACE-ME` with the new hashes. 
References:
- [Broadbot commit through Github actions](https://github.com/broadinstitute/workbench-libs/commit/835b619070ca4c42a1ec067d26956edc5d3bd377)
- [Sharing Artifacts](https://github.com/broadinstitute/workbench-libs/blob/develop/GETTING_STARTED.md#sharing-artifacts-with-the-world)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
salonishah11: I understand your point. But we are not sure which service apart from Rawls is using it. Changing the name would be a breaking change. And since it would be a breaking change, according to the [Contributing](https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md#contributing) doc, we would have to create a new method, mark this one as deprecated and also do a major version bump. That feels like too much work for a name change.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
salonishah11: Yes that is correct. I missed that while writing the code for cloning workspace. Updated it 👍

salonishah11:
@cjllanwarne 
>I see @kcibul 's PR includes storageClass: Option[StorageClass] alongside the location: Option[String] parameter. Do we know why that was necessary for the hackathon project but not for us today?

As far as we know, we only wanted to add the ability to have the bucket created in a different region but not using a non-standard storage class. By default the buckets will be created with storage class set to `Standard`.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
cjllanwarne: I didn't realize this was already an existing function, I thought this was a new one we were adding. Sounds good.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -16,48 +16,99 @@ import scala.util.Try
  */
 trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSuite =>
 
+  /**
+   * Cleans up the workspace once the test code has been run
+   */
+  private def withRunAndCleanUpWorkspace(namespace: String, workspaceName: String, cleanUp: Boolean)(
```
cjllanwarne: TOL (since I already seagulled this once before, probably incorrectly):

Sorry to go back on what I said before - this new layout makes it clearer to me that this function is not actually _creating_ the workspace, just using it for a test then cleaning it up. I guess `withXyz` is actually less good as a name than just `runAndCleanUpWorkspace` would have been.

But since it's now so simple, I wonder whether what we really want is to standardize the cleanup function, like:
```
def cleanupTrial(namespace, workspaceName, cleanup) = Try {
  if (cleanUp) {
    Orchestration.workspaces.delete(namespace, workspaceName)
  }
}
```

and then just use `CleanUp` directly in the two main methods, like:
```scala
def withWorkspace(...) = {
  [...]
  CleanUp.runCodeWithCleanup(
    Try(testCode(workspaceName)), 
    cleanupTrial(namespace, workspaceName, cleanup)
  )
}
```

cjllanwarne (approved):
Minor quibble about the function name I asked you to add (🤦 ) but that's ignorable if you're fed up with me, and otherwise it all LGTM.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
salonishah11: I have updated the name to `runTestAndCleanUpWorkspace` to make it more clear what the function is doing. Regarding the refactoring: I think it makes more sense to keep it the way it is. Making the `cleanupTrial()` function simpler kind of complicates things in the `withClonedWorkspace()` function.

cjllanwarne:
Re-👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
Qi77Qi: ditto what @cjllanwarne said....that'd be really nice! fwiw, there's a custom type for Location already in `google2` I think [here](https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleKmsService.scala#L43) (it probably can be moved to a better place 🤷‍♀️ )

PR #423, opened 2020-12-08
AS-535: Add `overrideIamPolicy` to GoogleStorageService

MatthewBemis:
**PR checklist**

In all cases:


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -447,6 +440,28 @@ private[google2] class GoogleStorageInterpreter[F[_]: ContextShift: Timer](
     )
   }
 
+  override def overrideIamPolicy(bucketName: GcsBucketName,
```
MatthewBemis, line 450: Our implementation of `setIamPolicy` behaves much differently than the underlying Google implementation of `setIamPolicy`. Ours gets all existing members and adds any entries that were passed in, i.e. it is only additive. Google's does a strict override- the IAM policy that you pass it becomes the new IAM policy.

This method is analogous to Google's implementation of `setIamPolicy`

dvoet (approved):
I would be inclined to change the name of `setIamPolicy` to `addIamPolicy` but don't hold this PR for that

PR #461, opened 2021-01-04
[no ticket] Drop 2.12 for notification, errorReporting, openTelemetry modules

Qi77Qi:
* `workbench-notification` is only used by Sam, which is on Scala 2.13 now. Hence we can drop 2.12 for it.
* `workbench-errorReporting` and `workbench-openTelemetry` is added relatively recently and only used by Leonardo. Hence I think it's ok to drop 2.12 as well.
* Update `google2` hash for a bunch dependency updates

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-11a45ad"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -161,13 +161,13 @@ object Settings {
     version := createVersion("0.23")
   ) ++ publishSettings
 
-  val openTelemetrySettings = scala213 ++ commonSettings ++ List(
+  val openTelemetrySettings = commonSettings ++ List(
     name := "workbench-openTelemetry",
     libraryDependencies ++= openTelemetryDependencies,
     version := createVersion("0.3")
   ) ++ publishSettings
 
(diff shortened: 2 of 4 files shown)


MatthewBemis:
There is a ticket deep in our backlog to have rawls use `workbench-notification` but I highly doubt it gets done before rawls migrates to 2.13 so 👍  for dropping 2.12 support for `workbench-notification`.

That's the only one I have thoughts about 😄

Qi77Qi:
@jack-r-warren wonder if you added `workbench-openTelemetry` as dependency for services other than Leonardo

project/Settings.scala
```diff
@@ -112,13 +112,17 @@ object Settings {
...
+
+  val only213 = Seq(
+    crossScalaVersions := List("2.13.4")
   )
 
   //common settings for all sbt subprojects
   val commonSettings = commonBuildSettings ++ commonTestSettings ++ List(
     organization := "org.broadinstitute.dsde.workbench",
-    scalaVersion := "2.13.3",
+    scalaVersion := "2.13.4",
```
ehigham: Could you define this version in a val so that this string only appears once?

jack-r-warren:
> @jack-r-warren wonder if you added `workbench-openTelemetry` as dependency for services other than Leonardo

@Qi77Qi Uh no I don't think so. The only other services I've done have been Buffer with [terra-common-lib](https://github.com/DataBiosphere/terra-resource-buffer/pull/200/files) and Rawls with [no new Java dependencies](https://github.com/broadinstitute/rawls/pull/1615/files)

ehigham (approved):
very nice

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-11a45ad"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-TRAVIS-REPLACE-ME"`
```
Qi77Qi, line 75: this isn't really necessary since we don't look to get a new version...but it doesn't hurt to update hash either

PR #465, opened 2021-01-08
IA-2416 IA-2427 Improvements to Dataproc client

rtitle:
Supports Leo removing its legacy Dataproc client (https://github.com/DataBiosphere/leonardo/pull/1812).

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -382,7 +382,7 @@ object GoogleDataprocInterpreter {
     res.getOrElse(Map.empty)
   }
 
-  private def countPreemptibles(instances: Map[DataprocRoleZonePreemptibility, Set[InstanceName]]): Int =
+  private[google2] def countPreemptibles(instances: Map[DataprocRoleZonePreemptibility, Set[InstanceName]]): Int =
     instances.foldLeft(0) { case (r, (DataprocRoleZonePreemptibility(_, _, isPreemptible), instanceNames)) =>
       r + (if (isPreemptible) instanceNames.size else 0)
     }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -1,33 +1,31 @@
 package org.broadinstitute.dsde.workbench.google2
 
-import cats.{Parallel, Show}
 import cats.effect._
 import cats.effect.concurrent.Semaphore
-import cats.syntax.all._
 import cats.mtl.Ask
+import cats.syntax.all._
+import cats.{Parallel, Show}
 import com.google.api.core.ApiFutures
-import com.google.api.gax.rpc.StatusCode.Code
 import com.google.cloud.compute.v1.Operation
 import com.google.cloud.dataproc.v1.{RegionName => _, _}
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.protobuf.FieldMask
 import io.chrisdavenport.log4cats.StructuredLogger
-import org.broadinstitute.dsde.workbench.{DoneCheckable, RetryConfig, StreamTimeoutError}
+import org.broadinstitute.dsde.workbench.DoneCheckable
 import org.broadinstitute.dsde.workbench.google2.DataprocRole.Master
-import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates._
-import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchException}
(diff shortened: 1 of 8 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -69,23 +73,10 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
-        case Left(e: com.google.api.gax.rpc.ApiException) =>
-          if (e.getStatusCode.getCode == Code.ALREADY_EXISTS)
-            Async[F].pure(CreateClusterResponse.AlreadyExists: CreateClusterResponse)
-          else
-            Async[F].raiseError(e): F[CreateClusterResponse]
-        case Left(e)  => Async[F].raiseError(e): F[CreateClusterResponse]
-        case Right(v) => Async[F].pure(CreateClusterResponse.Success(v): CreateClusterResponse)
-      }
-    } yield result
+    retryF(
```
Qi77Qi: since we've been moving away from retrying explicitly, maybe we should use `withLogging`(or `tracedLogging`) directly here too?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -153,6 +144,67 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Timer:
...
+          resizeCluster(project,
+                        region,
+                        clusterName,
+                        numWorkers = None,
+                        numPreemptibles = Some(n)
+          ) >> streamUntilDoneOrTimeout(
+            getClusterInstances(project, region, clusterName),
+            15,
+            6 seconds,
+            s"Cannot stop the instances of cluster ${project.value}/${clusterName.value} before removing its preemptible instances first."
```
Qi77Qi: log message needs to be adjusted?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
rtitle: Yeah sounds good 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
```diff
@@ -159,3 +147,5 @@ final case class DataprocRoleZonePreemptibility(role: DataprocRole, zone: ZoneNa
 
 final case class ClusterError(code: Int, message: String)
 final case class ClusterErrorDetails(code: Int, message: Option[String])
+
+final case class DataprocOperation(name: OperationName, metadata: ClusterOperationMetadata)
```
Qi77Qi, line 151: `OperationName` is not available on `ClusterOperationMetadata`?

Qi77Qi (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocService.scala
rtitle: Nope 😞

Qi77Qi: do we actually need `OperationName` btw? I know we currently persist it in db when we create cluster, but we don't actually use it right?

rtitle: I think we use it to get the cluster error here: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/monitor/DataprocRuntimeMonitor.scala#L177

which is why I added it to the return type. 

Btw `DataprocRuntimeMonitor` is not currently using `Operation` to poll, however we could after making this change (I didn't do it in my PR though).

Qi77Qi: do you see error in `ClusterOperationMetadata`? I think I see error in other `Operation`...just wonder maybe `ClusterOperationMetadata` has it too

rtitle: I don't see errors in `ClusterOperationMetadata`:

image

And I think this is for async errors anyway.. `ClusterOperationMetadata` is returned as part of the API response. To get the async error you need to poll the `OperationName`.

Qi77Qi: ah, that makes sense 👍 forgot it's async

kyuksel (commented):
(sorry for the late review) LGTM 👍🏼

PR #473, opened 2021-01-11
[no ticket] bump kubernetes client version and publish

Qi77Qi:
kubernetes client has breaking change, but I don't think we expose the underlying api call directly.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
@@ -101,7 +101,7 @@ class KubernetesInterpreter[F[_]: Effect: Timer: ContextShift](
       client <- blockingF(getClient(clusterId, new CoreV1Api(_)))
       call = blockingF(
         F.delay(
-          client.listNamespacedPod(namespace.name.value, "true", null, null, null, null, null, null, null, null)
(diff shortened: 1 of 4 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -101,7 +101,7 @@ class KubernetesInterpreter[F[_]: Effect: Timer: ContextShift](
       client <- blockingF(getClient(clusterId, new CoreV1Api(_)))
       call = blockingF(
         F.delay(
-          client.listNamespacedPod(namespace.name.value, "true", null, null, null, null, null, null, null, null)
+          client.listNamespacedPod(namespace.name.value, "true", null, null, null, null, null, null, null, null, null)
```
rtitle, line 104: Any change to the parameter order? I recall another time we bumped the k8s version some of the `null`s and `"true"`s switched places

kyuksel (approved):
LGTM  👍🏼

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: `true` is still for `pretty`, which I think it was intended for....

rtitle (approved):

PR #474, opened 2021-01-11
[IA-2352] Add google2 billing and add some error handling to getInstance

jdcanas:
The resource validator has been having issues because some Deleted/Errored runtimes have billing disabled in their project. Before this change, the call would error with 403 and it would report the same anomaly every day. Now, workbench-libs will say that runtimes do not exist in a project with billing disabled.

It is worth noting that this behavior is not consistent for dataproc. All gce CRUD operations are disallowed in a project with disabled billing, but get/delete work for dataproc clusters.

For dataproc, we are allowed to issue the get/delete request, which happens every day as the cluster cycles error -> deleting -> error. For this reason, we need the resource validator to check if billling is enabled first.

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingInterpreter.scala
@@ -1,0 +1,42 @@
+package org.broadinstitute.dsde.workbench.google2
+
+import cats.Parallel
+import cats.effect.concurrent.Semaphore
+import cats.effect.{Async, Blocker, ContextShift, Timer}
+import cats.mtl.Ask
+import com.google.cloud.billing.v1.{CloudBillingClient, ProjectBillingInfo}
+import io.chrisdavenport.log4cats.StructuredLogger
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import cats.syntax.all._
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates._
+
+private[google2] class GoogleBillingInterpreter[F[_]: StructuredLogger: Parallel: Timer: ContextShift](
+  billingClient: CloudBillingClient,
+  blocker: Blocker,
+  blockerBound: Semaphore[F]
+)(implicit F: Async[F])
+    extends GoogleBillingService[F] {
+
+  override def getBillingInfo(project: GoogleProject)(implicit
+    ev: Ask[F, TraceId]
+  ): F[Option[ProjectBillingInfo]] =
+    for {
+      info <- tracedLogging(
+        blockerBound.withPermit(
+          blocker.blockOn(
+            recoverF(F.delay(billingClient.getProjectBillingInfo(s"projects/${project.value}")), whenStatusCode(404))
+          )
+        ),
+        s"com.google.cloud.billing.v1.CloudBillingClient.getProjectBillingInfo(${project.value})",
+        showBillingInfo
+      )
+    } yield info
+
+  override def isBillingEnabled(project: GoogleProject)(implicit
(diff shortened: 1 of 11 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingService.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.broadinstitute.dsde.workbench.RetryConfig
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import java.nio.file.Path
+
+import scala.collection.JavaConverters._
+
+trait GoogleBillingService[F[_]] {
+  def getBillingInfo(project: GoogleProject)(implicit
```
jdcanas, line 17: simple interface to start, thoughts on making a migration ticket to fully remove google1 billing from leo?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
```diff
@@ -137,7 +137,10 @@ private[google2] class GoogleComputeInterpreter[F[_]: Parallel: StructuredLogger
             .map(Option(_))
             .handleErrorWith {
               case e: ApiException if e.getStatusCode.getCode.getHttpStatusCode == 404 => F.pure(none[Instance])
-              case e                                                                   => F.raiseError[Option[Instance]](e)
+              case e: com.google.api.gax.rpc.PermissionDeniedException
```
jdcanas, line 140: this is the exception the call receives if billing is disabled. From the perspective of the consumer, a runtime in a project with disabled billing does not exist, so we return none here.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingInterpreter.scala
```diff
@@ -0,0 +1,41 @@
...
+  blockerBound: Semaphore[F],
+  retryConfig: RetryConfig
+)(implicit F: Async[F])
+    extends GoogleBillingService[F] {
+
+  override def getBillingInfo(project: GoogleProject)(implicit
+    ev: Ask[F, TraceId]
+  ): F[ProjectBillingInfo] =
+    for {
+      info <- retryF(
```
Qi77Qi: can you use `tracedLogging` instead of using `retryF`..I think we can remove `retryConfig` given these google libs retries under the hood

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingService.scala
Qi77Qi: a ticket for removing google1 billing from leo would be great!

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBillingInterpreter.scala
```diff
@@ -0,0 +1,39 @@
...
+  billingClient: CloudBillingClient,
+  blocker: Blocker,
+  blockerBound: Semaphore[F],
+  retryConfig: RetryConfig
+)(implicit F: Async[F])
+    extends GoogleBillingService[F] {
+
+  override def getBillingInfo(project: GoogleProject)(implicit
+    ev: Ask[F, TraceId]
+  ): F[ProjectBillingInfo] =
```
Qi77Qi: should this return `F[Option[ProjectBillingInfo]]`, and return `None` when 404 exception happens

Qi77Qi (commented):
one comment but otherwise looks great 👍

Qi77Qi (approved):
:shipit:

PR #476, opened 2021-01-19
[no ticket] add StreamTimeoutError.result

rtitle:
Sometimes it's useful to have the last result if a `StreamTimeoutError` occurs.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
@@ -171,7 +171,7 @@ package object google2 {
     streamFUntilDone(fa, maxAttempts, delay).last
       .evalMap {
         case Some(a) if a.isDone => Sync[F].pure(a)
-        case _                   => Sync[F].raiseError[A](StreamTimeoutError(timeoutErrorMessage))
+        case res                 => Sync[F].raiseError[A](StreamTimeoutError(timeoutErrorMessage, res))
       }
       .compile
       .lastOrError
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):
lgtm 👍

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -187,7 +187,7 @@ package object google2 {
     Show.show[Option[ProjectBillingInfo]](info => s"isBillingEnabled: ${info.map(_.getBillingEnabled)}")
 }
 
-final case class StreamTimeoutError(override val getMessage: String) extends WorkbenchException
+final case class StreamTimeoutError[A](override val getMessage: String, result: Option[A]) extends WorkbenchException
```
rtitle, line 190: Not sure I like this actually because the `A` will be erased..

kyuksel (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
kyuksel: Hmmm...are we not guaranteed in this case that `res` is an `Option[A]` 🤔

rtitle: Well since this is the exception type it just gets erased to `Throwable` by things like `F.attempt`. So the `A` here doesn't really propagate to the call site and would just need to be casted from `Any` (which could lead to runtime `ClassCastExceptions`).

Maybe better would be to have `streamUntilDoneOrTimeout` return some kind of ADT:
```
sealed trait StreamResult[A](result: A)
object StreamResult {
  case class Success[A](result: A) extends StreamResult[A]
  case class Timeout[A](result: A) extends StreamResult[A]
}
```
and:
```
def streamUntilDoneOrTimeout[F[_], A](…): F[StreamResult[A]] = // etc
```

But 🤷 , unless anyone has better ideas :)

(It's not that critical, I can achieve what I want with existing APIs too)

kyuksel: What was the use case?

gdlex4015:
As per code review policy all work must have JIRA ticket associated with it. Since JIRA is a source of truth for compliance fields such as CLIA impact and Security Impact Description
With no tickets we are bypassing those checks. This commit did not show up in a monolith release dashboard

rtitle:
@gdlex4015 this PR was closed without merging, I decided not to do it.

PR #490, opened 2021-01-27
[IA-2441] Add a service too replace google1 billing functionality in leonardo 

jdcanas:
This is the PR to use it in leonardo: https://github.com/DataBiosphere/leonardo/pull/1852

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
@@ -55,4 +55,14 @@ private[google2] class GoogleResourceInterpreter[F[_]: StructuredLogger: Paralle
         showProjects
       )
     } yield projects
+
+  override def getLabels(project: GoogleProject)(implicit ev: Ask[F, TraceId]): F[Option[Map[String, String]]] =
+    for {
+      project <- getProject(project)
+    } yield project.map(_.getLabels.asScala.toMap)
+
+  override def getProjectNumber(project: GoogleProject)(implicit ev: Ask[F, TraceId]): F[Option[Long]] =
+    for {
+      project <- getProject(project)
+    } yield project.map(_.getProjectNumber)
 }
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
@@ -35,27 +35,6 @@ private[google2] class GoogleResourceInterpreter[F[_]: StructuredLogger: Paralle
       )
     } yield project
 
-  override def listProject(filter: Option[GoogleProject])(implicit ev: Ask[F, TraceId]): F[List[Project]] =
-    for {
-      projects <- tracedLogging(
-        blockerBound.withPermit(
-          blocker.blockOn(
-            F.delay(
-              filter
-                .fold(resourceClient.list())(project =>
-                  resourceClient.list(ResourceManager.ProjectListOption.filter(s"name:${project.value}"))
-                )
-                .getValues()
-                .asScala
-                .toList
-            )
-          )
-        ),
-        s"com.google.cloud.resourcemanager.ResourceManager.list(${filter.getOrElse("")})",
-        showProjects
-      )
-    } yield projects
-
   override def getLabels(project: GoogleProject)(implicit ev: Ask[F, TraceId]): F[Option[Map[String, String]]] =
     for {
       project <- getProject(project)
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
@@ -1,0 +1,58 @@
+package org.broadinstitute.dsde.workbench.google2
+
+import cats.Parallel
+import cats.effect.concurrent.Semaphore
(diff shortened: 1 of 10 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
```diff
@@ -0,0 +1,68 @@
...
+    for {
+      projects <- tracedLogging(
+        blockerBound.withPermit(
+          blocker.blockOn(
+            F.delay(
+              filter
+                .fold(resourceClient.list())(project =>
+                  resourceClient.list(ResourceManager.ProjectListOption.filter(s"name:${project.value}"))
+                )
+                .getValues()
```
Qi77Qi: if this gives back an `Iterator`, let's use make this function return `Stream[F, Project]`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
```diff
@@ -0,0 +1,68 @@
...
+                .asScala
+                .toList
+            )
+          )
+        ),
+        s"com.google.cloud.resourcemanager.ResourceManager.list(${filter.getOrElse("")})",
+        showProjects
```
Qi77Qi: you can make this `Show[Project]` implicit unless you plan to have multiple instances of this

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
```diff
@@ -0,0 +1,68 @@
...
+      )
+    } yield projects
+
+  override def getLabels(project: GoogleProject)(implicit ev: Ask[F, TraceId]): F[Option[Map[String, String]]] =
```
Qi77Qi, line 38: return `F[Map[String, String]`...`Map` can be empty anyway

Qi77Qi (commented):
Looks great in general..left a few comments

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
jdcanas: If we return an empty map on 404 there's no way to differentiate between a project that has no labels and a project that doesn't exist, which Leo cares about at the call site

Qi77Qi: mm I see..so `None` means project doesn't exist...empty map means project exists, but no labels defined?...that makes sense ...can you add some this in the `trait`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
jdcanas: hmm.. I think I'm going to just remove this method after playing around with it more. The filter behaves very weird, and Leo doesn't use it at all (I thought it would at first over get)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleResourceInterpreter.scala
jdcanas: I did, but I don't believe the `tracedLogging` function expects the `Show[A]` to be explicit, and wants it passed in directly, no?

Qi77Qi: oh I thought it was implicit..nvm

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/mock/FakeGoogleResourceService.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import cats.effect.IO
+import cats.mtl.Ask
+import com.google.cloud.resourcemanager.Project
+import org.broadinstitute.dsde.workbench.google2.GoogleResourceService
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+
+class FakeGoogleResourceService extends GoogleResourceService[IO] {
+  override def getProject(project: GoogleProject)(implicit ev: Ask[IO, TraceId]): IO[Option[Project]] = IO(None)
```
Qi77Qi: can you use `IO.pure(None)` instead of `IO(None)`....IO() expects a function, so it's slightly less performant

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -181,6 +182,12 @@ package object google2 {
 
   val showBillingInfo: Show[Option[ProjectBillingInfo]] =
     Show.show[Option[ProjectBillingInfo]](info => s"isBillingEnabled: ${info.map(_.getBillingEnabled)}")
+
+  implicit val showProject: Show[Option[Project]] =
```
Qi77Qi, line 186: I thikn if you just define `Show[Project]]`....you might be able to get `Show[Option[Project]]` and `Show[List[Project]]` for free

Qi77Qi (approved):
one more minor comment

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
jdcanas: oh, the list variant can actually be removed

PR #501, opened 2021-02-11
[IA-2516] Add TraceId to ErrorReport

ehanna4:
Add TraceId to ErrorReport so we can show traceid in a more user friendly way on Leo

No version bump - adding an optional parameter

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for publishing custom metrics using openTelemetry (openCensus and openTracing). 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-error-reporting" % "0.1-92fcd96"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-error-reporting" % "0.1-TRAVIS-REPLACE-ME"`
 
 [Changelog](errorReporting/CHANGELOG.md)
 
--- a/model/CHANGELOG.md
+++ b/model/CHANGELOG.md
@@ -9,6 +9,7 @@ SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.1
 ### Changed
 - added IdentityConcentratorId to WorkbenchUser 
 - Cross build 2.13
+- added TraceId to ErrorReport
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.14-65bba14"`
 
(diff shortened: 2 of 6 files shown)


Qi77Qi (approved):
first wb-libs PR! 🎉

Qi77Qi (changes requested):
https://broadinstitute.slack.com/archives/GB1VAM14J/p1613144892047500?thread_ts=1613071326.027800&cid=GB1VAM14J

README.md
```diff
@@ -96,7 +96,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for publishing custom metrics using openTelemetry (openCensus and openTracing). 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-error-reporting" % "0.1-92fcd96"`
```
Qi77Qi, line 99: need to update this for `model` too :)

Qi77Qi (approved):
:shipit:

model/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ This file documents changes to the `workbench-model` library, including notes on
 
 ## 0.14
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.14-2e155f0"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-model" % "0.14-TRAVIS-REPLACE-ME"`
```
Qi77Qi, line 7: oh actually, can you add a comment, something like "add `TraceId` to `ErrorReport`"

model/src/main/scala/org/broadinstitute/dsde/workbench/model/ErrorReport.scala
```diff
@@ -28,57 +29,60 @@ case class ErrorReportSource(source: String)
...
   def apply(message: String,
             statusCode: Option[StatusCode],
             causes: Seq[ErrorReport],
             stackTrace: Seq[StackTraceElement],
-            exceptionClass: Option[Class[_]]
+            exceptionClass: Option[Class[_]],
+            traceId: Option[TraceId]
   )(implicit source: ErrorReportSource): ErrorReport =
-    ErrorReport(source.source, message, statusCode, causes, stackTrace, exceptionClass)
+    ErrorReport(source.source, message, statusCode, causes, stackTrace, exceptionClass, traceId)
```
kyuksel, line 85: Adding a parameter to a class constructor (`def apply()` in this case) with no default value seems like a breaking change to me as existing callers of the constructor won't be able to keep calling it the same way. Am I missing something?

kyuksel (changes requested):
Congratulations on your first workbench-libs PR, Emily 👏🏼 

Left a comment regarding the nature of the change (i.e. breaking) as opposed to what's implied from the change log (i.e. non-breaking).

kyuksel (approved):
:shipit:

PR #502, opened 2021-02-17
IA-2052 IA-2533 add getPersistentVolumeClaims

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2052
https://broadworkbench.atlassian.net/browse/IA-2533

Tested in fiab...See description in https://github.com/DataBiosphere/leonardo/pull/1866

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
@@ -3,7 +3,6 @@ package org.broadinstitute.dsde.workbench.google2
 import java.io.ByteArrayInputStream
 import java.util.UUID
 import java.util.concurrent.TimeUnit
-
 import cats.effect.concurrent.Semaphore
 import cats.effect.implicits._
 import cats.effect.{Async, Blocker, ContextShift, Effect, Timer}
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -199,6 +199,38 @@ class KubernetesInterpreter[F[_]: Effect: Timer: ContextShift](
...
+                                                     null,
+                                                     null,
+                                                     null,
+                                                     null,
+                                                     null,
+                                                     null
+          )
+        )
+      ).map(Option(_)).handleErrorWith {
+        case e: io.kubernetes.client.openapi.ApiException if e.getCode == 404 => F.pure(None)
```
jdcanas, line 224: I believe our existing RetryPredicates should handle this logic? (i.e. `recoverF` paired with `whenStatusCode`)

https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala#L36

jdcanas (approved):
minor re-use opportunity, otherwise looks good

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: `recoverF` doesn't save much code though..also it's less obvious using this function that it returns `None` when it's 404..

I can use `whenStatusCode` in `handleErrorWith`. But it's still a bit of a detour in order achieve this logic cuz it translates `Throwable` to a `Boolean` first, and then translating `Boolean` to `Option`....I think this is becuz `whenStatusCode` is originally implemented for retry, but we're not retrying in this case. We could potentially create a separate helper function specifically for translating `Throwable` to `Option` directly

PR #528, opened 2021-02-23
AS-593: Add BigQuery dataset support

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -9,11 +9,13 @@ Changed:
 - Removed `RetryConfig` from `GoogleDataprocService` constructors
 - `GoogleComputeInterpreter` now returns none if it encounters a disabled billing project during `getInstance`
 - Added `GoogleBillingInterpreter` and `GoogleBillingService`
+- Added `createDataset` and `setDatasetIam` to `GoogleBigQueryService`
+- Added new constructor to `GoogleBigQueryService` that accepts path to credentials JSON
 
 Added:
 - Added `GoogleDataprocService.startCluster`
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.19-1aba7fd"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.19-TRAVIS-REPLACE-ME"`
 
 ## 0.18
 Added:
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -8,14 +8,13 @@ Changed:
 - Changed return type of `GoogleDataprocService.{createCluster, deleteCluster, resizeCluster}`
 - Removed `RetryConfig` from `GoogleDataprocService` constructors
 - `GoogleComputeInterpreter` now returns none if it encounters a disabled billing project during `getInstance`
-- Added `GoogleBillingInterpreter` and `GoogleBillingService`
-- Added `createDataset` and `setDatasetIam` to `GoogleBigQueryService`
-- Added new constructor to `GoogleBigQueryService` that accepts path to credentials JSON
 
 Added:
 - Added `GoogleDataprocService.startCluster`
 - Added `getPersistentVolumeClaims` to `KubernetesService`
 - Added `GoogleBillingInterpreter` and `GoogleBillingService`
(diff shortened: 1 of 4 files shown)


google2/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Changed:
 - Changed return type of `GoogleDataprocService.{createCluster, deleteCluster, resizeCluster}`
 - Removed `RetryConfig` from `GoogleDataprocService` constructors
 - `GoogleComputeInterpreter` now returns none if it encounters a disabled billing project during `getInstance`
+- Added `GoogleBillingInterpreter` and `GoogleBillingService`
+- Added `createDataset` and `setDatasetIam` to `GoogleBigQueryService`
+- Added new constructor to `GoogleBigQueryService` that accepts path to credentials JSON
```
davidangb: is "- Added `GoogleBillingInterpreter` and `GoogleBillingService`" a cut-and-paste error? Did you actually change anything there?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
```diff
@@ -37,5 +41,41 @@ private[google2] class GoogleBigQueryInterpreter[F[_]: Sync: ContextShift: Timer
...
+    }
+
+    withLogging(
+      blockingF(Sync[F].delay[DatasetId] {
+        client.update(dataset.toBuilder.setAcl(newAclList.toList.asJava).build()).getDatasetId
+      }),
+      None,
+      s"com.google.cloud.bigquery.BigQuery.update(${datasetName})"
+    )
+  }
```
davidangb, line 68: is this authoritative? Does it overwrite/replace existing ACLs, or does it append?

google2/CHANGELOG.md
MatthewBemis: I honestly have no clue how that got in there

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
MatthewBemis: Unclear at the moment, this definitely requires further testing to avoid bugs. I read something that made me believe it would overwrite but we need to be sure

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleBigQueryInterpreter.scala
```diff
@@ -37,5 +41,41 @@ private[google2] class GoogleBigQueryInterpreter[F[_]: Sync: ContextShift: Timer
...
+  override def createDataset(datasetName: String, labels: Map[String, String]): F[DatasetId] = {
+    val datasetInfo = DatasetInfo.newBuilder(datasetName).setLabels(labels.asJava).build
+
+    withLogging(
+      blockingF(Sync[F].delay[DatasetId] {
+        client.create(datasetInfo).getDatasetId
+      }),
+      None,
+      s"com.google.cloud.bigquery.BigQuery.create(${datasetName})"
+    )
```
ansingh7115: what is the reasoning behind setting creating the dataset and setting the acls in two different calls? Couldn't we set the acls at the same time as we create the dataset?

MatthewBemis: You're totally right, we're going to make this code simplification tomorrow. It should also make handling default IAM much simpler too which is a nice bonus

google2/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Added:
 - Added `GoogleDataprocService.startCluster`
 - Added `listPersistentVolumeClaims` to `KubernetesService`
 - Added `GoogleBillingInterpreter` and `GoogleBillingService`
+- Added `createDataset` and `setDatasetIam` to `GoogleBigQueryService`
```
davidangb: remove `setDatasetIam` - that's no longer a thing, right?

davidangb (approved):
one change to the changelog, otherwise :+1:. I don't need to re-review if all you change is the changelog.

ansingh7115 (approved):
Pending update to changelog

PR #530, opened 2021-03-03
structured logging for google requests

dvoet:
Purpose of this PR is to support structured logging for google requests from the google[1] library.

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -240,11 +240,11 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
       }
 
     val googleRequest = GoogleRequest(request.getRequestMethod,
-      request.buildHttpRequestUrl().toString,
-      payload,
-      System.currentTimeMillis() - startTime,
-      statusCode,
-      errorReport
+                                      request.buildHttpRequestUrl().toString,
+                                      payload,
+                                      System.currentTimeMillis() - startTime,
+                                      statusCode,
+                                      errorReport
     )
     logGoogleRequest(googleRequest)
   }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -2,17 +2,17 @@ package org.broadinstitute.dsde.workbench.google
 
 import java.io.{ByteArrayOutputStream, IOException, InputStream}
 import java.util.concurrent.TimeUnit
-
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.client.http.{HttpResponseException => GoogleHttpResponseException}
 import com.google.api.client.http.{HttpResponse => GoogleHttpResponse}
 import com.google.api.client.http.json.JsonHttpContent
 import com.typesafe.scalalogging.LazyLogging
+import net.logstash.logback.argument.StructuredArguments
 import org.broadinstitute.dsde.workbench.metrics.GoogleInstrumented.GoogleCounters
 import org.broadinstitute.dsde.workbench.metrics.{GoogleInstrumented, Histogram, InstrumentedRetry}
 import org.broadinstitute.dsde.workbench.model.ErrorReport
-import spray.json.JsValue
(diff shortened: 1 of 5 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -240,14 +239,24 @@ trait GoogleUtilities extends LazyLogging with InstrumentedRetry with GoogleInst
...
     logger.debug(
-      GoogleRequest(request.getRequestMethod,
-                    request.buildHttpRequestUrl().toString,
-                    payload,
-                    System.currentTimeMillis() - startTime,
-                    statusCode,
-                    errorReport
-      ).toJson(GoogleRequestFormat).compactPrint
+      googleRequestJson,
+      StructuredArguments.raw("googleRequest", googleRequestJson)
```
dvoet, line 259: this line is the magic

project/Dependencies.scala
```diff
@@ -12,6 +12,7 @@ object Dependencies {
 
   def excludeGuavaJDK5(m: ModuleID): ModuleID = m.exclude("com.google.guava", "guava-jdk5")
 
+  val logstashLogback: ModuleID = "net.logstash.logback"      % "logstash-logback-encoder" % "6.6"
```
Qi77Qi: can you add `provided`

Qi77Qi (approved):
would be nice if you should show an example output

marctalbott (approved):
Seems good, seconding Qi's comment that it'd be nice to see an example

dvoet:
here is an example when `LogstashEncoder` is used
```
{
  "@timestamp": "2021-03-03T22:12:08.348Z",
  "@version": "1",
  "message": "{\"method\":\"GET\",\"statusCode\":200,\"time_ms\":245,\"url\":\"https://www.googleapis.com/admin/directory/v1/groups/GROUP_All_Users@dev.test.firecloud.org\"}",
  "logger_name": "org.broadinstitute.dsde.workbench.google.HttpGoogleDirectoryDAO",
  "thread_name": "scala-execution-context-global-205",
  "severity": "DEBUG",
  "level_value": 10000,
  "googleRequest": {
    "method": "GET",
    "statusCode": 200,
    "time_ms": 245,
    "url": "https://www.googleapis.com/admin/directory/v1/groups/GROUP_All_Users@dev.test.firecloud.org"
  }
}
```

The json is included as escaped json in the message but also embedded in the log json (see the `googleRequest` field). This is for backwards compatibility so when the `LogstashEncoder` is not in use the log message remains the same as it does today.

Here is another with a payload
```
{
  "@timestamp": "2021-03-03T22:09:05.676Z",
  "@version": "1",
  "message": "{\"method\":\"POST\",\"payload\":{\"email\":\"sam-dev-service-account@broad-dsde-dev.iam.gserviceaccount.com\",\"role\":\"MEMBER\"},\"statusCode\":409,\"time_ms\":240,\"url\":\"https://www.googleapis.com/admin/directory/v1/groups/PROXY_109949113883754608360@dev.test.firecloud.org/members\"}",
  "logger_name": "org.broadinstitute.dsde.workbench.google.HttpGoogleDirectoryDAO",
  "thread_name": "scala-execution-context-global-118",
  "severity": "DEBUG",
  "level_value": 10000,
  "googleRequest": {
    "method": "POST",
    "payload": {
      "email": "sam-dev-service-account@broad-dsde-dev.iam.gserviceaccount.com",
      "role": "MEMBER"
    },
    "statusCode": 409,
    "time_ms": 240,
    "url": "https://www.googleapis.com/admin/directory/v1/groups/PROXY_109949113883754608360@dev.test.firecloud.org/members"
  }
```

project/Dependencies.scala
dvoet: by adding `provided` it means we force the client to add this into their dependencies. Do we consider that a backwards incompatible change? I don't have strong feelings either way but I would like to avoid a backwards incompatible change.

dvoet: I just added a note in the changelog

Qi77Qi:
just to double check, this is with `net.logstash.logback.encoder.LogstashEncoder` used in logback as well? something like [this](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/logback.xml#L3) (note this change will help setting log level properly in cloud logging too)...mostly I'm wondering with the logback change, if that's enough to achieve this

project/Dependencies.scala
Qi77Qi: I'm assuming these logging libraries are relatively stable..but yea, good point

dvoet:
Correct, just that logback config change is all that is need to make this work. The examples I posted are from a local running sam with those changes.

PR #555, opened 2021-03-22
[IA-2526] add wording change

jdcanas:
There were 2 options to 'fixing' encountering a cluster not  found issue during app creation. I went with option 2. This does not address the develop use-case of cron job for cleaning up clusters not updating fiab db state.
1. We can check before doing anything in back-leo that the cluster still exists in google, and create it if it doesn’t. This has the con that it introduces logic to ‘fix’ state to only support an edge case that invovles polling that should only occur in test envs, but could be useful for later. It also could have unintended side effects later
2. Make the error message better. This would be unrecoverable if a user encounters it, but to my knowledge this doesnt occur in any error modes currently for a user. I'd be more interested to know if this occurs at all, and fix the state anomaly than to try to address it.

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
@@ -31,7 +31,7 @@ import ca.mrvisser.sealerate
 /** Common Kubernetes models */
 final case class KubernetesClusterNotFoundException(message: String) extends WorkbenchException {
   override def getMessage: String =
-    message + " If you are a user, please contact support. It is possible your cluster was cleaned up."
+    s"${message}. If you are a user, please contact support. It is possible your cluster was cleaned up."
 }
 
 final case class KubernetesInvalidNameException(message: String) extends WorkbenchException {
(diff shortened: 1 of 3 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -30,7 +30,8 @@ import ca.mrvisser.sealerate
 
 /** Common Kubernetes models */
 final case class KubernetesClusterNotFoundException(message: String) extends WorkbenchException {
-  override def getMessage: String = message
+  override def getMessage: String =
+    message + " If you are a user, please contact support. It is possible your cluster was cleaned up."
```
kyuksel: Nit
```suggestion
    s"${message}. If you are a user, please contact support. It is possible your cluster was cleaned up."
```

kyuksel: I also wonder if the piece you added should come before the exception message since the exception message on the ticket is quite long and a user may not read all the way to the most relevant portion.

jdcanas: Due to our 'stack' of errorhandling, its not possible to make this appear at the beginning since its packaged in a general error as it bubbles up from wb-libs from many different 'leaf nodes' within the leo code.  (aka, If I switched the ordering here, it would appear somewhere  in the middle)

I thought the end of the message was a good compromise instead of rewiring our exception handling to be aware of this particular case and repackaging the error entirely.

kyuksel: I see, okay.

kyuksel (approved):
👍🏼

PR #570, opened 2021-03-31
IA-2624 Make dataproc client support multi region

Qi77Qi:
Tested `getCluster` in console.

We have to have different clients for different regions, see the following error
```
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Region 'us-east1' specified in request does not match endpoint region 'us-central1'. To use 'us-east1' region, specify 'us-east1' region in request and configure client to use 'us-east1-dataproc.googleapis.com:443' endpoint.
```

Tried published version in https://github.com/DataBiosphere/leonardo/pull/1933...automation tests passed other than 2 new subnet specific related tests

**PR checklist**

In all cases:

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.20
+Breaking Changes:
+- Make `GoogleDataprocService` support multiple regions
+
+Dependency Updates:
+- Update cats-effect from 2.3.3 to 2.4.0 (#569) (82 seconds ago) <Scala Steward>
+- Update google-cloud-errorreporting from 0.120.34-beta to 0.120.36-beta (#561) (2 hours ago) <Scala Steward>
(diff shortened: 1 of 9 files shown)


wnojopra (approved):
Overall lgtm (but do keep in mind I am a Scala newbie so you may also want Rob's review). 

Quick question - is Leonardo the only user of workbench-libs? Do we need to inform other clients that they need to pass in a Set of RegionNames instead of just a RegionName?

google2/CHANGELOG.md
```diff
@@ -2,13 +2,29 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.20
```
rtitle, line 5: I think there's a `README.md` file to update too?

rtitle (approved):
LGTM.

To @wnojopra's question: Leo is not the only user of wb-libs, but is the only user of the Dataproc client. But even so, this PR includes a version bump, so clients can choose when to upgrade the library version without having breaking changes forced on them.

PR #573, opened 2021-04-02
IA-2623 update createInstance

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2623
I bumped version just yesterday and GoogleComputeService is probably only used by leo team..So figured probably not super necessary to bump again.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
@@ -36,13 +36,9 @@ private[google2] class GoogleComputeInterpreter[F[_]: Parallel: StructuredLogger
   ): F[Option[Operation]] = {
     val projectZone = ProjectZoneName.of(project.value, zone.value)
     retryF(
-      F.delay(instanceClient.insertInstance(projectZone, instance)),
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleComputeInterpreter.scala
```diff
@@ -33,12 +33,16 @@ private[google2] class GoogleComputeInterpreter[F[_]: Parallel: StructuredLogger
 
   override def createInstance(project: GoogleProject, zone: ZoneName, instance: Instance)(implicit
     ev: Ask[F, TraceId]
-  ): F[Operation] = {
+  ): F[Option[Operation]] = {
     val projectZone = ProjectZoneName.of(project.value, zone.value)
     retryF(
```
rtitle, line 38: How about
```
retryF(
  recoverF(F.delay(instanceClient.insertInstance(projectZone, instance)), whenStatusCode(409))
  s"com.google.cloud.compute.v1.InstanceClient.insertInstance(${projectZone.toString}, ${instance.getName})"
)
```

I think `recoverF` is designed for this purpose, and that way the recovery is inside the retry.

Qi77Qi: updated

rtitle (approved):
👍 

Assume Leo side has to change as well to use this

PR #576, opened 2021-04-08
[IA-2528] rename retry functions from being google specific

jdcanas:
Only real changes are in the two files [here](https://github.com/broadinstitute/workbench-libs/pull/576/files#diff-53488ce64518c6d884041bff7b8608b1f3a1cc3f1ed92c98ae5f9b0bb1f92de8R16) and [here](https://github.com/broadinstitute/workbench-libs/pull/576/files#diff-bce2d05ebb5605c6c1d03f730b7f8292dd63516cb99583f1d6923fdc4ab35db3R37), rest is cascading renames.

### DIFF
--- a/google2/CHANGELOG.md
+++ b/google2/CHANGELOG.md
@@ -2,10 +2,17 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.21
+Breaking Changes:
+- Rename `retryGoogleF` and `tracedRetryGoogleF` to `retryF` and `tracedRetryF`
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.21-TRAVIS-REPLACE-ME"`
+
 ## 0.20
 Breaking Changes:
 - Make `GoogleDataprocService` support multiple regions
 - Make `GoogleComputeService.createInstance` and `GoogleDataprocService.createCluster` return `F[Option[Operation]]`
+- Rename `retryGoogleF` and `tracedRetryGoogleF` to `retryF` and `tracedRetryF`
 
 Dependency Updates (latest):
 - Update google-cloud-nio from 0.122.5 to 0.122.11 (#563) (2 hours ago) <Scala Steward>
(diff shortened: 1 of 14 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala
```diff
@@ -13,17 +13,19 @@ object RetryPredicates {
     org.broadinstitute.dsde.workbench.util2.addJitter(1 seconds, 1 seconds),
     x => x * 2,
     5,
-    standardRetryPredicate
+    _ => true
```
Qi77Qi, line 16: is `standardRetryConfig` still used anywhere? this is basically saying retrying all exceptions, which I feel probably shouldn't be used

jdcanas: We are only using it for helm install, because our go helm client doesn't return friendly java exceptions. See comment here https://github.com/DataBiosphere/leonardo/pull/2029/files#diff-bb0f62cc5d31e1e657707a10806f19b9d07c5969f4188cf66e2fb919509a1163R410

Qi77Qi: how about name this `retryAllPredicate` or something? cuz I don't think it should be "standard"

Qi77Qi: can we remove `standard` in the naming :)

google2/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This file documents changes to the `workbench-google2` library, including notes
 Breaking Changes:
 - Make `GoogleDataprocService` support multiple regions
 - Make `GoogleComputeService.createInstance` and `GoogleDataprocService.createCluster` return `F[Option[Operation]]`
+- Rename `retryGoogleF` and `tracedRetryGoogleF` to `retryF` and `tracedRetryF`
```
rtitle, line 15: If this is a new breaking change, shouldn't we bump to 0.21?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/util/RetryPredicates.scala
rtitle: +1, we should use `retryAllConfig` with care, so shouldn't have `standard` in the name imo.

jdcanas: Ah sorry, thought I pushed the change, juggling too many things

google2/CHANGELOG.md
jdcanas: I'm always scared of updating versions because the CI never seems to work for me to generate :(

rtitle (approved):

Qi77Qi (approved):

PR #698, opened 2021-07-24
IA-2881 IA-2878 Upgrade to cats-effect 3

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2878
https://broadworkbench.atlassian.net/browse/IA-2881

* See [cats-effect migration guide](https://typelevel.org/cats-effect/docs/migration-guide#run-the-scalafix-migration)
* google-cloud-compute doesn't have a migration guide. But it seems like there're lots of singleton to plural naming refactoring. They also seem to try really hard not to return `null`, but instead using some arbitrary default, for instance, some default error instance for `error` when `Operation` has no error, and return empty string for `zone` when an Operation isn't zonal. It seems to me a bad idea and led to lots of subtle unit test failures, but those are the changes I've noticed so far.

**PR checklist**

In all cases:


project/Dependencies.scala
```diff
@@ -60,14 +60,14 @@ object Dependencies {
...
-  val googlePubsubNew: ModuleID = "com.google.cloud" % "google-cloud-pubsub" % "1.113.4"
+  val googleStorageNew: ModuleID = "com.google.cloud" % "google-cloud-storage" % "1.117.1"
+  val googleStorageLocal: ModuleID = "com.google.cloud" % "google-cloud-nio" % "0.123.3" % "test"
+  val googlePubsubNew: ModuleID = "com.google.cloud" % "google-cloud-pubsub" % "1.113.5"
   val googleKms: ModuleID = "com.google.cloud" % "google-cloud-kms" % "1.43.0"
   val googleDataproc: ModuleID =    "com.google.cloud" % "google-cloud-dataproc" % "1.5.2"
-  val googleComputeNew: ModuleID = "com.google.cloud" % "google-cloud-compute" % "0.119.11-alpha"
-  val googleContainer: ModuleID = "com.google.cloud" % "google-cloud-container" % "1.4.0"
-  val kubernetesClient: ModuleID = "io.kubernetes" % "client-java" % "11.0.1"
+  val googleComputeNew: ModuleID = "com.google.cloud" % "google-cloud-compute" % "1.3.0-alpha"
```
kyuksel: It's strange to me that all `google-cloud-compute` versions on Maven has the suffix `alpha` on them 🤷🏼‍♂️

kyuksel (approved):
Thanks for taking care of this major upgrade 👍🏼

project/Dependencies.scala
Qi77Qi: yea...they've been on `alpha` forever 🤷‍♀️

PR #728, opened 2021-08-25
[CA-1318] Add list roles to IAM Dao

s-rubenstein:
This should be safe to have no version bump

**PR checklist**
- [ ] For each library you've modified here, decide whether it requires a major, minor, or no version bump. (Click [here](/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md) for guidance)

If you're doing a **major** or **minor** version bump to any libraries:
- [ ] Bump the version in `project/Settings.scala` `createVersion()`
- [ ] Update `CHANGELOG.md` for those libraries
- [ ] I promise I used `@deprecated` instead of deleting code where possible

In all cases:
- [ ] Replace the appropriate version hashes in `README.md` and the `CHANGELOG.md` for any libs you changed with `TRAVIS-REPLACE-ME` to auto-update the version string
- [ ] Get two thumbsworth of PR review
- [ ] Verify all tests go green (CI _and_ coverage tests)
- [ ] Squash commits and **merge to develop**
- [ ] Delete branch after merge

PR #729, opened 2021-08-26
[CA-1318] Add list iam role method to googleIamDAO

s-rubenstein:
This should be safe to have no version bump

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
@@ -1,6 +1,9 @@
 package org.broadinstitute.dsde.workbench.google
 
 import ca.mrvisser.sealerate
+import com.google.api.services.cloudresourcemanager.model.{
+  Policy => ProjectPolicy
+}
 import org.broadinstitute.dsde.workbench.google.GoogleIamDAO.MemberType
 import org.broadinstitute.dsde.workbench.model._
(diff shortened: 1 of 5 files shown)


andy7i (approved):
reviewed this PR in its previous form https://github.com/broadinstitute/workbench-libs/pull/728

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -145,6 +148,13 @@ trait GoogleIamDAO {
                      rolesToRemove: Set[String]
   ): Future[Boolean]
 
+  /**
+   * Lists project-level IAM roles for the given project
+   * @param iamProject the google project in which to list the roles
```
ehigham: For which, or simply "to list the roles for"

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -145,6 +148,13 @@ trait GoogleIamDAO {
...
+   * @return the policy of the project, which lists all roles
+   */
+  def listIamRoles(iamProject: GoogleProject): Future[ProjectPolicy]
```
ehigham: Also feels a bit odd to call an operation "list" when it doesn't return a list (or some monadic transformation thereof). Maybe call it `getProjectPolicy`?

ehigham (approved):

PR #746, opened 2021-09-09
[no ticket] a few more dependency upgrades

Qi77Qi:
* Most the changes are scalafmt change.
* Bump a few dependencies in `Dependencies.scala`
* Replaced cache implemention in `KubernetesService` to use scalacache to void `unsafeRunSync`
* Make publisher, subscriber, and compute constructors to take `numOfThreads` as a param (defaulted to 20) so that we can specify how many threads we want each for each thread pool...From thread dumps I see, each google client initialize a threadpool with quite large number of threads (before I made the changes, I see almost 600 GAX threads..all of which are just in waiting state, and not doing anything from all the thread dumps I've taken). After the change, I'm seeing 300+ GAX threads, and 20 threads for publisher, subscriber and compute respectively. This was good enough to make leo automation tests stable enough. Previous, auto tests will fail 1 or 2 out of each 3 runs (multi-test triggers 3 runs each time)

**PR checklist**

In all cases:


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
```diff
@@ -72,40 +87,51 @@ trait GoogleIamDAO {
...
+   * Adds project-level IAM roles for the given user. This method will perform a read-modify-write of the project's IAM
+   * policy, and return a Boolean indicating whether a change was actually made.
+   * @param iamProject
+   *   the project in which to add the roles
+   * @param email
+   *   the user email address
+   * @param rolesToAdd
+   *   Set of roles to add (example: roles/storage.admin)
+   * @return
+   *   true if the policy was updated; false otherwise.
```
kyuksel: Hmmm…the new format seems to take up a lot of vertical space unnecessarily. If there's a way to revert to the previous setting, I'd vote for doing that.

kyuksel (commented):
I had a difficult time locating the dependency upgrades. If there are any besides those in `.scalafmt.conf`, could you call them out? For the next time, it'd be good to have separate PRs for formatting changes.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleIamDAO.scala
Qi77Qi: I think this is due to the length of the line...new formatter restricts line length more probably

kyuksel: I see - too bad we can't exclude comments from line restrictions

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
```diff
@@ -92,11 +94,11 @@ trait KubernetesService[F[_]] {
 object KubernetesService {
   def resource[F[_]: StructuredLogger: Async](
     pathToCredential: Path,
-    gkeService: GKEService[F]
+    gkeService: GKEService[F],
+    apiClientCache: Cache[F, ApiClient]
```
rtitle, line 98: So the cache is now provided by Leo?

project/Dependencies.scala
```diff
@@ -86,14 +86,15 @@ object Dependencies {
...
-  val fs2Io: ModuleID = "co.fs2" %% "fs2-io" % "3.1.1"
+  val fs2Io: ModuleID = "co.fs2" %% "fs2-io" % "3.1.3"
   val rawlsModel: ModuleID = "org.broadinstitute.dsde" %% "rawls-model" % "0.1-384ab501b" exclude("com.typesafe.scala-logging", "scala-logging_2.13") exclude("com.typesafe.akka", "akka-stream_2.13")
   val openCensusApi: ModuleID = "io.opencensus" % "opencensus-api" % "0.28.3"
   val openCensusImpl: ModuleID = "io.opencensus" % "opencensus-impl" % "0.28.3"
   val openCensusStatsStackDriver: ModuleID = "io.opencensus" % "opencensus-exporter-stats-stackdriver" % "0.28.3"
   val openCensusTraceStackDriver: ModuleID = "io.opencensus" % "opencensus-exporter-trace-stackdriver" % "0.28.3"
   val openCensusTraceLogging: ModuleID = "io.opencensus" % "opencensus-exporter-trace-logging" % "0.26.0"
   val sealerate: ModuleID = "ca.mrvisser" %% "sealerate" % "0.0.6"
+  val scalaCache = "com.github.cb372" %% "scalacache-caffeine" % "1.0.0-M4"
```
rtitle, line 97: Wonder if we can remove `google-guava` dependency? :) (L58)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
```diff
@@ -23,61 +23,36 @@ trait CleanUp extends TestSuiteMixin with ExceptionHandling with LazyLogging { s
...
-   * <pre>
-   * "tries to clean-up parent before child" in {
-   *   withParent { parent =>  // withParent loan-fixture clean-up will run before registered clean-up functions
-   *     child = Child(parent)
-   *     register cleanUp { delete child }
-   *     ...
-   *   }
-   * }
-   * </pre>
+   * <pre> "tries to clean-up parent before child" in { withParent { parent => // withParent loan-fixture clean-up will
```
rtitle: This formatting is a bit unfortunate… wonder if we can make it exclude comments/javadoc?

util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
```diff
@@ -19,14 +19,12 @@ trait Retry {
...
   /**
-   * A Future that has potentially been retried, with accumulated errors.
-   * There are 3 cases:
-   * 1. The future failed 1 or more times, and the final result is an error.
-   *   - This is represented as {{{Left(NonEmptyList(errors))}}}
-   * 2. The future failed 1 or more times, but eventually succeeded.
-   *   - This is represented as {{{Right(List(errors), A)}}}
-   * 3. The future succeeded the first time.
-   *   - This is represented as {{{Right(List.empty, A)}}}
+   * A Future that has potentially been retried, with accumulated errors. There are 3 cases:
```
rtitle: Another bad formatting change

rtitle:
Looks good overall - the logic/dependency changes all look good. 👍 I mostly skimmed over the formatting changes, but some of the new JavaDoc formatting looks a bit destructive (`<pre>` sections with code examples squished, ordered lists messed up, etc). Wonder if there's a way to disable Javadoc formatting?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesService.scala
Qi77Qi: yep..I also added it to cache metrics stream :)

project/Dependencies.scala
Qi77Qi: will do

Qi77Qi: `workbenchModel` rely on `com.google.common.net.UrlEscapers` that comes from guava, which is a bit questionable if that's the right thing to do, but think I'll leave guava in for now

Qi77Qi:
@rtitle @kyuksel I updated the scalafmt to use `docstring.style=keep`...this seems will skip docs

rtitle (approved):
Looks good! Thanks for fixing the javadoc

Qi77Qi:
thanks @rtitle  for review! @kyuksel if you have more comments, I can address retrospectively

PR #770, opened 2021-09-28
[QA-1571] retry iam update when group "does not exist"

andy7i:
fixes Issue # 2 from https://broadworkbench.atlassian.net/browse/QA-1571?src=confmacro

Example of the issue we're trying to retry on:
During workspace creation, updating the google iam on the project sometimes fails, which seems to be a race condition.
```
[INFO] [11:08:40.133] [scala-execution-context-global-574] o.b.d.w.google.HttpGoogleIamDAO - retry-able operation failed: retries remain but predicate failed, not retrying
com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
POST https://cloudresourcemanager.googleapis.com/v1/projects/terra-quality-9bc8e9d5:setIamPolicy
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "Group policy-c9a8c0f1-bb5a-40a7-b685-c7746aa10254@quality.firecloud.org does not exist.",
    "reason" : "badRequest"
  } ],
  "message" : "Group policy-c9a8c0f1-bb5a-40a7-b685-c7746aa10254@quality.firecloud.org does not exist.",
  "status" : "INVALID_ARGUMENT"
}

```

This PR adds a retry predicate for this scenario. 

We can't depend on google to propagate changes everywhere. Even if we try waiting until we can _get_ the group, it doesn't guarantee that we will be able to update that group in the very next call. So this was changed to take an optional parameter to retry. Instead of retrying from Rawls, we're adding the retry predicate here so that others can also use this. 

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -91,6 +91,16 @@ object GoogleUtilities {
       case _                              => false
     }
 
+    /**
+     * Groups may take some time to sync to Google and propogate. This will retry if the group does not exist (yet).
+     */
+    def when400WithDoesNotExistMessage(throwable: Throwable): Boolean = throwable match {
+      case t: GoogleJsonResponseException =>
+        t.getStatusCode == 400 &&
+          compareErrorMessage(t)(_.contains("does not exist"))
+      case _ => false
+    }
+
     private def compareErrorDomain(ex: GoogleJsonResponseException)(fn: String => Boolean): Boolean = {
       val res = for {
         details <- Option(ex.getDetails)
@@ -108,6 +118,16 @@ object GoogleUtilities {
         headError <- errors.headOption
         reason <- Option(headError.getReason)
       } yield fn(reason)
+      res.getOrElse(false)
+    }
+
+    private def compareErrorMessage(ex: GoogleJsonResponseException)(fn: String => Boolean): Boolean = {
+      val res = for {
+        details <- Option(ex.getDetails)
+        errors <- Option(details.getErrors).map(_.asScala)
+        headError <- errors.headOption
+        message <- Option(headError.getMessage)
+      } yield fn(message)
       res.getOrElse(false)
     }
   }
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
@@ -94,7 +94,7 @@ object GoogleUtilities {
     /**
      * Groups may take some time to sync to Google and propogate. This will retry if the group does not exist (yet).
      */
-    def when400WithDoesNotExistMessage(throwable: Throwable): Boolean = throwable match {
+    def whenGroupDoesNotExist(throwable: Throwable): Boolean = throwable match {
       case t: GoogleJsonResponseException =>
         t.getStatusCode == 400 &&
           compareErrorMessage(t)(_.contains("does not exist"))
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
@@ -215,6 +215,7 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
     // Note the project here is the one in which we're removing the IAM roles
     // Retry 409s here as recommended for concurrent modifications of the IAM policy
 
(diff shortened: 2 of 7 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala
```diff
@@ -110,6 +120,16 @@ object GoogleUtilities {
       } yield fn(reason)
       res.getOrElse(false)
     }
+
+    private def compareErrorMessage(ex: GoogleJsonResponseException)(fn: String => Boolean): Boolean = {
```
andy7i, line 124: follows the pattern of the above private functions

marctalbott (approved):
Is there a corresponding Rawls PR?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -211,13 +211,15 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
...
-          when409
+    retry(
+      when5xx,
+      whenUsageLimited,
+      whenGlobalUsageLimited,
+      when404,
+      whenInvalidValueOnBucketCreation,
+      whenNonHttpIOException,
+      when409,
+      when400WithDoesNotExistMessage
```
gpolumbo-broad: I know we had talked about this before, but I'm not sure how I feel about this approach now.  On one hand, making a request to `modifyIamRoles` or `getProjectPolicy` should legitimately return a `400` if the group/policy does not exist.  

However, as written, if the group/policy legitimately does not exist, this won't just fail and return, and instead it will always retry until it times out.  That's what we want for this `createWorkspace` bug we're trying to address, but that might not be what we want globally.  The reality is, this method looks to actually only be used during `createWorkspace`, at least in Rawls, but this might get used elsewhere in the future (or it might be already, I did not do a comprehensive search through everything that uses `workbench-libs`).

In other words this, "If Google thinks the group doesn't exist, I don't believe them!" logic should be isolated to place where we feel that way.  I think 99% of the time, if Google says the group doesn't exist, we should believe them.  

So maybe the better answer here is to put some retry logic in `createWorkspace`, after we call `syncPolicies`, we actually query Google and say, "I just told you to make this group, does it exist yet?" and we retry that until it succeeds (or times out) and then proceed with the `createWorkspace` flow after Google finally notices that the group exists.

gpolumbo-broad (changes requested):
Sorry, my comment is a major change to overall solution.  If you feel strongly that your implementation is the right way to go or my concerns are actually not so bad, that's fine, just let me know.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
andy7i: great points. I think that does seem like a better approach instead of making changes in a shared library like this.

andy7i: We can't depend on google to propagate changes everywhere. Even if we try waiting until we can _get_ the group, it doesn't guarantee that we will be able to update that group in the very next call. So this was changed to take an optional parameter to retry. Instead of retrying from Rawls, we're adding the retry predicate here so that others can also use this.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
```diff
@@ -192,48 +192,72 @@ class HttpGoogleIamDAO(appName: String, googleCredentialMode: GoogleCredentialMo
...
     // Retry 409s here as recommended for concurrent modifications of the IAM policy
-    retry(when5xx,
-          whenUsageLimited,
-          whenGlobalUsageLimited,
-          when404,
-          whenInvalidValueOnBucketCreation,
-          whenNonHttpIOException,
-          when409
+
+    val when400WithDoesNotExistMessagePredicateList: Seq[(Throwable => Boolean)] = Seq(when400WithDoesNotExistMessage)
```
marctalbott: nit: I'd prefer to skip this seq and append `when400WithDoesNotExistMessage` directly to the `basePredicateList` down in that if block.

marctalbott (approved):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleIamDAO.scala
andy7i: that's what I had originally, but the PR tests do not compile..

gpolumbo-broad (approved):

PR #783, opened 2021-10-12
IA-2994 use default stopCluster call

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2994

* Introduce `isFullStop` flag. When true, we'll call dataproc sdk's `stopClusterAsync` method to stop the fully stop the cluster; when false, we'll do what we've been doing previously, which is simply stopping all instances, but the cluster itself is still Running
* With the ability of "fully" stop a cluster, we need to update `startCluster` to call dataproc's `startClusterAsync` method to start the cluster first if it's not already in Running status; otherwise, we do what we've been doing, which is starting all instances manually (in the case when user previously chose to not fully stop the cluster)

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -117,6 +117,31 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Paralle
           }
       }
 
+      // First, remove preemptible instances (if any) and wait until the removal is done
+      _ <-
+        if (countPreemptibles(clusterInstances) > 0)
+          resizeCluster(project,
+                        region,
+                        clusterName,
+                        numWorkers = None,
+                        numPreemptibles = Some(0)
+          ) >> streamUntilDoneOrTimeout(
+            getClusterInstances(project, region, clusterName),
+            15,
+            6 seconds,
+            s"Timeout occurred removing preemptible instances from cluster ${project.value}/${clusterName.value}"
+          )(implicitly, instances => countPreemptibles(instances) == 0)
+        else F.unit
+
+      // If removal of preemptibles is done, wait until the cluster's status transitions back to RUNNING (from UPDATING)
+      // Otherwise, stopping the remaining instances may cause the cluster to get in to ERROR status
+      _ <- streamUntilDoneOrTimeout(
+        getCluster(project, region, clusterName),
+        15,
+        3 seconds,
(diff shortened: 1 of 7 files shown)


kyuksel (commented):
Scanned quickly. Looks good overall. A few high-level questions below and inline:

1. Is removing preemptibles still a prerequisite to stopping while using the new API? Wanted to double-check since it'd save some complexity if we could remove that logic.
2. Do we need to change `ResourceValidator`'s [checkDataprocCluster()](https://github.com/broadinstitute/leonardo-cron-jobs/blob/4d844ca56a96471800c7ff572796fbad7735c246/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala#L59) since the logic there may not apply after we start _actually stopping_ Dataproc clusters?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -158,59 +197,97 @@ private[google2] class GoogleDataprocInterpreter[F[_]: StructuredLogger: Paralle
...
-            }).recoverWith {
-              case _: com.google.api.gax.rpc.NotFoundException => F.pure(Operation.getDefaultInstance)
-              case e                                           => F.raiseError[Operation](e)
+                ).as(None)
+              }
+              .map(metadata => DataprocOperation(OperationName(javaFuture.getName), metadata))
+          } yield Some(operation)
+        else {
+          // We support non-full stop when we "stop" a dataproc cluster
+          // (when we're patching a dataproc cluster or for existing `Stopped` dataproc clusters).
```
kyuksel, line 255: Thinking out loudly if we still need to support the old version of stopping a cluster: 
1. What about patching that requires us to use the old version?
2. Could we deal with existing clusters that were "stopped" the old way?

Qi77Qi:
@kyuksel 
for 1, yea, it is...I was surprised the api doesn't handle it as well (saw failure in automation tests without it)...
for 2, good point...I'll create a ticket for that...(it'll be nice to have)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
Qi77Qi: for 1, we're supporting the old way..specifically for patching
for 2, yes...we do the old way if cluster is in `Running` status; and call `startCluster` if cluster is `Stopped`

kyuksel (approved):

PR #792, opened 2021-10-26
[CI-1096] Add Google Storage Transfer Service Support to `workbench-google2`

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1096
Add functionality to `workbench-google2` to support transferring Google Cloud Storage objects between GCS buckets via the Google Storage Transfer Service.
Add test to demo creating and monitoring STS jobs.
Bumping minor version to 0.23.
Removing cross compile support for Scala 2.12 for google2 - library has upgraded to Scala 2.13.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.22-82a345c"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
@@ -359,10 +359,10 @@ object StorageRole {
   final case object StorageAdmin extends StorageRole {
     def name: String = "roles/storage.admin"
   }
-  final case object LegacyStorageReader extends StorageRole {
+  final case object LegacyBucketReader extends StorageRole {
     def name: String = "roles/storage.legacyBucketReader"
   }
-  final case object LegacyStorageWriter extends StorageRole {
+  final case object LegacyBucketWriter extends StorageRole {
     def name: String = "roles/storage.legacyBucketWriter"
   }
   //The custom roleId must be in the form "organizations/{organization_id}/roles/{role}",
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
@@ -179,7 +179,7 @@ trait GoogleStorageService[F[_]] {
    */
   def getObjectMetadata(bucketName: GcsBucketName,
                         blobName: GcsBlobName,
-                        traceId: Option[TraceId],
+                        traceId: Option[TraceId] = None,
                         retryConfig: RetryConfig = standardGoogleRetryConfig
   ): Stream[F, GetMetadataResponse]
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
@@ -359,6 +359,12 @@ object StorageRole {
   final case object StorageAdmin extends StorageRole {
     def name: String = "roles/storage.admin"
   }
+  final case object LegacyStorageReader extends StorageRole {
+    def name: String = "roles/storage.legacyBucketReader"
+  }
+  final case object LegacyStorageWriter extends StorageRole {
+    def name: String = "roles/storage.legacyBucketWriter"
+  }
   //The custom roleId must be in the form "organizations/{organization_id}/roles/{role}",
   //or "projects/{project_id}/roles/{role}"
   final case class CustomStorageRole(roleId: String) extends StorageRole {
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
@@ -14,7 +14,7 @@ import org.broadinstitute.dsde.workbench.model.google._
 import scala.jdk.CollectionConverters._
 import scala.util.Using
 
-private[google2] class GoogleStorageTransferInterpreter[F[_]](client: StorageTransferServiceClient)(implicit
+private[google2] final class GoogleStorageTransferInterpreter[F[_]](client: StorageTransferServiceClient)(implicit
   F: Async[F]
 ) extends GoogleStorageTransferService[F] {
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
@@ -1,0 +1,94 @@
+package org.broadinstitute.dsde.workbench
+package google2
+
+import java.nio.channels.Channels
+import java.nio.file.{Path, Paths}
+import java.time.Instant
+import fs2.io.file.Files
+import cats.data.NonEmptyList
+import cats.effect._
+import cats.effect.std.Semaphore
+import cats.syntax.all._
+import com.google.auth.oauth2.ServiceAccountCredentials
+import com.google.cloud.storage.BucketInfo.LifecycleRule
+import com.google.cloud.storage.Storage.{BlobListOption, BlobSourceOption, BlobTargetOption, BlobWriteOption, BucketGetOption, BucketSourceOption}
+import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, Bucket, BucketInfo, Storage, StorageOptions}
+import com.google.cloud.{Identity, Policy, Role}
+import fs2.{Pipe, Stream, text}
+import org.typelevel.log4cats.StructuredLogger
+import io.circe.Decoder
+import io.circe.fs2._
+import org.broadinstitute.dsde.workbench.google2.util.RetryPredicates.standardGoogleRetryConfig
+import org.broadinstitute.dsde.workbench.model.{TraceId, WorkbenchException}
+import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject}
+import com.google.auth.Credentials
+import com.google.storagetransfer.v1.proto.StorageTransferServiceClient
+import com.google.storagetransfer.v1.proto.TransferTypes.{TransferJob, TransferSpec}
+
+import scala.collection.JavaConverters._
+
+private[google2] class GoogleStorageTransferInterpreter[F[_]](
+                                                       db: Storage,
+                                                       blockerBound: Option[Semaphore[F]]
+                                                     )(implicit logger: StructuredLogger[F], F: Async[F])
+  extends GoogleStorageTransferService[F] {
+  override def transferBucket(jobName: String,
+                               projectToBill: GoogleProject
+
+
+                             ): F[Unit] =
+    StorageTransferServiceClient.create().createTransferJob(
+      new TransferJob.Builder()
+        .setName(jobName)
+        .setProjectId(projectToBill.value)
+        .setTransferSpec(
+          new TransferSpec.Builder()
+            .
+            .build()
+
+        )
+
+      .build()
+  )
+}
+
+object GoogleStorageInterpreter {
+  def apply[F[_]: Async: StructuredLogger](
+                                            db: Storage,
+                                            blockerBound: Option[Semaphore[F]]
+                                          ): GoogleStorageInterpreter[F] =
+    new GoogleStorageInterpreter(db, blockerBound)
+
+  def storage[F[_]: Sync: Files](
+                                  pathToJson: String,
+                                  project: Option[GoogleProject] =
+                                  None // legacy credential file doesn't have `project_id` field. Hence we need to pass in explicitly
+                                ): Resource[F, Storage] =
+    for {
+      credential <- org.broadinstitute.dsde.workbench.util2.readFile(pathToJson)
+      project <- project match { //Use explicitly passed in project if it's defined; else use `project_id` in json credential; if neither has project defined, raise error
+        case Some(p) => Resource.pure[F, GoogleProject](p)
+        case None    => Resource.eval(parseProject(pathToJson).compile.lastOrError)
+      }
+      db <- Resource.eval(
+        Sync[F].delay(
+          StorageOptions
+            .newBuilder()
+            .setCredentials(ServiceAccountCredentials.fromStream(credential))
+            .setProjectId(project.value)
+            .build()
+            .getService
+        )
+      )
+    } yield db
+
+  implicit val googleProjectDecoder: Decoder[GoogleProject] = Decoder.forProduct1(
+    "project_id"
+  )(GoogleProject.apply)
+
+  def parseProject[F[_]: Files: Sync](pathToJson: String): Stream[F, GoogleProject] =
+    Files[F]
+      .readAll(Paths.get(pathToJson), 4096)
+      .through(byteStreamParser)
+      .through(decoder[F, GoogleProject])
+}
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
@@ -1,13 +1,11 @@
 package org.broadinstitute.dsde.workbench.google2
 
 import cats.effect._
-import com.google.`type`.{Date, TimeOfDay}
+import com.google.`type`.Date
 import com.google.longrunning.{ListOperationsRequest, Operation}
 import com.google.storagetransfer.v1.proto.TransferProto._
 import com.google.storagetransfer.v1.proto.TransferTypes._
 import com.google.storagetransfer.v1.proto.{StorageTransferServiceClient, TransferProto}
-import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.ObjectDeletionOption._
-import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.ObjectOverwriteOption._
 import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService._
 import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
 import org.broadinstitute.dsde.workbench.model.google._
@@ -28,16 +26,6 @@ final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: Stor
         .build
   }
 
-  private object TimeOfDay {
-    def fromZonedDateTime(datetime: ZonedDateTime): TimeOfDay =
-      com.google.`type`.TimeOfDay.newBuilder
-        .setHours(datetime.getHour)
-        .setMinutes(datetime.getMinute)
-        .setSeconds(datetime.getSecond)
-        .setNanos(datetime.getNano)
-        .build
-  }
-
   private def makeSchedule(schedule: JobTransferSchedule): Schedule = schedule match {
     case JobTransferSchedule.Immediately =>
       val today = Date.fromZonedDateTime(ZonedDateTime.now)
@@ -45,23 +33,6 @@ final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: Stor
         .setScheduleStartDate(today)
         .setScheduleEndDate(today)
         .clearStartTimeOfDay()
-        .build
-
-    case JobTransferSchedule.Once(datetime) =>
-      val date = Date.fromZonedDateTime(datetime)
-      Schedule.newBuilder
-        .setScheduleStartDate(date)
-        .setScheduleEndDate(date)
-        .setStartTimeOfDay(TimeOfDay.fromZonedDateTime(datetime))
-        .build
-  }
-
-  private def makeTransferOptions(options: JobTransferOptions): TransferOptions = options match {
-    case JobTransferOptions(overwrite, delete) =>
-      TransferOptions.newBuilder
-        .setOverwriteObjectsAlreadyExistingInSink(overwrite == OverwriteObjectsAlreadyExistingInSink)
-        .setDeleteObjectsUniqueInSink(delete == DeleteObjectsUniqueInSink)
-        .setDeleteObjectsFromSourceAfterTransfer(delete == DeleteSourceObjectsAfterTransfer)
         .build
   }
 
@@ -85,8 +56,7 @@ final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: Stor
                                  projectToBill: GoogleProject,
                                  originBucket: GcsBucketName,
                                  destinationBucket: GcsBucketName,
-                                 schedule: JobTransferSchedule,
-                                 options: Option[JobTransferOptions]
+                                 schedule: JobTransferSchedule
   ): F[TransferJob] = {
     val transferJob = TransferJob.newBuilder
       .setName(jobName.value)
@@ -97,7 +67,6 @@ final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: Stor
         TransferSpec.newBuilder
           .setGcsDataSource(GcsData.newBuilder.setBucketName(originBucket.value).build)
           .setGcsDataSink(GcsData.newBuilder.setBucketName(destinationBucket.value).build)
-          .setTransferOptions(options map makeTransferOptions getOrElse TransferOptions.getDefaultInstance)
           .build
       )
       .setSchedule(makeSchedule(schedule))
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
@@ -12,8 +12,7 @@ import org.broadinstitute.dsde.workbench.model.google._
 import scala.jdk.CollectionConverters._
 import scala.util.Using
 
-class GoogleStorageTransferInterpreter[F[_]]()(implicit F: Async[F])
-    extends GoogleStorageTransferService[F] {
+class GoogleStorageTransferInterpreter[F[_]]()(implicit F: Async[F]) extends GoogleStorageTransferService[F] {
 
   private def makeJobTransferSchedule(schedule: TransferJobSchedule) = schedule match {
     case TransferOnce(date) =>
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
@@ -43,13 +43,13 @@ class GoogleStorageTransferInterpreter[F[_]]()(implicit logger: StructuredLogger
     }))
   }
 
-  override def transferBucket(jobName: TransferJobName,
-                              jobDescription: String,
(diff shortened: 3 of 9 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService._
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.google._
+
+import scala.jdk.CollectionConverters._
+import scala.util.Using
+
+class GoogleStorageTransferInterpreter[F[_]]()(implicit F: Async[F]) extends GoogleStorageTransferService[F] {
+
+  private def makeJobTransferSchedule(schedule: TransferJobSchedule) = schedule match {
```
Qi77Qi: can you specify return type

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,103 @@
...
+    case TransferOnce(date) =>
+      Schedule.newBuilder
+        .setScheduleStartDate(date)
+        .setScheduleEndDate(date)
+        .build
+  }
+
+  private def makeJobTransferOptions(options: TransferJobOptions) = options match {
```
Qi77Qi: same return type

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
```diff
@@ -0,0 +1,91 @@
...
+  def getTransferOperation(operationName: TransferOperationName): F[Operation]
+
+}
+
+object GoogleStorageTransferService {
+
+  sealed trait TransferJobOverwriteOption
+
+  /** Transfer objects from source if not binary equivalent to those at destination. */
+  object OverwriteObjectsIfDifferent extends TransferJobOverwriteOption
```
Qi77Qi: I usually do 
```
sealed trait TransferJobOverwriteOption extends Product with Serialiazable 
object TransferJobOverwriteOption {
 ....
}
```

and at call site, do `TransferJobOverwriteOption.OverwriteObjectsIfDifferent`..just to make it a bit more obvious what enum we're dealing with

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  /** Delete objects from source after they've been transferred. */
+  object DeleteSourceObjectsAfterTransfer extends TransferJobDeletionOption
+
+  /** Delete files from destination if they're not at source. */
+  object DeleteObjectsUniqueInSink extends TransferJobDeletionOption
+
+  sealed trait TransferJobSchedule
+
+  case class TransferOnce(date: Date) extends TransferJobSchedule
```
Qi77Qi: `final case class TransferOnce(date: Date) extends TransferJobSchedule extends AnyVal`
 by doing [AnyVal](https://www.scala-lang.org/api/2.12.4/scala/AnyVal.html), there will less overhead of allocation (sometimes still will have some overhead)

project/Settings.scala
```diff
@@ -154,10 +154,10 @@ object Settings {
     coverageExcludedPackages := ".*HttpGoogle.*DAO.*"
   ) ++ publishSettings
 
-  val google2Settings = cross212and213 ++ commonSettings ++ List(
+  val google2Settings = commonSettings ++ List(
```
Qi77Qi: mmm...I'm not sure if all services have migrated to 2.13 at this point...maybe ask in #dsp-workbench? (definiltey would be nice to drop this)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,103 @@
...
+        .setDeleteObjectsFromSourceAfterTransfer(delete == DeleteSourceObjectsAfterTransfer)
+        .build
+  }
+
+  override def getStsServiceAccount(project: GoogleProject): F[ServiceAccount] = {
+    val request = GetGoogleServiceAccountRequest.newBuilder
+      .setProjectId(project.value)
+      .build
+
+    F.delay(Using.resource(StorageTransferServiceClient.create) { client =>
```
Qi77Qi: in other clients, what we typically do is have the underlying client being passed to the class as param. And use `Resource` at when we construct `GoogleStorageTransferService`

sth like
```
private class GoogleStorageTransferInterpreter[F[_]](client: StorageTransferServiceClient){
 ...
}
```
(not this class should be private because you don't want user to use it directly)
and at companion object of the trait, you'll have 
```
object GoogleStorageTransferService {
   def resource: F[GoogleStorageTransferService] = for {
       client <- cats.effect.Resource(xxx)(_.close)
       res <- new GoogleStorageTransferInterpreter(client) 
   } yield res
}
```

this way, you don't have to create and close the client for each function call...and instead re-use the same client

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: I can but I don't think it's worth it for these simple private helper functions. The return type is obvious from the implementation; adding type annotations here just adds clutter - if one were to change the implementation and not maintain type correctness the compiler would complain (this IMO is the beauty of type inference).
If you feel strongly about it I'll add a return type but please explain why.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: I don't see the ambiguity or the need for the extra namespace. At the call site it'll read

```scala
options = TransferJobOptions(
    whenToOverwrite = OverwriteObjectsIfDifferent
    whenToDelete = DeleteObjectsUniqueInSink
).some
```

Is the extra `TransferJobOverwriteOption.` really necessary? I'd probably just import this if I was going to use it.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: Interesting - reading into this, is there a reason we don't go one further and use `@newtype`s?
dodgy article: https://blog.rockthejvm.com/value-classes/

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: I'm not really worried about performance at this point. Doing this only moves the problem - someone else then has to close/`use` the resource. Also, should your signature not read:
```scala
def resource: Resource[F, GoogleStorageTransferService[F]]
```
?

I'm not sure what using dependency injection buys us at this moment - there's a `trait` that someone could mock should they wish whose interface is somewhat simpler.

ehigham:
Thank you for your review @Qi77Qi - you've raised some interesting discussion points. Look forward to hearing back!

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
Qi77Qi: there're several newtype libs...no obvious winners..and for most cases, `AnyVal` is good enough

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
Qi77Qi: if someone ever needs to mock, it's the `GoogleStorageTransferService`..the interpreter itself should never to exposed to user.

> I'm not really worried about performance at this point.

Why deviate from the more common pattern in this lib for no gain then?

> someone else then has to close/use the resource

The lib handles the closing...User doen't need to remember closing

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
Qi77Qi: the return type isn't obvious to me by looking at the code.. (I'm not sure why you need to use `match` honestly)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
Qi77Qi: not just call site, I think it's cleaner in the definition site too so that you don't have to look closely which cases extends which sum type..especially you have several of them here....(enum will be a lot better in scala 3 tho)

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: Google libraries (guava etc) love to use builder patterns. 

> (I'm not sure why you need to use match honestly)

There are many ways of scheduling; the sum type should be extended to enumerate them as required.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: > if someone ever needs to mock, it's the GoogleStorageTransferService

That's precisely what I'm saying

> the interpreter itself should never to exposed to user.

OK. I'd argue it's more important to not expose the underlying client though. But by not exposing the interpreter I guess you're doing that anyway.
I'd actually do away with the trait entirely and adopt a more FP solution using records / case classes.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: Scala 3 enums do seem nicer. OK, It'll group them as you suggest.

ehigham: Why `extends Product with Serialiazable`? It's not a product type and all java books I've read (effective java and so on) recommend not using java serialisation citing security vulerabilities?

Qi77Qi: https://stackoverflow.com/a/36526562

ehigham: Ugh scala...

ehigham: In light of that, I don't think I want them to support collections - these options are mutually exclusive.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: I can't seem to get this to work. Seems `AnyVal` doesn't work on sum types?

ehigham: Ah, trait must extend `Any` then case class extends `AnyVal` with trait

project/Settings.scala
Qi77Qi: checked with @andy7i quickly, rawls is still on 2.12...so we can't drop it yet

andy7i: reason for that: 
```
    scalaVersion  := "2.12.15", // `cromwell-client` needs to support 2.13 for rawls to be able to upgrade
```
https://github.com/broadinstitute/rawls/blob/f6c96e945f169d6d6620c7540d23d9a057f0c247/project/Settings.scala#L74

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,111 @@
...
+        .setDeleteObjectsFromSourceAfterTransfer(delete == DeleteSourceObjectsAfterTransfer)
+        .build
+  }
+
+  override def getStsServiceAccount(project: GoogleProject): F[ServiceAccount] = {
+    val request = GetGoogleServiceAccountRequest.newBuilder
+      .setProjectId(project.value)
+      .build
+
+    F.delay {
```
Qi77Qi: I think this function is mostly pure now..probably don't need `F.delay` anymore...unless some of these java calls can throw exception, in which case they should be wrapped in `F.delay`

ehigham: I'm wrapping all external calls via the client with `delay`. These can indeed throw exceptions.

project/Settings.scala
ehigham: I don't think it'll compile in 2.12 :'(

ehigham: fixed

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
Qi77Qi: oh I didn't realize there's `client.getGoogleServiceAccount(request)` there....so I would do this instead

`F.delay(client.getGoogleServiceAccount(request)).map(sa => ServiceAccount(
        ServiceAccountSubjectId(sa.getSubjectId),
        WorkbenchEmail(sa.getAccountEmail),
        ServiceAccountDisplayName(sa.getAccountEmail)
      ))`

The idea is you want to wrap just the effect call in F.delay...instead of having your entire program in a giant F.delay, which defeats the purpose.

I would also put the that google call inside [this](https://github.com/broadinstitute/workbench-libs/blob/6140478c9cbf80ddb1d0fd65e0581b94660bf0ca/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala#L54), which will log the call and the response

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.22-f394f70"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-TRAVIS-REPLACE-ME"`
```
andy7i: based on this doc (https://github.com/broadinstitute/workbench-libs/blob/develop/CONTRIBUTING.md#no-version-bump-necessary), I think it doesn't need a version bump.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,110 @@
...
+import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService.ObjectOverwriteOption._
+import org.broadinstitute.dsde.workbench.google2.GoogleStorageTransferService._
+import org.broadinstitute.dsde.workbench.model.WorkbenchEmail
+import org.broadinstitute.dsde.workbench.model.google._
+
+import scala.collection.JavaConverters._
+
+final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: StorageTransferServiceClient)(implicit
+  F: Async[F]
+) extends GoogleStorageTransferService[F] {
```
andy7i: what is the `F: Async[F]`?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  private def makeSchedule(schedule: JobTransferSchedule): Schedule = schedule match {
+    case JobTransferSchedule.Once(date) =>
```
andy7i: are nonexhaustive match statements (here and below) generally ok?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
```diff
@@ -0,0 +1,104 @@
...
+
+  def getTransferJob(jobName: JobName, project: GoogleProject): F[TransferJob]
+
+  def listTransferOperations(jobName: JobName, project: GoogleProject): F[Seq[Operation]]
+
+  def getTransferOperation(operationName: OperationName): F[Operation]
+
+}
+
+object GoogleStorageTransferService {
```
marctalbott, line 34: Are we going to be using all of these options when transferring workspace buckets? I expect we'll want to configure all of our jobs in an identical fashion, so we might be able to simplify this by omitting these options that we won't be using and just specify a default

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,110 @@
...
+    val request = TransferProto.CreateTransferJobRequest.newBuilder
+      .setTransferJob(transferJob)
+      .build
+
+    F.delay {
+      client.createTransferJob(request)
+    }
+  }
+
+  override def getTransferJob(jobName: JobName, project: GoogleProject): F[TransferJob] = {
```
andy7i, line 68: what's the difference between a transfer job and a transfer operation?

andy7i (commented):
I had some questions for clarification to better understand the PR

README.md
ehigham: Going by [semver item 7](https://semver.org/#spec-item-7), I think we should as new, backwards compatible functionality has been added.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: This is exhaustive. If new schedules are added then I hope scala would issue a compiler error.
So to answer your question, no. Non-exhaustive match statements are bad. Very bad.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: As far as I understand it: the `TransferJob` describes what you want STS to do. In that job, STS may execute one or more `Operation`s - ie how it was done.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: This is a constraint that `F` should extend the `cats.effect.Async` trait. It actually just needs to be `cats.effect.Sync` but I'm following the pattern of other services in this library.

ehigham: The formatting is a little jarring i will admit.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -359,6 +359,12 @@ object StorageRole {
   final case object StorageAdmin extends StorageRole {
     def name: String = "roles/storage.admin"
   }
+  final case object LegacyBucketReader extends StorageRole {
```
marctalbott, line 362: We could use our Terra custom roles instead of the legacy roles as they grant the required permissions described here: https://cloud.google.com/storage-transfer/docs/iam-transfer#source-permissions

`Terra Bucket Reader` grants
- resourcemanager.projects.get
- resourcemanager.projects.list
- storage.buckets.get
- storage.objects.get
- storage.objects.list

and `Terra Bucket Writer` grants
- resourcemanager.projects.get
- resourcemanager.projects.list
- storage.buckets.get
- storage.objects.create
- storage.objects.delete
- storage.objects.get
- storage.objects.list
- storage.objects.update

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
andy7i: should we wrap `client.getGoogleServiceAccount(request) ` with `withLogging` (as Qi suggested)?

Qi77Qi: definitely should....

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: I'm not sure how we'll call this in Rawls and so I'm trying to keep optiions open so that if things need to change, we can do that in rawls without needing a new version of the library.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
ehigham: Yeah I've been following what that doc! I didn't know about these (thank you!). Would following the reference not reduce conceptual overhead and not grant unnecesary permissions? I'm trying to weigh up pros and cons of this. I'd be interested to hear your thoughts.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: I've simplified this and removed most of the things I don't think we'd use for ppw migration.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
ehigham: Changed to be `Sync`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
ehigham: I think for this test, I'll leave it using the documented roles. We can investigate using our custom terra roles in Rawls.

marctalbott (approved):

andy7i (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,110 @@
...
+import org.broadinstitute.dsde.workbench.model.google._
+
+import java.time.ZonedDateTime
+import scala.collection.JavaConverters._
+
+final private[google2] class GoogleStorageTransferInterpreter[F[_]](client: StorageTransferServiceClient)(implicit
+  F: Sync[F]
+) extends GoogleStorageTransferService[F] {
+
+  private object Date {
```
Qi77Qi: nit, I usually prefer to have these helper functions at the bottom of the file just so it's easier to see what this `Interpreter` file is doing without

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
```diff
@@ -0,0 +1,65 @@
...
+  def getTransferJob(jobName: JobName, project: GoogleProject): F[TransferJob]
+
+  def listTransferOperations(jobName: JobName, project: GoogleProject): F[Seq[Operation]]
+
+  def getTransferOperation(operationName: OperationName): F[Operation]
+}
+
+object GoogleStorageTransferService {
+
+  sealed trait JobTransferSchedule extends Any
```
Qi77Qi: this should extend `extends Product with Serializable`...even if you don't care about that, `Any` isn't needed

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
```diff
@@ -0,0 +1,65 @@
...
+    private val prefix: String = "transferOperations/"
+
+    def fromString(name: String): Either[String, OperationName] =
+      if (name.startsWith(prefix)) Right(OperationName(name))
+      else Left(s"""Illegal operation name - "$name" must start with "$prefix".""")
+  }
+
+  def resource[F[_]](implicit F: Sync[F]): Resource[F, GoogleStorageTransferService[F]] =
+    Resource
+      .fromAutoCloseable(F.delay(StorageTransferServiceClient.create))
```
Qi77Qi, line 69: don't you need to give the client a SA file as well?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: Doing this fails to compile
```
google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala:41:18
object creation impossible.
Missing implementations for 3 members. Stub implementations follow:
  // Members declared in scala.Equals
  def canEqual(that: Any): Boolean = ???
  
  // Members declared in scala.Product
  def productArity: Int = ???
  def productElement(n: Int): Any = ???

    final object Immediately extends JobTransferSchedule

```
I'll remove `Any`.

Qi77Qi: oh, that's becuz your `Immediately` needs to be `final case object`....notice you're missing `case`

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: I'm not sure tbh. I've been using `GOOGLE_APPLICATION_CREDENTIALS` until now. I'll add it in case.

Qi77Qi: most terra services use their own service specific SA instead of application default...

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferService.scala
ehigham: Ahhhh!

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageTransferInterpreter.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  override def listTransferOperations(jobName: JobName, project: GoogleProject): F[Seq[Operation]] = {
+    val request = ListOperationsRequest.newBuilder
+      .setFilter(s"""{"projectId":"$project","jobNames":["$jobName"]}""")
+      .build
+
+    val operationsClient = client.getOperationsClient
+    val listOperations = F.delay(client.getOperationsClient.listOperations(request))
+    withLogging(listOperations, None, s"${operationsClient.getClass.getName}.listOperations")
```
Qi77Qi, line 85: I typically also add arguments as well..sth like `${operationsClient.getClass.getName}.listOperations(${jobName}, ${googleProject})`...things that helps debugging

ehigham: I don't think I'll do this as often I'm marshalling between data types and I'd be logging as if it were some hybrid API that matches neither.

Qi77Qi (approved):
lgtm...thanks @ehigham for bearing with all the comments

PR #798, opened 2021-11-03
[QA-1564] Expire Credentials cache after 50 mins. Log when it happens.

kyuksel:
https://broadworkbench.atlassian.net/browse/QA-1564

- Have `Credentials` cache expire tokens after 50 mins to see if that can solve a flaky test failure mode that we believe is due to token expiration when tests take long.
- Log when a new token is created.

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
@@ -8,25 +8,23 @@ import org.broadinstitute.dsde.workbench.auth.{AuthToken, AuthTokenScopes, UserA
 
 case class ScopedCredentials(credentials: Credentials, scopes: Seq[String])
 
-case class Credentials(email: String, password: String) extends LazyLogging {
+case class Credentials(email: String, password: String) {
   // use this form of override instead of specifying a default for the scopes argument below, because lots of
   // code expects the no-arg signature.
   def makeAuthToken(): AuthToken = makeAuthToken(AuthTokenScopes.userLoginScopes)
-  def makeAuthToken(scopes: Seq[String]): AuthToken = {
-    val auth = Credentials.cache.get(ScopedCredentials(this, scopes))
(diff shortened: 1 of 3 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
```diff
@@ -8,25 +8,23 @@ import org.broadinstitute.dsde.workbench.auth.{AuthToken, AuthTokenScopes, UserA
...
+case class Credentials(email: String, password: String) {
   // use this form of override instead of specifying a default for the scopes argument below, because lots of
   // code expects the no-arg signature.
   def makeAuthToken(): AuthToken = makeAuthToken(AuthTokenScopes.userLoginScopes)
-  def makeAuthToken(scopes: Seq[String]): AuthToken = {
-    val auth = Credentials.cache.get(ScopedCredentials(this, scopes))
-    // logger.debug(s"AuthToken: ${auth}, ${auth.value}")
-    auth
-  }
+  def makeAuthToken(scopes: Seq[String]): AuthToken = Credentials.cache.get(ScopedCredentials(this, scopes))
```
xuscrbw, line 15: Question about this simplification, was this needed for the lazy logging or just done to make things more readable?

Qi77Qi (approved):

jdcanas (approved):
Wow I'm surprised we set this to exactly 60 minutes for a token refresh. Good to remember for future things-that-expire.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/config/Credentials.scala
kyuksel: It wasn't needed for lazy logging. Since the `logger.debug(…)` line was commented out, 

```scala
   val auth = Credentials.cache.get(ScopedCredentials(this, scopes))
   auth
```
could be written as
```scala
   Credentials.cache.get(ScopedCredentials(this, scopes))
```
and since that updated version would have a single line, we didn't need the enclosing `{}`.

PR #799, opened 2021-11-03
IA-3037 add more wait before resizing dataproc cluster

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3037

I tested things out in staging, and things failed (probably doesn't happen often since auto-tests have been passing). But I think there're 2 issues
1. We should only attempt to resize if number of preemptiables is non 0. (My test cluster didn't have preemptible, but it still entered the resize branch
2. If cluster is previously `Stopped`, it'll take a bit longer for it to be in `Running`

Leo PR https://github.com/DataBiosphere/leonardo/pull/2366

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -280,7 +280,7 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Parallel](
 
       // Add back the preemptible instances, if any
       _ <- numPreemptibles match {
-        case Some(n) if n != 0 =>
+        case Some(n) if n > 0 =>
           for {
             _ <-
               if (cluster.getStatus.getState == ClusterStatus.State.STOPPED) F.sleep(30 seconds)
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
@@ -282,11 +282,12 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Parallel](
       _ <- numPreemptibles match {
         case Some(n) =>
           for {
+            _ <- F.sleep(30 seconds)
             _ <- streamUntilDoneOrTimeout(
               getCluster(project, region, clusterName),
               30,
               3 seconds,
-              s"Cannot start the instances of cluster ${project.value}/${clusterName.value} unless the cluster is in RUNNING status."
+              s"Cannot resize the instances of cluster ${project.value}/${clusterName.value} unless the cluster is in RUNNING status."
             )
             _ <- resizeCluster(project, region, clusterName, numWorkers = None, numPreemptibles = Some(n))
             _ <- streamUntilDoneOrTimeout(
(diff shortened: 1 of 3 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -282,11 +282,12 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Parallel](
       _ <- numPreemptibles match {
         case Some(n) =>
           for {
+            _ <- F.sleep(30 seconds)
```
kyuksel: I couldn't understand why/how this sleep helps. We have a `streamUntilDoneOrTimeout` following it and the `Done` condition is for the cluster to be in `Running` status, with or without sleep. What am I missing?

kyuksel:
> We should only attempt to resize if number of preemptiables is non 0. (My test cluster didn't have preemptible, but it still entered the resize branch

I couldn't see where we're addressing that?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
Qi77Qi: oops...forgot to push the latest code

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -280,13 +280,16 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Parallel](
 
       // Add back the preemptible instances, if any
       _ <- numPreemptibles match {
-        case Some(n) =>
+        case Some(n) if n != 0 =>
```
Qi77Qi: @kyuksel this is where we're checking there's indeed preemptible

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
```diff
@@ -280,13 +280,16 @@ private[google2] class GoogleDataprocInterpreter[F[_]: Parallel](
...
           for {
+            _ <-
+              if (cluster.getStatus.getState == ClusterStatus.State.STOPPED) F.sleep(30 seconds)
```
Qi77Qi, line 286: the error is the following wait line hits timeout..so this sleep here will hopefully reduce that

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: Would it be safer as follows?

```suggestion
        case Some(n) if n > 0 =>
```

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleDataprocInterpreter.scala
kyuksel: With increased `maxAttempts` and/or `delay`, aren't we already accounting for the additional time needed? Would love to not have to resort to sleep.

Qi77Qi: when I was going through log, I think it's pretty safe to assume dataproc cluster won't transition from `Stopped` to `Running` within 30 seconds....with this sleep, it's actuallly nicer cuz we'll be polling less

rtitle (approved):
Seems ok.. the sleep may not be needed since it's just polling immediately after, but shouldn't harm anything.

Qi77Qi:
thanks @rtitle !

PR #793, opened 2021-10-29
more helpful logging for RestClient and CleanUp [AS-1018]

davidangb:
Two changes in this PR:
1. better request/response logging
2. better logging when a test fails during cleanup

## Better request/response logging
I find the request/response logging in service-test's `RestClient` to be almost completely useless, just adding noise. I have updated the logging to hopefully be more helpful. Especially as we've been trying to unblock monolith QA, these logging statements really annoyed me. So I changed them.

The old logging looks like:
```
16:55:44.652 [default-akka.actor.default-dispatcher-22] INFO org.broadinstitute.dsde.workbench.service.Orchestration$ - API request: HttpRequest(HttpMethod(GET),https://firecloud-orchestration-fiab.dsde-dev.broadinstitute.org:23443/api/workspaces/gpalloc-dev-master-pambyyh/6e0846cf-1493-4fd1-a585-e77875c8169e-reader-storage-cost%2053c6deb7-95a0-428a-99a2-63a1ca3e551e/storageCostEstimate,List(Authorization),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))
API response: HttpResponse(403 Forbidden,List(Date: Fri, 29 Oct 2021 20:55:44 GMT, Server: akka-http/10.2.4, X-Frame-Options: SAMEORIGIN, X-XSS-Protection: 1; mode=block, X-Content-Type-Options: nosniff, Strict-Transport-Security: max-age=31536000; includeSubDomains, Access-Control-Allow-Origin: *, Access-Control-Allow-Headers: authorization, content-type, accept, origin, x-app-id, Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS, HEAD, Access-Control-Max-Age: 1728000),HttpEntity.Strict(application/json,490 bytes total),HttpProtocol(HTTP/1.1))
```

The new logging looks like:
```
16:48:33.772 [default-akka.actor.default-dispatcher-28] DEBUG org.broadinstitute.dsde.workbench.service.Orchestration$ - API request: GET https://firecloud-orchestration-fiab.dsde-dev.broadinstitute.org:23443/api/workspaces/gpalloc-dev-master-ozldzgf/7286106f-5730-487c-8915-59277d3bb884-reader-storage-cost%204fdd10c9-a21b-4704-a538-47b3d206e66b/storageCostEstimate with headers (Authorization) 
[...]

PR #801, opened 2021-11-04
QA-1566 During test cleanUp, report test and cleanup failures if BOTH fail

gpolumbo-broad:
Noticed this while investigating QA-1566.  

The matches were written such that in the scenario when both the test itself and the test cleanup both fail, the cleanup failure was being discarded and not reported at all.  This might make debugging problems and failures difficult.  

Since we have more than one exception happening at the same time, we needed a way to report those out together, so I followed the lead of line 114 and wrap the exceptions up in an `ErrorReport`.  

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
@@ -116,11 +116,11 @@ object CleanUp {
   def runCodeWithCleanup[T, C](testTrial: Try[T], cleanupTrial: Try[C]): T =
     (testTrial, cleanupTrial) match {
       case (Success(outcome), Success(_)) => outcome
+      case (Failure(t), Success(_)) => throw t
       case (Success(_), Failure(t)) =>
         throw new WorkbenchExceptionWithErrorReport(
           ErrorReport(s"Test passed but cleanup failed: ${t.getMessage}", ErrorReport(t))
         )
-      case (Failure(t), Success(_)) => throw t
       case (Failure(t), Failure(c)) =>
         throw new WorkbenchExceptionWithErrorReport(
           ErrorReport(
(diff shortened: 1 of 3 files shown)


ehigham (approved):
Looks good

davidangb:
Let's coordinate! I'm working on #793

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
```diff
@@ -118,10 +111,24 @@ trait CleanUp extends TestSuiteMixin with ExceptionHandling with LazyLogging { s
...
+          ErrorReport(
+            "Test and CleanUp both failed",
+            ErrorReport(
+              Map(
+                "testFailure" -> ErrorReport(t),
+                "cleanUpFailure" -> ErrorReport(c)
+              ).toJson.prettyPrint
+            )
+          )
+        )
```
ehigham: One thing that this makes me wish for is a type-safe version of clojure's `ExceptionInfo` class - it's a exception that has a message, cause as well as a map of additional data (no more formatted strings - hooray): 
```clojure
(throw (ex-info "Test and clean-up both failed" {:test-failure t :cleanup-failure c}))
```

(I'm not suggesting you do this, just it would be nice if this Json business was deferred to later)

davidangb: To avoid all the json business and string-escaping and wrapping, how about:
```
ErrorReport(
    "Test and CleanUp both failed. This ErrorReport's causes contain the test and cleanup exceptions, in that order",
    Seq(ErrorReport(t), ErrorReport(c))
)
```
you could then also remove a bunch of the json support imports.

davidangb (commented):
one comment inline about avoiding "this json business"

Also, this PR needs updates to `README.md` and `serviceTest/CHANGELOG.md`

davidangb (approved):
nit: I *think* the `import spray.json.DefaultJsonProtocol._` is now unused and could be deleted.

Do you want to add any notes to the changelog to describe the logging change? (ok if not)

otherwise :+1: and thanks for getting this in!

davidangb (approved):
still :+1: thank you!

PR #891, opened 2022-02-02
[CA-1743] Add `removeIamPolicy` to `GoogleStorageService`

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1743

Add `removeIamPolicy` as an interface method on `GoogleStorageService`. I tried to see if a later version of the client library existed that had this functionality but couldn't find much.
I wonder if we want to use `Map[StorageRole, NonEmptyList[Identity]]` instead of `Policy` when using this service (ie not expose `Policy`)...

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-a07be6a"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.24-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.24-TRAVIS-REPLACE-ME"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-TRAVIS-REPLACE-ME"`
 
 To start the Google PubSub emulator for unit testing:
 
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
@@ -10,14 +10,17 @@ import com.google.auth.Credentials
 import com.google.auth.oauth2.{AccessToken, GoogleCredentials}
 import com.google.cloud.storage.BucketInfo.LifecycleRule
 import com.google.cloud.storage.{Acl, Blob, BlobId, Bucket, StorageOptions}
-import com.google.cloud.{Identity, Policy}
+import com.google.cloud.{Identity, Policy, Role}
 import fs2.{Pipe, Stream}
 import com.google.cloud.storage.Storage.{BucketGetOption, BucketSourceOption}
+import org.broadinstitute.dsde.workbench.google2.GoogleStorageService.policyToStorageRoles
+import org.broadinstitute.dsde.workbench.google2.Implicits.PolicyToStorageRoles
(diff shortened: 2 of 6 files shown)


s-rubenstein (approved):
Looks great, I like the tests! I agree that avoiding exposing `Policy` seems like the right choice to make here.

README.md
```diff
@@ -72,7 +72,7 @@ Workbench utility libraries, built for Scala 2.12 and 2.13. You can find the ful
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.23-a07be6a"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.24-TRAVIS-REPLACE-ME"`
```
Qi77Qi: we don't bump version if doesn't break any existing APIs...

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -289,6 +291,25 @@ trait GoogleStorageService[F[_]] {
...
   ): Stream[F, Policy]
+
+  /**
+   * Remove the specified roles from the bucket IAM policy
+   */
+  def removeIamPolicy(bucketName: GcsBucketName,
+                      rolesToRemove: Map[StorageRole, NonEmptyList[Identity]],
+                      traceId: Option[TraceId] = None,
+                      retryConfig: RetryConfig = standardGoogleRetryConfig
+  ): Stream[F, Unit] =
```
Qi77Qi, line 302: I know I've set examples where return type is `Stream` in many other places...But thinking back, I think this can probably be returning `F[Unit]` instead. Most apps aren't using either `IO` or `Future`, and rarely `Stream` anyway

README.md
ehigham: We use a version string that kinda looks like semver but doesn't follow semver. I find that really confusing. As a user and contributor here, I really don't want to have to learn a new versioning scheme to understand changes made to this one library. Can we just use semver please?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
ehigham: I agree in part but also feel that it should be `Stream` to be compatible with other methods in this class without explicit coercion.
I don't think there's a right answer here, if I make it return `IO` and someone wants to use it in a `for`-comp over `Steam` them they have to `Stream.eval`:
```scala
for {
    ...
    _ <- Stream.eval { 
        storageService.removeIamPolicy(
            bucketName,
            rolesToRemove
            ...
        )
    }
    ...
```
 If I make it over `Stream` then you need a `.compile.drain` in a `for`-comp over `IO`:
```scala
for {
    ...
    _ <- storageService.removeIamPolicy(
        bucketName,
        rolesToRemove
        ...
    ).compile.drain
    ...
```
 The only advantage here is I don't need to do any of that if I'm working with another method of this class - I don't much like special cases.

ehigham:
> Looks great, I like the tests! I agree that avoiding exposing `Policy` seems like the right choice to make here.

Thanks! This would be a breaking change though so I meant it more as proposed change for the future - I don't think I'll attempt now :).

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
ehigham: Alternatively, might we change the methods on this interface to simply return `F[A]` and let the caller decide if they want a `Stream` or simply an `IO`?

Qi77Qi: that's true...yea, I'm ok with either I guess...I see your point about consistency

README.md
Qi77Qi: The schema predates my contributions to this library.

>Can we just use semver please?

Are you saying moving away from hashes? I'm all for that...Think I might even attempted that, but didn't follow through...You'll need to update the CI for this repo to make the change, and also need to inform larger engieering group about the change

Qi77Qi: in any case, what's the breaking change in this PR?

ehigham: It's not a breaking change, it's added functionality which by semver means we should bump the minor version: https://semver.org/#spec-item-7

> Are you saying moving away from hashes?

That would be cool, though not in this PR. I'll create a todo for myself :)

Qi77Qi (approved):
one more TRAVIS_REPPLACE_ME in CHANGELOG.md..otherwise LGTM

PR #899, opened 2022-02-10
[CA-1103] Remove GPAlloc Support

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1103
GPAlloc is the source of many sporadic test failures. It's long been replaced by the Resource Buffer Service and its use proliferates v1 billing projects. This change removes all support for GPAlloc in workbench-libs.
Note that this is a breaking change. Continuing support for deprecated functionality and GPAlloc is the wrong thing to do as it causes so many test failures downstream. Once this is merged I will go and address the fallout in each broad repo.

**PR checklist**

In all cases:


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,219 +1,78 @@
...
 
 /**
  * Mix in this trait to allow your test to access billing projects managed by the GPAlloc system, or create new
  * billing projects of your own.  Using GPAlloc will generally be much faster, limit the creation of billing projects
  * to those tests which truly require them.
  */
-trait BillingFixtures extends ExceptionHandling with LazyLogging with CleanUp with RandomUtil {
-  self: TestSuite =>
-
-  protected def addMembersToBillingProject(projectName: String, memberEmails: List[String], role: BillingProjectRole)(
```
s-rubenstein, line 25: Why is this method going away? Don't we still want to add and remove members from the project? I would expect removing gpalloc support to just include changing the withNewProject logic to create and delete rather than claim and release. These seem very much still usable (although might need minor updates to point at v2 billing? Unless Orch has changed under the hood to do the right thing/point at v2 billing, which I think it might have?)

ehigham: Because It's not being used - we should be adding and removing memebers via the Rawls or Ochestration api (which has v2 support).

Qi77Qi (commented):
before we move this, I think we need to make sure things will still work for all clients...for instance, leo uses [this](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala#L68)...what's the replacement or what leo needs to do once `claimGPAllocProject` is removed?

ehigham:
> before we move this, I think we need to make sure things will still work for all clients...for instance, leo uses [this](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/workbench/leonardo/LeonardoSuite.scala#L68)...what's the replacement or what leo needs to do once `claimGPAllocProject` is removed?

I'll be creating PRs for this sort of thing. In cases where `withBillingProject` are being used then we'll need to include which billing account to bill. In cases when you want a billing project (like the one you linked) then you need to call Rawls or Orch to get one. That code I'd be tempted to refactor into a `Resource` or `with-` pattern to be more exception safe.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
s-rubenstein: Okay, unused seems like a good enough reason to me.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,219 +1,78 @@
...
-      case _ =>
-        throw new Exception("claimGPAllocProject got no project back from GPAlloc")
+  def withCleanBillingProjectF[F[_], A](billingAccountName: String,
+                                        projectNamePrefix: String = "tmp-billing-project-",
+                                        ownerEmails: List[String] = List.empty,
+                                        userEmails: List[String] = List.empty
+  )(testCode: String => F[A])(implicit ownerAuthToken: AuthToken, F: Sync[F]): F[A] = {
+    def createProject: F[String] = F.delay {
+      val projectName = projectNamePrefix ++ UUID.randomUUID.toString.replace("-", "")
+      Orchestration.billingV2.createBillingProject(projectName, billingAccountName)
```
s-rubenstein: Probably not a huge deal because its not leaving orphaned anymore, but should this clean up after itself? (e.g., delete the billing project once the tests are done?)

s-rubenstein (approved):
Overall lgtm. One comment about cleanup, but I am excited to see gpalloc go.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: It does - that's taken care of by the `Resource`

Qi77Qi:
>I'll be creating PRs for this sort of thing. In cases where withBillingProject are being used then we'll need to include which billing account to bill. In cases when you want a billing project (like the one you linked) then you need to call Rawls or Orch to get one. That code I'd be tempted to refactor into a Resource or with- pattern to be more exception safe.

Great....You can publish a snap version locally via `sbt publish`, and then try out the new code in other services, if tests pass, then I think the change will be good to go

ehigham:
> > I'll be creating PRs for this sort of thing. In cases where withBillingProject are being used then we'll need to include which billing account to bill. In cases when you want a billing project (like the one you linked) then you need to call Rawls or Orch to get one. That code I'd be tempted to refactor into a Resource or with- pattern to be more exception safe.
> 
> Great....You can publish a snap version locally via `sbt publish`, and then try out the new code in other services, if tests pass, then I think the change will be good to go

Thanks! I was wondering how I could make a head start with this!

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GroupFixtures.scala
```diff
@@ -26,9 +26,8 @@ trait GroupFixtures extends ExceptionHandling with LazyLogging with RandomUtil {
     testCode: (String) => Any
   )(implicit token: AuthToken): Unit = {
     val groupName = uuidWithPrefix(namePrefix)
-
+    Orchestration.groups.create(groupName)
     val testTrial = Try {
-      Orchestration.groups.create(groupName)
```
andy7i: Why pull this out of the Try?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
```diff
@@ -111,4 +110,31 @@ trait WorkspaceFixtures extends ExceptionHandling with RandomUtil { self: TestSu
...
+  /**
+   * Create a workspace `Resource` whose lifetime is bound to the `Resource`'s `use` method.
+   *
+   * @param billingProjectName Billing project for workspace (or workspace namespace).
+   * @param namePrefix         Prefix for workspace name. [default "tmp-workspace-"]
+   * @param authDomain         Optional auth domain for workspace
+   * @param bucketLocation     Optional region where workspace bucket should be created
+   * @param creatorAuthToken   Auth token of workspace creator
+   */
+  def resource[F[_]](billingProjectName: String,
```
andy7i: Does this approach have benefits over the existing use of `runTestAndCleanUpWorkspace`? Should we refactor WorkspaceFixtures to use this?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,219 +1,89 @@
...
-
-    CleanUp.runCodeWithCleanup(testTrial, cleanupTrial)
-  }
+object BillingFixtures {
 
   /**
-   * Manually claim a project provisioned by GPAlloc and optionally add members.
-   * Consider using `withCleanBillingProject()` instead if you don't need to control the use of projects.
+   * Create a new v2 billing project for the activation of `testCode`. The billing project will be
+   * destroyed when control exists `testCode`.
```
andy7i:
```suggestion
   * destroyed when control exits `testCode`.
```

andy7i (commented):
`Resource` is pretty interesting stuff!

I took a look at the RBS buffer size to make sure we can handle the switchover from gpalloc to RBS. It looks like we have a 5k buffer size which seems reasonable? Links to the configs if you're curious: (https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/tools/pool_schema.yml ; https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/rawls/rawls.conf.ctmpl#L42)

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/GroupFixtures.scala
ehigham: I only want to enter the `Try` when a group is created. If creating the group fails, I don't want to try to clean it up because it doesn't exist.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/WorkspaceFixtures.scala
ehigham: Yeah good point

PR #943, opened 2022-03-31
[CA-1103] Remove v1 billing project creation from workbench-libs

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1103
Creating temporary v1 billing projects via GPAlloc has been a source of
sporadic failures in swatomation tests across terra core services.
Since project-per-workspace was launched, v1 billing projects have been
deprecated in favour of newer, shinier v2 billing projects and CI have
sought to purge GPAlloc from our testing infrastructure.
In this change, I have removed all mention of GPAlloc and v1 billing
projects from workbenchLibs/workbenchServiceTest. I've replaced the
BillingFixtures trait with an object that facilitates creation, use and
clean-up of temporary v2 billing projects. Why an object and not a
trait? Because inheritance is the root-cause of evil in this world. It
is safe to say that Vladimir Putin would not be where he is today if
it were not for inheritance. Thus, it is our moral obligation to
remove it from the world.

**PR checklist**

In all cases:


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,171 +1,80 @@
...
-
-        addMembersToBillingProject(project.projectName, ownerEmails, BillingProjectRole.Owner)(newOwnerToken())
-        addMembersToBillingProject(project.projectName, userEmails, BillingProjectRole.User)(newOwnerToken())
-
-        ClaimedProject(project.projectName, gpAlloced = true)
-      case _ =>
-        throw new Exception("claimGPAllocProject got no project back from GPAlloc")
+  def temporaryBillingProject[F[_]](billingAccountName: String,
+                                    creatorAuthToken: AuthToken,
+                                    prefix: Option[String] = None,
```
andy7i, line 50: I like the idea of being explicit up front and using a default parameter value, but it's not a strong preference.
```suggestion
                                    prefix: String = "tmp-",
```

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,171 +1,80 @@
...
+                                    owners: Option[List[String]] = None,
+                                    users: Option[List[String]] = None
+  )(implicit F: Sync[F]): Resource[F, String] = {
+    def createBillingProject: F[String] = F.delay {
+      val projectName = prefix
+        .getOrElse("tmp-billing-project-")
+        .++(UUID.randomUUID.toString.replace("-", ""))
```
andy7i, line 57: Not entirely sure how worried we should be, but I'm mildly concerned about how the length of the UUIDs will be 10 characters long if the default prefix is 20 characters. (I guess I'm mostly concerned due to being burned by flaky tests and the risk of collisions if the UUID part is too short).

andy7i: I looked into this a bit more and UUID.randomUUID generates a random string (aka a version 4 uuid), which is not prefixed with a timestamp. Also, a 10 character UUID gives us ~1 trillion IDs. In conclusion, I think we're good..

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,171 +1,80 @@
...
-   * As opposed to `Credentials`, accepts `AuthToken` and `String` values for the new owner.
-   * This way a GPAlloc project can be claimed as a pet SA.
+   * Create a v2 billing project `Resource` whose lifetime is bound to the `Resource`'s `use` method.
    *
-   * @param newOwnerEmail Email for the new billing project owner
-   * @param ownerEmails a List of emails (as Strings) to add as owners of this project
-   * @param userEmails a List of emails (as Strings) to add as users of this project
-   * @param newOwnerToken Function that returns an AuthToken for the new billing project owner to pass to billing project endpoints
-   * @return Some(GPAllocProject) if it succeeded, none if it failed
+   * @param billingAccountName Name of Google billing account the new billing project will bill to.
```
andy7i: the API takes in a billing account ID (like `"billingAccount":"billingAccounts/00708C-45D19D-27AAFA"`) so I think we should rename this to billingAccount or billingAccountId or something like that

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: I spent a long time thinking about this and I went with `Option` because:
1. In the signature it tells me which parameters are optional
3. it allows me to know if a user has specified preference or not and compute defaults from other arguments. 

Consider the following example. Let's say I want to default the clone billing project to the source billing project and compute a prefix for the clone based upon the name of the source workspace. I cannot do this using default parameter values because default parameter values cannot be dependent on actual parameter values.
```scala
def temporaryWorkspaceClone(
  sourceBillingProject: String,
  sourceWorkspaceName: String.
  cloneBillingProject: String = sourceBillingProject,
  cloneWorkpacePrefix: String = sourceWorkspaceName + "-clone-",
   ...
```
For this reason, using `Option` provides a consistent interface for documenting and handling default parameters.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: We're very very unlikely to get a UUID collision. Do we have a limit on the length of a v2 billing project name?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: sure.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
gpolumbo-broad: love this discussion!  If we ever wanted to start creating a team style guide, this is the sort of thing I would put there.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
```diff
@@ -1,171 +1,81 @@
...
-    CleanUp.runCodeWithCleanup(testTrial, cleanupTrial)
+    def addMembers(projectName: String, emails: List[String], role: BillingProjectRole): Resource[F, Unit] =
+      emails.traverse_ { email =>
+        Resource.eval(F.delay {
+          Orchestration.billingV2.addUserToBillingProject(projectName, email, role)(creatorAuthToken)
+        })
+      }
+
+    for {
+      billingProject <- Resource.make(createBillingProject)(destroyBillingProject)
```
gpolumbo-broad, line 76: In order to delete a Billing Project, it must no longer have any Workspaces within it.  We can maybe count on users cleaning up their own Workspaces or using a "withTemporaryWorkspace" context in their test.  Or we can just force that cleanup to happen here, before destroying the Billing Project or as part of the `destroyBillingProject` method, we first delete all Workspaces within it.

gpolumbo-broad (changes requested):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: I actually prefer not to as this has helped identify issues in the tests with permissions that helped me fix them.

ehigham: It's not just workspaces too - it's groups. There could be anything. I don't think we should enumerate and clean all them up here.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
andy7i: looks like we still limit it to 30 chars, though I don't think that's necessary anymore
https://github.com/broadinstitute/rawls/blob/develop/model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala#L44

andy7i (approved):
LGTM!

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/fixture/BillingFixtures.scala
ehigham: Thanks. I guess that's a hangover from v1. Yeah we probably should increase this.

gpolumbo-broad (approved):

PR #960, opened 2022-04-06
[no ticket] more dependency bumps

Qi77Qi:
Leo PR https://github.com/DataBiosphere/leonardo/pull/2610

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
@@ -65,7 +65,7 @@ class KubernetesInterpreter[F[_]](
       call =
         recoverF(
           F.blocking(
-            client.createNamespacedPod(namespace.name.value, pod.getJavaSerialization, "true", null, null)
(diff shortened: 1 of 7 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
```diff
@@ -233,7 +233,7 @@ class KubernetesInterpreter[F[_]](
       call =
         recoverF(
           F.blocking(
-            client.createNamespace(namespace.getJavaSerialization, "true", null, null)
+            client.createNamespace(namespace.getJavaSerialization, "true", null, null, null)
```
rtitle, line 236: Assuming this is okay but I recall the k8s java API changing order of parameters in the past. 😿

rtitle (approved):
Looks good, if Leo tests are passing.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/KubernetesInterpreter.scala
Qi77Qi: I checked the first 2 and it was the same...leo tests passed too, so I think this is okay

PR #964, opened 2022-04-14
[TOAZ-110] Add new oauth2 module to support B2C rollout

rtitle:
Ticket: https://broadworkbench.atlassian.net/browse/TOAZ-110
[Design doc](https://docs.google.com/document/d/1fnfLSbHWMKi8F__zJm-p7C2URlXIEKzB0-bgyrf-qBg/edit#heading=h.t8i9f7a84yrl)

Adds a new `oauth2` module to workbench-libs. This module facilitates adding 2 backend routes needed for B2C migration.
* GET /oauth2/authorize -- redirects to the OAuth2 authorize endpoint
* POST /oauth2/token -- proxies request to the OAuth2 token endpoint

See comment in `OpenIDConnectConfiguration` for more details on usage.

I am initially testing with Orchestration which will support Terra UI. Eventually all services should adopt this to support their own Swagger pages.

Related PRs:
- Java/Spring version of this in TDR (will move to TCL eventually): https://github.com/DataBiosphere/jade-data-repo/pull/1259
- Terra UI draft PR to make use of this: https://github.com/DataBiosphere/terra-ui/pull/2949

**PR checklist**

In all cases:

### DIFF
--- a/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectConfiguration.scala
+++ b/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectConfiguration.scala
@@ -9,7 +9,42 @@ import org.http4s.Uri
 import org.http4s.blaze.client._
 import org.http4s.circe.CirceEntityDecoder._
 
+/**
+ * Allows services to configure their mode of OAuth by providing 2 backend routes:
+ * `/oauth2/authorize` and `/oauth2/token`.
+ *
+ * To use this class, first instantiate by calling `OpenIDConnectConfiguration.apply`
+ * with the following options:
+ *   - authorityEndpoint (required): the OAuth2.0 authority, e.g. https://accounts.google.com
+ *   - clientId (optional): if present injects the clientId as a scope in the authorize request.
+ *       This is needed for B2C but not Google.
+ *   - clientSecret (optional): if present injects the clientSecret in the token request.
+ *       This is needed for Google but not B2C.
+ *   - extraAuthParams (otpional): if present appends extra params to the query string of the
+ *       authorization request. This is needed for B2C for some clients, including Swagger UI.
+ *
+ * There are 2 choices for using this class:
+ *
+ *   1. If the service is using akka-http, you can can generate an akka-http route using
+ *      `toAkkaHttpRoute` and add it directly to your service. Note: ensure the service is
+ *      using a compatible akka-http version with the version workbench-libs is compiled against.
+ *
+ *   2. Otherwise, the service should add 2 backend routes as follows:
+ *     - GET /oauth2/authorize:
+ *         This route should call `processAuthorizeQueryParams` on the incoming querystring params
+ *         and redirect to the endpoint returned by `getAuthorizationEndpoint`.
+ *     - POST /oauth2/token:
+ *         This route should only accept Content-Type: application/x-www-form-urlencoded.
+ *         It should call `processTokenFormFields` on the incoming form fields and _proxy_ the request
+ *         to the endpoint returned by `getTokenEndpoint`.
+ */
 trait OpenIDConnectConfiguration {
+  def getAuthorizationEndpoint: String
+  def processAuthorizeQueryParams(params: Seq[(String, String)]): Seq[(String, String)]
+
+  def getTokenEndpoint: String
+  def processTokenFormFields(fields: Seq[(String, String)]): Seq[(String, String)]
+
   def toAkkaHttpRoute(implicit actorSystem: ActorSystem): Route
 }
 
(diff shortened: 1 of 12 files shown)


project/Dependencies.scala
```diff
@@ -8,7 +8,7 @@ object Dependencies {
   val scalaLoggingV = "3.9.4"
   val scalaTestV    = "3.2.11"
   val circeVersion = "0.14.1"
-  val http4sVersion = "1.0.0-M32"
+  val http4sVersion = "1.0.0-M30"
```
rtitle, line 11: Had to downgrade because `1.0.0-M32` only targeted 2.13. Hopefully we will drop 2.12 from the platform soon.

rtitle:
@nmalfroy fyi this is the Scala version of this, I put it in a common library and including it in Orch so far.

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectConfiguration.scala
```diff
@@ -0,0 +1,81 @@
...
+  def getTokenEndpoint: String
+  def processTokenFormFields(fields: Seq[(String, String)]): Seq[(String, String)]
+
+  def toAkkaHttpRoute(implicit actorSystem: ActorSystem): Route
+}
+
+object OpenIDConnectConfiguration {
+  private val oidcMetadataUrlSuffix = ".well-known/openid-configuration"
+
+  def apply[F[_]: Async](authorityEndpoint: String,
```
Qi77Qi, line 53: would be nice to have some custom type for these since they're all `String`s

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectConfiguration.scala
```diff
@@ -0,0 +1,81 @@
...
+ *         to the endpoint returned by `getTokenEndpoint`.
+ */
+trait OpenIDConnectConfiguration {
+  def getAuthorizationEndpoint: String
+  def processAuthorizeQueryParams(params: Seq[(String, String)]): Seq[(String, String)]
+
+  def getTokenEndpoint: String
+  def processTokenFormFields(fields: Seq[(String, String)]): Seq[(String, String)]
+
+  def toAkkaHttpRoute(implicit actorSystem: ActorSystem): Route
```
Qi77Qi: maybe make this an extension method on `OpenIDConnectConfiguration`? Sth like

```
final class OpenIDConnectConfigurationOps(private val config: OpenIDConnectConfiguration) extemds AnyVal{
     def toAkkaHttpRoute(implicit actorSystem: ActorSystem): Route = ??? 
}
```

project/Dependencies.scala
Qi77Qi: 😭

Qi77Qi (approved):

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectInterpreter.scala
```diff
@@ -0,0 +1,53 @@
...
+                                                extraGoogleClientId: Option[ClientId]
+) extends OpenIDConnectConfiguration {
+  private val scopeParam = "scope"
+  private val clientSecretParam = "client_secret"
+
+  override def getAuthorizationEndpoint: String = providerMetadata.authorizeEndpoint
+
+  override def processAuthorizeQueryParams(params: Seq[(String, String)]): Seq[(String, String)] = {
+    // TODO seems like this is not needed
+//    val paramsWithScope = if (!providerMetadata.isGoogle) {
```
nmalfroy: That's really more for dealing with B2C and requiring the client id be sent in as a scope in order to get an access token.  For TDR I made the behavior trigger with an explicit boolean config option

nmalfroy (approved):
looks good (I think?  not super familiar with Scala).  Just wonder if you want to leave that bit commented or add explicit config there?

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectInterpreter.scala
rtitle: Even with B2C as the authority it didn't seem like I needed it… B2C was actually complaining `"unrecognized scope <clientId>"`. I could be missing something though..

rtitle: Ohh I might have been using the `openid` scope to get an id_token. Looks like you pass the client-id-as-scope to get an access token: https://docs.microsoft.com/en-us/azure/active-directory-b2c/authorization-code-flow#2-get-an-access-token. So maybe I'll add this back and re-test.

nmalfroy: Right exactly.  And the access_token is what we need to pass around I think

Qi77Qi:
think we can drop 2.12 now since orch has migrated to 2.13

rtitle:
Oh awesome - will do that

PR #997, opened 2022-05-23
[TOAZ-152] Some swagger endpoints responding 502

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-152

Fixed issue causing akka-http `Stream cannot be materialized more than once` on `/oauth2/token` endpoint.

Tested in a fiab with Sam -- set it to Google auth, it failed before this fix and worked after this fix.

Will need to bump: Leo, Orch, Rawls, Sam

**PR checklist**

In all cases:

### DIFF
--- a/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
+++ b/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
@@ -15,11 +15,13 @@ import io.circe.syntax._
 import org.broadinstitute.dsde.workbench.oauth2.OpenIDConnectAkkaHttpOps.ConfigurationResponse
 
 import java.nio.file.Paths
+import scala.concurrent.duration._
 
 class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
   private val swaggerUiPath = "META-INF/resources/webjars/swagger-ui/4.10.3"
 
-  def oauth2Routes(implicit actorSystem: ActorSystem): Route =
+  def oauth2Routes(implicit actorSystem: ActorSystem): Route = {
+    implicit val ec = actorSystem.dispatcher
     pathPrefix("oauth2") {
       path("authorize") {
         get {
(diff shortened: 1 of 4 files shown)


oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -39,7 +41,7 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
                   uri = Uri(config.providerMetadata.tokenEndpoint),
                   entity = FormData(config.processTokenFormFields(fields): _*).toEntity
                 )
-                Http().singleRequest(newRequest)
+                Http().singleRequest(newRequest).map(_.toStrict(5.seconds))
```
rtitle, line 44: This is the fix -- since this endpoint is proxying requests to the authority's `token` endpoint, need to read all data into memory before responding.

oauth2/src/test/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpSpec.scala
```diff
@@ -64,19 +69,48 @@ class OpenIDConnectAkkaHttpSpec extends AnyFlatSpecLike with Matchers with Workb
     res.unsafeRunSync()
   }
 
-  // Ignored because of issues calling the third party endpoint from github actions
-  "the token endpoint" should "proxy requests" ignore {
+  "the token endpoint" should "proxy requests" in {
```
rtitle, line 72: Un-ignored test, using a mock server instead of https://accounts.google.com

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
Qi77Qi: what does `toStrict(5 seconds)` do?

Qi77Qi (approved):

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
rtitle: It eagerly reads all entity data to memory with a timeout of 5 seconds. Otherwise the data is streamed and something tries to unmarshal it twice, causing an error.

We could make this endpoint streaming I think -- but it's pretty small and infrequent so I think `toStrict` is fine.

rtitle:
GHA job strangely got stuck, re-ran it.

rtitle:
Now it worked

PR #998, opened 2022-06-02
Requester pays

dvoet:
Changes to support getting and setting requester pays status of a bucket for PPW migration. The change from returning `Bucket` to `BucketInfo` is not strictly backwards compatible but pretty close. `Bucket` extends `BucketInfo` with some operational features which should not be used anyway. This change makes writing tests easier.

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -389,8 +389,7 @@ private[google2] class GoogleStorageInterpreter[F[_]](
     withLogging(fa, traceId, s"com.google.cloud.storage.Storage.get(${bucketName.value}, $bucketGetOptions)")
   }
 
-  override def setRequesterPays(googleProject: GoogleProject,
-                                bucketName: GcsBucketName,
+  override def setRequesterPays(bucketName: GcsBucketName,
                                 requesterPaysEnabled: Boolean,
                                 traceId: Option[TraceId] = None,
                                 retryConfig: RetryConfig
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -383,12 +383,28 @@ private[google2] class GoogleStorageInterpreter[F[_]](
...
-    val fa = Async[F].delay(dbForProject.get(bucketName.value, bucketGetOptions: _*)).map(Option(_))
+    val fa: F[Option[BucketInfo]] = Async[F].delay(dbForProject.get(bucketName.value, bucketGetOptions: _*)).map(Option(_))
     withLogging(fa, traceId, s"com.google.cloud.storage.Storage.get(${bucketName.value}, $bucketGetOptions)")
   }
 
+  override def setRequesterPays(bucketName: GcsBucketName,
+                                requesterPaysEnabled: Boolean,
+                                traceId: Option[TraceId] = None,
+                                retryConfig: RetryConfig
+  ): Stream[F, Unit] = {
```
Qi77Qi, line 397: I'm totally fine with this returning `F[Unit]` btw....

It'd be nice more services using `TraceId` instead of ignoring :)

Qi77Qi (approved):

ehigham (approved):
looks good

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
dvoet: I was being consistent with other bucket update calls

PR #1002, opened 2022-06-15
PPWM-54 add parameters to accept user project

dvoet:
https://broadworkbench.atlassian.net/browse/PPWM-54

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -11,10 +11,18 @@ import cats.effect.std.Semaphore
 import cats.syntax.all._
 import com.google.auth.oauth2.ServiceAccountCredentials
 import com.google.cloud.storage.BucketInfo.LifecycleRule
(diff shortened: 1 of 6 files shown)


jdcanas (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -393,57 +413,71 @@ private[google2] class GoogleStorageInterpreter[F[_]](
   override def setRequesterPays(bucketName: GcsBucketName,
                                 requesterPaysEnabled: Boolean,
                                 traceId: Option[TraceId] = None,
```
ehigham: Is there a reason you're not defining / using the defaults from the trait here? (Im guessing that was the motivation behind removing the defaults in this type)

ehigham (approved):
Blobby.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
dvoet: I just missed this one. And yes, the defaults were inconsistently used from the trait and duplicating them seemed silly so ...

Qi77Qi (approved):
this will be breaking change..so probably worth updating version

PR #1005, opened 2022-06-27
IA-3289 Add Azure module

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3289

* Add Azure module
* Moved some utility to `util2` so that it can be shared between `google2` and `azure`

Related Leo PR https://github.com/DataBiosphere/leonardo/pull/2761

**PR checklist**

In all cases:


project/Dependencies.scala
```diff
@@ -102,6 +102,10 @@ object Dependencies {
   val scalaCache = "com.github.cb372" %% "scalacache-caffeine" % "1.0.0-M6"
   val swaggerUi = "org.webjars" % "swagger-ui" % "4.10.3"
 
+  val azureResourceManager = "com.azure.resourcemanager" % "azure-resourcemanager" % "2.16.0"
```
rtitle, line 105: Hm.. should this depend on CRL instead to get the Azure dependencies from one place?

Have we considered using CRL from Leo directly and not adding a wrapper in wb-libs? That's what I did in Sam: https://github.com/broadinstitute/sam/pull/737

Qi77Qi: Maybe in the future if we have a reason to. Right now, Leo isn't really creating much resources (the only thing is hybrid connection, which we decided a while back that we'll just have it in Leo becuz deleting relay namespace will automatically delete hybrid connections)

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureVmService.scala
```diff
@@ -0,0 +1,32 @@
...
+import cats.effect.{Async, Resource}
+import cats.mtl.Ask
+import com.azure.identity.ClientSecretCredentialBuilder
+import com.azure.resourcemanager.compute.models.VirtualMachine
+import com.azure.resourcemanager.resources.fluentcore.model.Accepted
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.typelevel.log4cats.StructuredLogger
+
+trait AzureVmService[F[_]] {
+  def getAzureVm(name: String, cloudContext: AzureCloudContext)(implicit
```
jdcanas: These are already defined as `RuntimeName` in leo and exposed in core, so it may be worth using that here so we can get some more type safety...

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureVmServiceInterp.scala
```diff
@@ -0,0 +1,72 @@
...
+
+      fa = F
+        .delay(
+          azureComputeManager
+            .virtualMachines()
+            .getByResourceGroup(cloudContext.managedResourceGroupName.value, name)
+        )
+        .map(Option(_))
+        .handleErrorWith {
+          case e: ManagementException if e.getValue.getCode().equals("ResourceNotFound") => F.pure(none[VirtualMachine])
```
jdcanas, line 36: Like that these are idempotent

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureVmServiceInterp.scala
```diff
@@ -0,0 +1,72 @@
...
+  def deleteAzureVm(name: String, cloudContext: AzureCloudContext, forceDeletion: Boolean)(implicit
+    ev: Ask[F, TraceId]
+  ): F[Option[Accepted[Void]]] =
+    for {
+      azureComputeManager <- buildComputeManager(cloudContext)
+      fa = F
+        .delay(
+          azureComputeManager
+            .virtualMachines()
+            .beginDeleteByResourceGroup(cloudContext.managedResourceGroupName.value, name, forceDeletion)
```
jdcanas: what exactly does `beginDeleteByResuorceGroup` do? May be worth including a comment on how it differs from `deleteByResourceGroup`.

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/package.scala
```diff
@@ -0,0 +1,6 @@
+package org.broadinstitute.dsde.workbench
+
+package object azure {
+  val CLOUD_PLATFORM_SCOPE = "https://www.googleapis.com/auth/cloud-platform"
```
jdcanas: huh... strikes me as odd that this google link is in the azure package object. I don't see it used anywhere with a simple ctrl+f, what is its purpose?

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/package.scala
```diff
@@ -40,4 +46,51 @@ package object util2 {
       .through(fs2.text.utf8Decode)
       .through(_root_.io.circe.fs2.stringParser(AsyncParser.SingleValue))
       .through(decoder)
+
+  implicit private val loggableGoogleCallEncoder: Encoder[LoggableCloudCall] = Encoder.forProduct2(
+    "response",
+    "result"
+  )(x => LoggableCloudCall.unapply(x).get)
+
+  def withLogging[F[_]: Temporal, A](fa: F[A],
```
jdcanas, line 55: what exactly is `Temporal` ?

jdcanas (approved):
looks good overall, just a few questions/nits

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureVmService.scala
Qi77Qi: `RuntmeName` is more of a leo specific data model. We use `InstanceName` in `google2`, maybe I'll just move that to `util2` and use it for both google and azure

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureVmServiceInterp.scala
Qi77Qi: I didn't test for real. But judging from the type signature, `deleteByResourceGroup` is synchronous vs `beginDeleteByResuorceGroup` will be async...and you can call function to wait on `Accepted[Void]` returned by  `beginDeleteByResuorceGroup` if caller decides too

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/package.scala
Qi77Qi: oh good catch...definitely copy paste error

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/package.scala
Qi77Qi: it's just a typeclass that gives you time related functions like `Temporal[F].timed(some F)`...At call site, since most times we already have `Async[F]` in scope, we get `Temporal` automatically. `Temporal` is just a bit more precise constrain for `withLogging`

PR #1039, opened 2022-07-15
[WOR-257] Preserve BQ job location when fetching query status and results

marctalbott:
Ticket: [WOR-257](https://broadworkbench.atlassian.net/browse/WOR-257)
* BigQuery jobs run in the same location that the dataset is located in ([source](https://cloud.google.com/bigquery/docs/locations#specify_locations)). In order to fetch a job's status or results, you *must* specify the job location if it is not in the `US` or `EU` multi-region or if it is in a single region (like `us-central1` or `northamerica-northeast1`)([source](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/get#query-parameters)). 
* This has been the source of the bug tracked in the linked ticket. While the Broad billing export dataset works fine because it is in the `US` multi-region, our per-workflow spend reporting has not been working for users with billing export datasets in other GCP regions. While the jobs to fetch spend data can be created and run correctly, we then fail to load the job because we do not specify the location the job is in. This PR specifies the job location when fetching the job.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Workbench utility libraries, built for 2.13. You can find the full list of packa
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery. 
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.21-c7d6911"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.21-TRAVIS-REPLACE-ME"`
 
 To depend on the `MockGoogle*` classes, additionally depend on:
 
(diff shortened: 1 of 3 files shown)


ungwudik (approved):
Is there a way to know if this change can somehow break other services, before merging?

Also, I imagine you tested (or plan to test) against the upcoming Rawls changes/PR.

aherbst-broad (approved):

marctalbott:
> Is there a way to know if this change can somehow break other services, before merging?
> 
> Also, I imagine you tested (or plan to test) against the upcoming Rawls changes/PR.

I have confirmed that these changes work in Rawls by running a local Rawls with a locally published version of this library and confirming that submission cost querying works. Also confirmed with a BQ dataset in `australia-southeast1` that Rawls can now find the BQ job when it is not located in the `US` multi-region.

I'm not all that concerned about this change somehow breaking other services. I've tested that if the location in the `JobReference` is null, the query will still work as long as the BQ job is located in the `US` multi-region. In other words, even if an upstream service is passing in a `JobReference` as a parameter that does not have the location set, `getQueryStatus` and `getQueryResult` will still work

PR #1052, opened 2022-07-25
[IA-3514] add azure storage module

jdcanas:
Tested all methods with the manual test class, via this storage acct https://portal.azure.com/#@broadinstitute.org/resource/subscriptions/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/resourceGroups/CARresourcegroup/providers/Microsoft.Storage/storageAccounts/jcstorageacct/containersList

You can get the connection string via `access keys` tab, and theres a container created that you can interact with. You can also create your own container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Workbench utility libraries, built for 2.13. You can find the full list of packa
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-azure" % "0.1-91d3580"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-azure" % "0.1-TRAVIS-REPLACE-ME"`
 
 [Changelog](azure/CHANGELOG.md)
 
--- a/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
+++ b/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
@@ -1,0 +1,128 @@
+package org.broadinstitute.dsde.workbench.azure
+
+import java.io.{InputStream, OutputStream}
+
+import cats.effect.Async
+import com.azure.storage.blob.models.{BlobItem, ListBlobsOptions}
+import com.azure.storage.blob.{BlobClient, BlobContainerClient, BlobContainerClientBuilder}
+import cats.implicits._
+import java.time.Duration
+
+import org.broadinstitute.dsde.workbench.util2.tracedLogging
+import fs2.{Pipe, Stream}
+import java.nio.file.Path
+
+import cats.mtl.Ask
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+import scala.jdk.CollectionConverters._
+import org.typelevel.log4cats.StructuredLogger
+
+class AzureStorageInterp[F[_]](config: AzureStorageConfig)(implicit val F: Async[F], logger: StructuredLogger[F])
+    extends AzureStorageService[F] {
+
+  override def listObjects(connectionString: ConnectionString,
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -15,7 +15,7 @@ Workbench utility libraries, built for 2.13. You can find the full list of packa
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-azure" % "0.1-91d3580"`
+Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-azure" % "0.2-TRAVIS-REPLACE-ME"`
```
Qi77Qi: I don't thikn we need to bump version since we're not breaking any existing functions

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
```diff
@@ -0,0 +1,128 @@
...
+  // See below article for more info on the code that can be used for uploading as a stream
+  // https://docs.microsoft.com/en-us/java/api/overview/azure/storage-blob-readme?view=azure-java-stable#upload-data-to-a-blob
+  override def uploadBlob(connectionString: ConnectionString, containerName: ContainerName, blobName: BlobName)(implicit
+    ev: Ask[F, TraceId]
+  ): Pipe[F, Byte, Unit] = {
+    val outputStream = for {
+      blobClient <- buildBlobClient(connectionString, containerName, blobName)
+      outputStream <- F.delay(blobClient.getBlockBlobClient.getBlobOutputStream)
+      // This is a subclass of OutputStream, but the scala code is not happy without the explicit conversion since its a java subclass
+    } yield outputStream.asInstanceOf[OutputStream]
```
Qi77Qi: you can do `outputStream: OutputStream`, which is a bit safer then `asInstanceOf`....the former will complain if the type doesn't actually match at compile time; vs if you do `asInstanceOf`, it'll throw exception at runtime if it doesn't match

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  override def getBlob(connectionString: ConnectionString, containerName: ContainerName, blobName: BlobName)(implicit
+    ev: Ask[F, TraceId]
+  ): Stream[F, Byte] =
+    for {
+      client <- Stream.eval(
+        buildBlobClient(connectionString, containerName, blobName)
+      )
+      is <- fs2.io.readInputStream(
+        F.delay(client.openInputStream().asInstanceOf[InputStream]),
```
Qi77Qi: same thing, use `:` instead of `asInstanceOf`

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala
```diff
@@ -0,0 +1,59 @@
...
+  def fromAzureAppRegistrationConfig[F[_]: Async: StructuredLogger](
+    azureStorageConfig: AzureStorageConfig
+  ): Resource[F, AzureStorageService[F]] =
+    Resource.eval(Async[F].pure(new AzureStorageInterp(azureStorageConfig)))
+}
+
+case class AzureStorageConfig(listTimeout: FiniteDuration)
+
+// See this article for more information on connection strings: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-java?tabs=powershell%2Cenvironment-variable-windows#get-the-connection-string
+case class ConnectionString(value: String) extends AnyVal
```
Qi77Qi: nit: add `final`

Qi77Qi:
looks great in general!

Qi77Qi (approved):

PR #1053, opened 2022-08-01
[IA-3514] sas token used over conn string

jdcanas:


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala
```diff
@@ -4,53 +4,54 @@ import java.nio.file.Path
...
+            .sasToken(azureStorageConfig.sasToken)
+            .endpoint(azureStorageConfig.endpointUrl)
+            .buildClient()
+        )
+      )
+    } yield new AzureStorageInterp(azureStorageConfig, storageServiceClient)
 }
 
-final case class AzureStorageConfig(listTimeout: FiniteDuration)
+final case class AzureStorageConfig(listTimeout: FiniteDuration, sasToken: String, endpointUrl: String)
```
Qi77Qi: can we use one custom type for either sasToken or endpointUrl, two `String` types next to each other makes me nervous

Qi77Qi (approved):

PR #1054, opened 2022-08-04
[SU-158] Support terminal submission email notifications 

MatthewBemis:
Companion PRs:

https://github.com/broadinstitute/rawls/pull/1971
https://github.com/broadinstitute/thurloe/pull/83
https://github.com/broadinstitute/firecloud-develop/pull/3057

**PR checklist**

In all cases:

### DIFF
--- a/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
+++ b/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
@@ -119,6 +119,20 @@ object Notifications {
     override val description = "Workspace changed"
   })
 
+  case class TerminalSubmissionNotification(recipientUserId: WorkbenchUserId,
+                                            workspaceName: WorkspaceName,
+                                            submissionId: String,
+                                            comment: String,
+                                            workflowStatuses: Map[String, Int],
+                                            workflowConfiguration: String,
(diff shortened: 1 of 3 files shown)


notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
```diff
@@ -119,6 +119,48 @@ object Notifications {
     override val description = "Workspace changed"
   })
 
+  case class SuccessfulSubmissionNotification(recipientUserId: WorkbenchUserId,
```
MatthewBemis: Having 3 very similar but different notifications might not seem great from a code perspective but it makes templating on the SendGrid side much easier because we don't have to deal with conditional templates, which aren't great, and we're still on legacy templates (which is even less great).

I'm also not convinced that these 3 notifications will always look so similar.

nawatts (approved):

notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
Qi77Qi: can you add `final` just so users won't keep extending the case class (which is unlikely what they want)

Qi77Qi (approved):
small nit

PR #1070, opened 2022-08-08
[IA-3511] modify azure apis

jdcanas:
Wanted `deleteBlob` to be consistent with google trait signature so the consuming welder interface can have the same signature across azure/google storage interps

Shouldn't require anything more than patch revision since azure API is one that was changed and it is not in use yet.

### DIFF
--- a/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
+++ b/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
@@ -95,16 +95,14 @@ class AzureStorageInterp[F[_]](config: AzureStorageConfig, blobServiceClient: Bl
       client <- buildBlobClient(containerName, blobName)
       resp <- tracedLogging(
         F.delay(
-          RemoveObjectResult(
-            202 == client
-              .deleteWithResponse(
-                null,
-                null,
-                Duration.ofMillis(config.generalTimeout.toMillis),
-                new Context("traceId", traceId)
-              )
-              .getStatusCode
-          )
+          client
+            .deleteWithResponse(
+              null,
+              null,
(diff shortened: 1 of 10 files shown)


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageInterp.scala
```diff
@@ -88,14 +89,31 @@ class AzureStorageInterp[F[_]](config: AzureStorageConfig, blobServiceClient: Bl
...
+              )
+              .getStatusCode
+          )
+        ).handleErrorWith {
+          case e: com.azure.storage.blob.models.BlobStorageException if e.getStatusCode == 404 =>
+            F.pure(RemoveObjectResult(false))
+          case e => F.raiseError(e)
+        },
+        s"com.azure.storage.blob.BlobClient($containerName, $blobName).deleteWithResponse(null, null, ${Duration
+            .ofMillis(config.generalTimeout.toMillis)}, Context('traceId', $traceId))"
```
Qi77Qi, line 113: let's put the cloud call in a separate `F.delay()`, sth like
```
_ <- tracedLogging(F.delay(client.delete()),
                         s"com.azure.storage.blob.BlobClient($containerName, $blobName).deleteWithResponse()")
                         .map(res => if(res.getStatusCdoe==202) ??? else ???)
                         .handleErrorWith {
          case e: com.azure.storage.blob.models.BlobStorageException if e.getStatusCode == 404 =>
            F.pure(RemoveObjectResult(false))
          case e => F.raiseError(e)
        }
```

Assuming `res.getStatusCdoe` doesn't actually have side effect. If it does, we should wrap that in a separate `F.delay` as well

Qi77Qi (approved):
Can you add ticket number for the PR

PR #1074, opened 2022-08-17
IA-3664 Make Azure client support multiple containers properly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3664

**PR checklist**

In all cases:

### DIFF
--- a/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala
+++ b/azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala
@@ -1,10 +1,9 @@
 package org.broadinstitute.dsde.workbench.azure
 
 import java.nio.file.Path
-
 import cats.effect.{Async, Resource}
 import cats.mtl.Ask
-import com.azure.storage.blob.{BlobServiceClient, BlobServiceClientBuilder}
+import com.azure.storage.blob.{BlobContainerClientBuilder, BlobServiceClient, BlobServiceClientBuilder}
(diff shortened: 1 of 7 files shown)


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureStorageService.scala
```diff
@@ -37,26 +36,29 @@ trait AzureStorageService[F[_]] {
 }
 
 object AzureStorageService {
+  // Note SAS token has expiration time. If the token expires, the client will not work
   def fromSasToken[F[_]: Async: StructuredLogger](
-    azureStorageConfig: AzureStorageConfig
+    azureStorageConfig: AzureStorageConfig,
+    azureStorageConfigs: Map[ContainerName, ContainerAuthConfig]
```
jdcanas, line 42: hmm... does this help? It just means the caller has to know all containers it needs to interact with at initialization... I would think we would need to defer the specification of a token/container until the method calls themselves, and the `AzureStorageService` would determine if it needs to build a new `BlobContainerClientBuilder` based on the params. That way, it could cache `BlobContainerClient`s and new containers could be interacted with on the fly

Qi77Qi: this should suffice for our use case in welder...welder is where the [rotation](https://github.com/DataBiosphere/welder/blob/master/server/src/main/scala/org/broadinstitute/dsp/workbench/welder/server/package.scala#L32-L34) will happen, I don't know if we wanna bring all that into the library

jdcanas (approved):

PR #1095, opened 2022-09-06
Revert swagger-ui to 4.11.1

ehigham:
Latest versions are breaking rawls swagger page. Downgrade.

**PR checklist**

In all cases:

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -41,7 +41,10 @@ pullRequests.frequency = "15 days"
 # Each pattern must have `groupId`, `version` and optional `artifactId`.
 # Defaults to empty `[]` which mean Scala Steward will update all dependencies.
 # the following example will allow to update foo when version is 1.1.x
-updates.pin  = [ { groupId = "org.broadinstitute.dsde", artifactId="rawls-model", version = "0.1-2356e282" } ]
+updates.pin = [
+  { groupId = "org.broadinstitute.dsde", artifactId="rawls-model", version = "0.1-2356e282" },
+  { groupId = "org.webjars", artifactId="swagger-ui", version="4.11.1" }
+]
 
 # The dependencies which match the given pattern are NOT updated.
 #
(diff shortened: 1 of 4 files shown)


rtitle (approved):
Maybe change the hash to `TRAVIS-REPLACE-ME` [here](https://github.com/broadinstitute/workbench-libs/blob/develop/README.md?plain=1#L137) and [here](https://github.com/broadinstitute/workbench-libs/blob/develop/oauth2/CHANGELOG.md?plain=1#L18) so the hash gets updated in the readme's.

Thanks for fixing

cahrens (approved):
I wonder if it would be worth adding a comment that 4.14.0 caused an issue, so upgrades should be tested.

rtitle:
Comment sounds good.. it’s also possible to exclude some library updates in .scala-steward.conf.

cahrens:
> Comment sounds good.. it’s also possible to exclude some library updates in .scala-steward.conf.

I considered that, but I hope that some later release of Swagger-ui will actually work, and I don't think we want to exclude upgrades on it permanently. But I could see using the exclusion with a comment explaining why.

Qi77Qi:
do we have to revert the upgrade in wb-libs now? I mean rawls could just revert the wb-lib upgrade for now (if there's a reason for rawls to upgrade, then I can see why we'd have to revert the upgrade here)...also, do we know it's bad swagger-ui version or it's sth in Rawls need to be updated for the upgrade?

rtitle:
I think the reason the upgrade broke is we check in a version of swagger-ui `index.html` which is not compatible with 4.14: https://github.com/broadinstitute/workbench-libs/blob/develop/oauth2/src/main/resources/swagger/index.html 😞 

I think this was done to add some B2C-related customizations, but there may be a better way to make this tolerant to upgrades. For now though it's probably best to peg the swagger-ui version in wb-libs.

Qi77Qi (approved):

PR #1116, opened 2022-10-03
Initial AKS client, support for authentication

rtitle:
Draft for now, cc @Qi77Qi 

Added AKS service/interpreter, with methods for authentication. I think this will plug into the Leo helm client.

Gonna trying making a "hello world" helm deploy from Leo next..

**PR checklist**

In all cases:

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -1,0 +1,47 @@
+# Testing Locally in sbt console
+
+This is an example on how to test Azure methods locally. In this example, we are testing the
+AzureContainerServiceInterp.getClusterCredentials method.
+
+Start sbt console:
+```
+sbt "project workbenchAzure" console
+```
+
+Set up imports. Type `:paste` to get a copy-paste editor:
+```scala
+import cats.effect.IO
+import org.broadinstitute.dsde.workbench.azure._
+import org.broadinstitute.dsde.workbench.azure.AzureContainerService._
+import org.typelevel.log4cats._
+import org.typelevel.log4cats.slf4j._
+import org.broadinstitute.dsde.workbench.model.TraceId
+import cats.mtl.Ask
+import java.util.UUID
+import cats.effect.unsafe.implicits.global
+
+implicit val logger: SelfAwareStructuredLogger[IO] = LoggerFactory[IO].getLogger
+implicit val traceId = Ask.const[IO, TraceId](TraceId(UUID.randomUUID()))
+
+// Replace with your tenant/subscription/MRG
+val cloudContext = AzureCloudContext(
+  TenantId("0cb7a640-45a2-4ed6-be9f-63519f86e04b"), 
+  SubscriptionId("3efc5bdf-be0e-44e7-b1d7-c08931e3c16c"), 
+  ManagedResourceGroupName("mrg-rt-aks-end-to-end-20220930090032"))
+// Get from vault
+val config = AzureAppRegistrationConfig(
+  ClientId("<client>>"), 
+  ClientSecret("<secret>"), 
+  ManagedAppTenantId("<tenant>"))
+val clusterName = AKSClusterName("lz961d62cced17ccac037ce22")
+```
+
+Create AzureContainerService resource.
+```scala
+val service = AzureContainerService.fromAzureAppRegistrationConfig[IO](config)
+```
+
+Call methods on the AzureContainerService object.
+```scala
+service.use(x => x.getClusterCredentials(clusterName, cloudContext)).unsafeRunSync
+```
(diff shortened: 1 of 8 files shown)


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureContainerServiceInterp.scala
```diff
@@ -0,0 +1,59 @@
...
+            .manager()
+            .serviceClient()
+            .getManagedClusters()
+            .listClusterMonitoringUserCredentials(cloudContext.managedResourceGroupName.value, name.value)
+        ),
+        s"com.azure.resourcemanager.containerservice.fluent.ManagedClustersClient.listClusterMonitoringUserCredentials(${cloudContext.managedResourceGroupName.value}, ${name})"
+      )
+      kubeConfig <- F.fromOption(resp.kubeconfigs().asScala.headOption, new WorkbenchException("No AKS credential"))
+      // Parse the kubeconfig file
+      kubeConfig <- F.delay(
```
Qi77Qi: can we name this variable sth different....it's the same as previous variable :)

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/models.scala
```diff
@@ -37,3 +37,9 @@ final case class AzureAppRegistrationConfig(clientId: ClientId,
 final case class RelayNamespace(value: String) extends AnyVal
 final case class RelayHybridConnectionName(value: String) extends AnyVal
 final case class PrimaryKey(value: String) extends AnyVal
+
+final case class AKSClusterName(value: String) extends AnyVal
```
Qi77Qi, line 41: wonder if it would make sense to re-use some of these from google's definition (merely a question, I don't know the answer :) )

Qi77Qi (approved):

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/models.scala
rtitle: https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala#L80 is similar I guess, but I don't think we have google models for server/token/cert (since it comes from the google object). Also `workbenchAzure` doesn't depend on `workbenchGoogle2` making sharing harder.

Maybe I will rename these though, don't think we need to prefix each type with `AKS`

Qi77Qi: I moved some things that could be reused to https://github.com/broadinstitute/workbench-libs/tree/develop/util2/src/main/scala/org/broadinstitute/dsde/workbench/util2 previously

rtitle: Cool, I can look at that

azure/README.md
```diff
@@ -0,0 +1,47 @@
+# Testing Locally in sbt console
```
rtitle, line 1: @Qi77Qi fyi added a README with some manual test steps

Qi77Qi: ah awesome!

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureContainerServiceInterp.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.typelevel.log4cats.StructuredLogger
+
+import java.io.{ByteArrayInputStream, InputStreamReader}
+import scala.jdk.CollectionConverters._
+
+class AzureContainerServiceInterp[F[_]](clientSecretCredential: ClientSecretCredential)(implicit
+  val F: Async[F],
+  logger: StructuredLogger[F]
+) extends AzureContainerService[F] {
+  override def getClusterCredentials(name: AKSClusterName, cloudContext: AzureCloudContext)(implicit
```
rtitle, line 39: Hm I tried this with the `scala-helm-sdk` against a landing zone-created cluster. The fields look right but I got on helm install:
```
err is %!(EXTRA *errors.withStack=rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource: services "rel-1-hello" is forbidden: User "clusterMonitoringUser" cannot get resource "services" in API group "" in the namespace "ns-1", string=
```

Seems this credential doesn't have permissions to helm install, will investigate more/read docs.

rtitle: Fixed, helm install a hello world app is working now :)

rtitle:
Going to merge this as it's working from Leo, can iterate after.

PR #1206, opened 2023-01-10
[WOR-708] Add model for billing project invitation email

MatthewBemis:
**PR checklist**

In all cases:

### DIFF
--- a/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
+++ b/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
@@ -114,7 +114,7 @@ object Notifications {
 
   case class BillingProjectInvitedNotification(recipientUserEmail: WorkbenchEmail,
                                           requesterId: WorkbenchUserId,
-                                          billingProjectName: WorkspaceName
+                                          billingProjectName: String
                                          ) extends Notification
 
   val BillingProjectInvitedNotificationType = register(new NotificationType[BillingProjectInvitedNotification] {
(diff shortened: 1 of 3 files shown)


notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
```diff
@@ -112,6 +112,17 @@ object Notifications {
     override val alwaysOn = true
   })
 
+  case class BillingProjectInvitedNotification(recipientUserEmail: WorkbenchEmail,
```
MatthewBemis, line 115: This model will be used by Rawls and Thurloe. See Thurloe example [here](https://github.com/broadinstitute/thurloe/pull/126/files). Rawls coming soon.

nawatts (approved):

cahrens (approved):

notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
```diff
@@ -112,6 +112,17 @@ object Notifications {
...
+                                               requesterId: WorkbenchUserId,
+                                               billingProjectName: String
```
Qi77Qi, line 117: does `GoogleProject` type exist in this project...would be nice to not use `String`

Qi77Qi (approved):

n8landolt (approved):

notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
MatthewBemis: `GoogleProject` is not quite right because they could be Azure billing projects too. What I really want is something like:

`case class BillingProjectName(value: String)`

But that already exists in Rawls so there would be some work to harmonize. I think it should be pushed into workbench-libs at some point because it's useful for many services.

Qi77Qi: ah that'd be great! having that type in wb-libs...

would be nice to have comments here as well what this String is

PR #1267, opened 2023-02-24
Add (optional) opentelemetry to Google API requests

tlangs:
Here's an idea I'm workshopping to add metrics to google api requests. Given the recent incident with the Google Directory API, it might be good to actually have metrics (on our end) of which apps are doing the talking to the api. This would be tested out in Sam first, but any app could theoretically take advantage of it.

The new `TelemeterizedHttpGoogleDirectoryDAO` would take in app-specific opentelemetry config as an implicit class input, so each app would have its own counter.

**PR checklist**

In all cases:


google/src/main/scala/org/broadinstitute/dsde/workbench/google/TelemeterizedHttpGoogleDirectoryDAO.scala
```diff
@@ -0,0 +1,96 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+class TelemeterizedHttpGoogleDirectoryDAO(appName: String,
+                                        googleCredentialMode: GoogleCredentialMode,
+                                        workbenchMetricBaseName: String
+                                       )(implicit
+                                         system: ActorSystem,
+                                         executionContext: ExecutionContext,
+                                         implicit val openTelemetry: OpenTelemetryMetrics[IO]
```
dvoet: 1. you don't need the `implicit` keyword here, the `implicit` at the beginning of the parameter list applies to all parameters in the list
2.  I would just stay away from creating your own implicits, they usually lead to pain
3. could you add `openTelemetry` as an option to `HttpGoogleDirectoryDAO` instead of creating this delegate pattern?

dvoet (commented):
You should note that this is technically not capturing metrics on google api requests. It is close but one step removed from reality. It misses retries and any time one of these functions requires more than one google api call e.g. `createGroup` and `listGroupMembers`.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/TelemeterizedHttpGoogleDirectoryDAO.scala
tlangs: whoops copy-pasta. 

I like the idea of adding it as an option to the `HttpGoogleDirectoryDAO` instead. The fewer classes we have in the inheritance chain, the better.

tlangs: Updated it so that there's not yet another class

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -100,6 +105,12 @@ class HttpGoogleDirectoryDAO(appName: String,
 
   private lazy val directory =
     new Directory.Builder(httpTransport, jsonFactory, googleCredential).setApplicationName(appName).build()
+  private def sendMetrics[A](metricName: String): PartialFunction[A, Unit] = {
+    case Success(_) =>
+      openTelemetry.foreach(ot => ot.incrementCounter(s"${metricName}Success", tags = openTelemetryTags))
+    case Failure(_) =>
+      openTelemetry.foreach(ot => ot.incrementCounter(s"${metricName}Failure", tags = openTelemetryTags))
```
dvoet, line 112: `ot.incrementCounter()` produces an `IO` which is never actually executed. You need `unsafeRunSync()` or the like.

dvoet (changes requested):
This is still not instrumenting the actual google calls. To do so it needs to count each call to `executeGoogleRequest` which are currently inside retry blocks. So this is missing any retries.

However in looking closer, I see this is [already instrumented](https://github.com/broadinstitute/workbench-libs/blob/develop/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleUtilities.scala#L214). Maybe you want to see how to hook into those metrics instead?

google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
tlangs: Oh, good catch!

tlangs:
According to @rtitle, the current instrumentation should "probably be scrapped and redone with more modern opencensus/prometheus tools". 

Apparently, the current stuff is for statsd/HostedGraphite and not pushed to grafana.

rtitle:
Haven't dug into this PR but yeah, the existing `GoogleCounters` via `workbench-metrics` module (https://github.com/broadinstitute/workbench-libs/blob/develop/google/src/main/scala/org/broadinstitute/dsde/workbench/metrics/GoogleInstrumented.scala) was written long ago for statsd/Hosted Graphite and is probably not hooked up to anything anymore. Maybe there is some code to borrow, but we should probably re-do metrics with current tools (OpenCensus/prometheus).

PR #1336, opened 2023-04-27
ID-586 signed gs urls

tlangs:
**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
@@ -7,7 +7,7 @@ import cats.effect._
 import cats.effect.std.Semaphore
 import cats.syntax.all._
 import com.google.auth.Credentials
-import com.google.auth.oauth2.{AccessToken, GoogleCredentials}
+import com.google.auth.oauth2.{AccessToken, GoogleCredentials, ServiceAccountCredentials}
 import com.google.cloud.storage.BucketInfo.LifecycleRule
 import com.google.cloud.storage.{Acl, Blob, BlobId, BucketInfo, StorageOptions}
 import com.google.cloud.{Identity, Policy, Role}
@@ -29,7 +29,9 @@ import org.broadinstitute.dsde.workbench.model.TraceId
 import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName, GoogleProject, IamPermission}
 import org.broadinstitute.dsde.workbench.util2.RemoveObjectResult
 
+import java.net.URL
 import scala.collection.convert.ImplicitConversions._
+import scala.concurrent.duration.{HOURS, TimeUnit}
 import scala.language.higherKinds
 
 /**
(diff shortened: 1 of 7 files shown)


Ghost-in-a-Jar (approved):

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
```diff
@@ -186,6 +188,23 @@ trait GoogleStorageService[F[_]] {
...
+   * @param traceId uuid for tracing a unique call flow in logging
+   * @return Signed URL
+   */
+  def getSignedBlobUrl(bucketName: GcsBucketName,
+                       blobName: GcsBlobName,
+                       signingCredentials: ServiceAccountCredentials,
+                       traceId: Option[TraceId] = None,
+                       retryConfig: RetryConfig = standardGoogleRetryConfig,
+                       expirationTime: Long = 1,
+                       expirationTimeUnit: TimeUnit = HOURS
```
gpolumbo-broad, line 208: Missing a couple params in the scaladoc.  Also - do we want to _require_ an expiration durations?  I think you can also sign URLs that will expire after being accessed a certain number of times for example and I think there are other rules too.  Maybe we'll have other methods for those.  But I'm not sure if we ant `expirationTime: Long` or `expirationTime: Option[Long]`

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreterSpec.scala
```diff
@@ -87,6 +90,27 @@ class GoogleStorageInterpreterSpec extends AsyncFlatSpec with Matchers with Work
     }
   }
 
+  "ioStorage getSignedBlobUrl" should "be able to get a signed blob URL" in ioAssertion {
```
gpolumbo-broad, line 93: What about other failure scenarios like if the bucket or blob doesn't exist?  Or if the signing credentials don't work or are invalid some how?

gpolumbo-broad (changes requested):

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreterSpec.scala
tlangs: I don't think we have a way to test that credentials are invalid. The mock here just assumes that credentials are valid, as long as the credential exists. I think we can trust Google that they won't provide access to a URL that the service account doesn't have access to.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageService.scala
tlangs: I've replicated what Martha does here as the defaults. I don't think we should change the functionality now, and rather focus on replicating the existing functionality. Certainly something we could revisit, but I don't think its something to do now.

google2/src/test/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreterSpec.scala
tlangs: I just tested the deployed BEE code, and Google will happily sign a URL for a blob that doesn't exist. Once that URL is clicked on, the user gets an error. How much do we want workbench-libs to do?

gpolumbo-broad: Great question, I would say to do what Martha does, so it will probably just happily sign a blob that doesn't exist.  I would see if you can write a test asserting this too, if anything just so that if anybody looks at this in the future, they can say, "huh, well, I guess that's the way it's supposed to work because there's a test for it"  You can comment saying we think it's weird and why we left it in.  But these sorts of little decisions are usually worth asserting on and/or commenting.

gpolumbo-broad (approved):

PR #1347, opened 2023-05-11
[WOR-920] Pass a string map instead of AttributeMap

aherbst-broad:
AttributeMap is not serializing properly, pass a Map[String, String] instead. 

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
@@ -282,7 +282,7 @@ trait Rawls extends RestClient with LazyLogging {
               destName: String,
               authDomain: Set[String] = Set.empty,
               copyFilesWithPrefix: Option[String] = None,
-              attributes: AttributeMap = Map.empty
+              attributes: Map[String, String] = Map.empty
     )(implicit token: AuthToken): Unit = {
       logger.info(
         s"Cloning workspace: $sourceNamespace/$sourceName into $destNamespace/$destName authDomain: $authDomain, copyFilesWithPrefix: $copyFilesWithPrefix"
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
@@ -259,7 +259,7 @@ trait Rawls extends RestClient with LazyLogging {
     def create(namespace: String,
                name: String,
                authDomain: Set[String] = Set.empty,
-               attributes: AttributeMap = Map.empty
+               attributes: Map[String, String] = Map.empty
     )(implicit
       token: AuthToken
     ): Unit = {


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -259,7 +259,7 @@ trait Rawls extends RestClient with LazyLogging {
     def create(namespace: String,
                name: String,
                authDomain: Set[String] = Set.empty,
-               attributes: AttributeMap = Map.empty
+               attributes: Map[String, String] = Map.empty
```
aherbst-broad, line 262: My scala-fu is failing me. The issue is that passing an AttributeMap was failing in the call to rawls as it unmarshals incorrectly. Changing this to a Map[String, String] does the right thing, not sure if there's a cleaner way to express this.

cahrens (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
marctalbott: Not sure what else you've tried, but `AttributeMap` is a type alias for `Map[AttributeName, Attribute]`, maybe ignoring the alias and passing `Map[AttributeName, Attribute]` would work?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -259,7 +259,7 @@ trait Rawls extends RestClient with LazyLogging {
     def create(namespace: String,
```
marctalbott, line 259: do other automated tests pass with this version? I'm not familiar with how much the attributes are used in our existing automated tests

cahrens: I just added it myself, so nobody else is using it.

But @aherbst-broad , I did add it to the clone signature as well! Could you update that too?

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
cahrens: What we had before was that I was trying to call this with the following code, which would fail. I'll post the error message next.

```
Rawls.workspaces.create(
        projectName,
        workspaceName,
        Set.empty,
        Map(AttributeName.withDefaultNS("disableAutomaticAppCreation") -> AttributeBoolean(true)))

Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:262]: {"causes":[],"message":"Invalid input: AttributeName(default,disableAutomaticAppCreation). Input may only contain alphanumeric characters, underscores, and dashes.","source":"rawls","stackTrace":[],"statusCode":400}

Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:262]: But looking at this error again, the cause of the issue is `AttributeName`. So perhaps I'm just not constructing that correctly!

Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:262]: Though what I'm doing seems legit based on uses in Rawls. For example:

`Map(AttributeName.withDefaultNS("samples") -> AttributeEntityReferenceList(Seq(sample.toReference)))`

Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:262]: Since we are just calling to immediately call the Rawls REST API, passing the argument in a way closer to the way we do from terra-ui seems fine, but the wrinkle is that the boolean value is getting converted to a string. I agree that this change isn't the best possible one, but it does allow us to move forward on the test.

File: serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala (hunk 2 above)
Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:259]: Clone method-- https://github.com/broadinstitute/workbench-libs/pull/1347/files#diff-78009bb6f538dc1b945e9a05b1b1eb802879133cf72b1b5441cbc97b31086a8cR285

File: serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala (hunk 1 above)
Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:262]: With this proposed change, we will call the method like this:

```
Rawls.workspaces.create(
        projectName,
        workspaceName,
        Set.empty,
        Map("disableAutomaticAppCreation" -> "true")
      )

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
aherbst-broad: @cahrens done.

marctalbott (approved):

PR #1399, opened 2023-06-16
Added generator function for creating backoff intervals and increased…

sjkobori:
… the default backoff interval length

Ticket: https://broadworkbench.atlassian.net/browse/ID-660

Adding longer backoff interval to attempt to not give up on retrying request too soon.

**PR checklist**

In all cases:

### DIFF
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
@@ -104,7 +104,7 @@ trait Retry {
                                                   jitterValue: Int = 1000
   ): Seq[FiniteDuration] = {
     val intervals = Seq.iterate(startingValue, elements)(_ * multiplyBy).map(_ milliseconds)
-    if (withJitter) intervals else intervals.map(i => addJitter(i, jitterValue milliseconds))
+    if (withJitter) intervals.map(i => addJitter(i, jitterValue milliseconds)) else intervals
   }
 
   // 1000, 2000, ...., 64000 milliseconds
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
@@ -100,11 +100,15 @@ trait Retry {
   protected def createExponentialBackOffIntervals(startingValue: Int,
                                                   multiplyBy: Int,
                                                   elements: Int,
-                                                  withJitter: Boolean = true,
-                                                  jitterValue: Int = 1000
+                                                  jitterValue: Option[Int] = Some(1000)
   ): Seq[FiniteDuration] = {
     val intervals = Seq.iterate(startingValue, elements)(_ * multiplyBy).map(_ milliseconds)
-    if (withJitter) intervals.map(i => addJitter(i, jitterValue milliseconds)) else intervals
+    jitterValue match {
+      case Some(value) =>
+        intervals.map(i => addJitter(i, value milliseconds))
+      case None =>
+        intervals
+    }
   }
 
   // 1000, 2000, ...., 64000 milliseconds
--- a/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
+++ b/util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
@@ -96,15 +96,20 @@ trait Retry {
 
   private val allBackoffIntervals = Seq(100 milliseconds, 1 second, 3 seconds, 60 seconds)
 
-  //starting value in milliseconds, multiply by multiples of this each step, how many steps/elements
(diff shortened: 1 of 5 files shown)


util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
```diff
@@ -94,19 +94,23 @@ trait Retry {
...
+
+  // starting value in milliseconds, multiply by multiples of this each step, how many steps/elements
+  protected def createExponentialBackOffIntervals(startingValue: Int,
+                                                  multiplyBy: Int,
+                                                  elements: Int,
+                                                  withJitter: Boolean = true,
+                                                  jitterValue: Int = 1000
+  ): Seq[FiniteDuration] = {
+    val intervals = Seq.iterate(startingValue, elements)(_ * multiplyBy).map(_ milliseconds)
+    if (withJitter) intervals else intervals.map(i => addJitter(i, jitterValue milliseconds))
```
Shakespeared: Is this conditional backwards? Should it be `if (withJitter) intervals.map(i => addJitter(i, jitterValue milliseconds)) else intervals`?

sjkobori: yeah, I missed that, good catch 👍

util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
```diff
@@ -94,19 +94,23 @@ trait Retry {
...
-    )
-    plainIntervals.map(i => addJitter(i, 1000 milliseconds))
+  private val allBackoffIntervals = Seq(100 milliseconds, 1 second, 3 seconds, 60 seconds)
+
+  // starting value in milliseconds, multiply by multiples of this each step, how many steps/elements
+  protected def createExponentialBackOffIntervals(startingValue: Int,
+                                                  multiplyBy: Int,
+                                                  elements: Int,
+                                                  withJitter: Boolean = true,
+                                                  jitterValue: Int = 1000
```
tlangs: We could combine these two values into a single `jitter: Option[Int]`, where the presence of a value indicates that jitter should happen.

util/src/test/scala/org/broadinstitute/dsde/workbench/util/RetrySpec.scala
```diff
@@ -37,24 +37,25 @@ class RetrySpec
     TestKit.shutdownActorSystem(system)
 
   "Retry" should "retry 3 times by default" in {
```
tlangs: This test name no longer lines up with the functionality of the test.

util/src/test/scala/org/broadinstitute/dsde/workbench/util/RetrySpec.scala
```diff
@@ -37,24 +37,25 @@ class RetrySpec
...
     val testable = new TestRetry(system, setUpMockLogger)
     import testable._
 
     val result: Future[Int] = testable.retry()(() => testable.failure)
 
     // result should be a failure
-    // invocationCount should be 4 (1 initial run plus 3 retries)
-    // log count should be 4 (3 retries plus 1 failure message)
+    // invocationCount should be 5 (1 initial run plus 4 retries)
+    // log count should be 5 (4 retries plus 1 failure message)
```
tlangs, line 48: Why did the number of retries change?

tlangs (changes requested):

util/src/test/scala/org/broadinstitute/dsde/workbench/util/RetrySpec.scala
sjkobori: Since on line 97 of Retry.scala, we added a new element to allBackoffIntervals which is used for the default amount of retries.

Ghost-in-a-Jar (approved):

util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
```diff
@@ -94,19 +94,27 @@ trait Retry {
...
+                                                  elements: Int,
+                                                  jitterValue: Option[Int] = Some(1000)
+  ): Seq[FiniteDuration] = {
+    val intervals = Seq.iterate(startingValue, elements)(_ * multiplyBy).map(_ milliseconds)
+    jitterValue match {
+      case Some(value) =>
+        intervals.map(i => addJitter(i, value milliseconds))
+      case None =>
+        intervals
+    }
```
tlangs, line 111: This can be expressed as 
```scala
jitterValue.flatMap(intervals.map(i => addJitter(i, value milliseconds)).getOrElse(intervals)
```

tlangs: Doesn't have to be rewritten this way. @Ghost-in-a-Jar thoughts?

tlangs (approved):

util/src/main/scala/org/broadinstitute/dsde/workbench/util/Retry.scala
sjkobori: yeah, I think I could go either way. I just figured this one is more legible, but I am happy to switch it if it is preferable.

PR #1501, opened 2023-08-28
[IA-4447] make billing project deletion optional for use in parallel tests

jdcanas:
Currently the function `withTemporaryAzureBillingProject` assumes the project is scoped to a single test, but this would force IA to run our tests sequentially. We would like to call it once and re-use the same billing project across tests, and control cleanup separately from that life cycle.


LizBaldo (approved):
LGTM, thanks 👍

cahrens:
Seems to be a scalafmt issue.

cahrens (approved):

PR #1563, opened 2023-10-20
Switch delete to async method (WOR-1284).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-1284
Successful Azure workspace e2e test run using this version: https://github.com/broadinstitute/rawls/actions/runs/6647542504/attempts/1

**PR checklist**

In all cases:

### DIFF
--- a/serviceTest/CHANGELOG.md
+++ b/serviceTest/CHANGELOG.md
@@ -4,11 +4,12 @@ This file documents changes to the `workbench-service-test` library, including n
 
 ## 4.1
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "4.1-7362eef"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "4.1-TRAVIS-REPLACE-ME"`
 
 ### Dependency upgrades
 - updated withTemporaryBillingProject to optionally not cleanup the billing project
 - updated `rawls-model` dependency to `0.1-fb0c9691b`
+- changed Rawls workspace deletion to v2 API
 
 | Dependency   | Old Version | New Version |
 |----------|:-----------:|------------:|
--- a/serviceTest/CHANGELOG.md
+++ b/serviceTest/CHANGELOG.md
@@ -2,14 +2,20 @@
 
 This file documents changes to the `workbench-service-test` library, including notes on how to upgrade to new versions.
 
+## 4.2
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "4.2-TRAVIS-REPLACE-ME"`
+
+### Changed
+- changed Rawls workspace deletion to v2 API (async plus polling)
+
 ## 4.1
 
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "4.1-TRAVIS-REPLACE-ME"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-service-test" % "4.1-7362eef"`
 
 ### Dependency upgrades
 - updated withTemporaryBillingProject to optionally not cleanup the billing project
 - updated `rawls-model` dependency to `0.1-fb0c9691b`
-- changed Rawls workspace deletion to v2 API
 
 | Dependency   | Old Version | New Version |
 |----------|:-----------:|------------:|
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
@@ -1,17 +1,17 @@
 package org.broadinstitute.dsde.workbench.service
 
+import akka.http.scaladsl.model.StatusCodes
 import com.fasterxml.jackson.databind.JsonNode
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
-import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{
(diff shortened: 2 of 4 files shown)


serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -253,11 +256,47 @@ trait Rawls extends RestClient with LazyLogging {
       postRequest(url + s"api/workspaces/$sourceNamespace/$sourceName/clone", request)
     }
 
-    def delete(namespace: String, name: String)(implicit token: AuthToken): Unit = {
+    /** Delete the workspace, using the asynchronous v2 workspaces API. This method will poll
+     * for the workspace to be deleted, and throw an Exception if it does not delete within the
+     * specified second timeout interval (which has a default of 300 seconds = 5 minutes).
+     * */
+    def delete(namespace: String, name: String, timeout: Long = 300)(implicit token: AuthToken): Unit = {
```
cahrens, line 263: Do folks think a default of 5 minutes is too long? I decided to make the timeout configurable since if one uses it to delete a workspace with Cromwell in it, for instance, the timeout length will have to be considerably longer. And now that it's configurable, perhaps a default of 3 minutes would be reasonable.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -253,11 +256,47 @@ trait Rawls extends RestClient with LazyLogging {
...
+    def isWorkspaceDeleted(namespace: String, name: String, authToken: AuthToken): Boolean =
+      try {
+        logger.info(s"Checking workspace details status ${namespace}/${name}...")
+        val response = getWorkspaceDetails(namespace, name)(authToken)
+        val workspaceState = mapper.readTree(response).at("/workspace/state").asText()
+        logger.info(s"Workspace ${namespace}/${name} is in state ${workspaceState}")
+        false
+      } catch {
+        case e: RestException =>
+          if (e.statusCode == StatusCodes.Forbidden) {
```
cahrens, line 286: While running the e2e tests, I twice encountered this issue. I imagine it's a race condition in the process of WSM related things being deleted. 

We could possibly try to make the Rawls deletion code handle this case. For now, I decided to simply continue polling if we encounter this, until we either get a 404, get some other error code, or timeout. Note though that this could put additional burden on other teams, unless their code just polls until it either gets a 404 or times out.

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala
```diff
@@ -253,11 +256,47 @@ trait Rawls extends RestClient with LazyLogging {
...
+            // Sometimes we get "User X is not authorized to perform action read on workspace Y".
+            logger.info(
+              s"Encountered ${e.statusCode} while deleting workspace ${namespace}/${name}, continuing polling."
+            )
+            false
+          } else if (e.statusCode == StatusCodes.NotFound) {
+            logger.info(s"Workspace ${namespace}/${name} deleted.")
+            true
+          } else {
+            throw new Exception(s"Error (${e.statusCode}) deleting workspace ${namespace}/${name}")
```
cahrens, line 296: At this point, I think it's helpful for us to know if we hit unexpected errors. For instance, without the fix in https://github.com/broadinstitute/rawls/pull/2591, this code did catch the unexpected 501 ([e2e test run with reverted PR](https://github.com/broadinstitute/rawls/actions/runs/6647547386)).

```
[info] - should allow creation and deletion of azure workspaces *** FAILED *** (1 minute, 2 seconds)
[info]   java.lang.Exception: Error (501 Not Implemented) deleting workspace tmp-billing-project-3b49a3ed-c/b8093944-0070-489a-a567-c2302e66f339-azure-test-workspace
[info]   at org.broadinstitute.dsde.workbench.service.Rawls$workspaces$.isWorkspaceDeleted(Rawls.scala:296)

File: serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala (hunk 1 above)
Comment (cahrens) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:263]: For reference, this is how long tests are taking when creating and deleting workspaces without WDS:

image

Comment (blakery) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:263]: I don't think it's too long at all.
I've regularly seen it take that long, or even longer at times, even with an empty workspace.

File: serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala (hunk 2 above)
Comment (blakery) [reply serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/Rawls.scala:286]: Seems like a reasonable solution: if you're trying to test something where you're expecting a 403 here, it probably shouldn't be using this library.

**Review by blakery — APPROVED** [review]

File: serviceTest/CHANGELOG.md [hunk 4]
```diff
@@ -4,11 +4,12 @@ This file documents changes to the `workbench-service-test` library, including n
 
 ## 4.1
```
marctalbott, line 12: Worth a minor version bump? I know it's backwards compatible and all that, but the polling may warrant it

marctalbott (approved):

serviceTest/CHANGELOG.md
cahrens: I considered it but was lazy. I think you are right though… I will do it.

cahrens: https://github.com/broadinstitute/workbench-libs/pull/1563/commits/2d0d8abd53a91e156b80b27168bac18feb695f58

PR #1622, opened 2023-11-29
[WOR-1192] Autoclass terminal storage class

marctalbott:
Ticket: [WOR-1192](https://broadworkbench.atlassian.net/browse/WOR-1192)
* Optional param to set [terminal autoclass storage class](https://cloud.google.com/storage/docs/using-autoclass). Defaults to `ARCHIVE`

**PR checklist**

In all cases:

[WOR-1192]: https://broadworkbench.atlassian.net/browse/WOR-1192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
@@ -22,7 +22,7 @@ import com.google.cloud.storage.Storage.{
   BucketTargetOption,
   SignUrlOption
 }
-import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, BucketInfo, Storage, StorageOptions}
+import com.google.cloud.storage.{Acl, Blob, BlobId, BlobInfo, BucketInfo, Storage, StorageClass, StorageOptions}
 import com.google.cloud.{Identity, Policy, Role}
 import fs2.{text, Pipe, Stream}
 import org.typelevel.log4cats.StructuredLogger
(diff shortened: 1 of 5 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleStorageInterpreter.scala
```diff
@@ -368,7 +369,13 @@ private[google2] class GoogleStorageInterpreter[F[_]](
       .toBuilder
       .setLabels(labels.asJava)
       .setIamConfiguration(iamConfig)
-      .setAutoclass(BucketInfo.Autoclass.newBuilder().setEnabled(autoclassEnabled).build())
+      .setAutoclass(
+        BucketInfo.Autoclass
+          .newBuilder()
+          .setEnabled(autoclassEnabled)
+          .setTerminalStorageClass(autoclassTerminalStorageClass.getOrElse(StorageClass.ARCHIVE))
```
marctalbott, line 376: In testing I confirmed that setting the terminal storage class when autoclass is disabled has no effect but doesn't fail either -- the bucket is successfully created without autoclass.

cahrens: Do we not have any tests about what we set on autoclass? I would expect test coverage for the default archive value and the fact that it can be overridden by autoclassTerminalStorageClass.

marctalbott: We don't have any tests around bucket creation at the moment

marctalbott: The `LocalStorageHelper` that's used in the existing `GoogleStorageInterpreter` tests doesn't support creating buckets so I'm not sure how we would go about adding tests either unfortunately.

cahrens (approved):

aherbst-broad (approved):

PR #1644, opened 2024-01-09
[IA 4766] Setting the transportChannelProvider to name more threads

lucymcnatt:
Wherever the backgroundExecutorProvider is set, now the transportChannelProvider is set as well, with the executor as the same between both.

Ran into some git weirdness, had to cherry pick to this PR (why the timeline might look off)

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEService.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEService.scala
@@ -6,6 +6,7 @@ import cats.mtl.Ask
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.gax.core.FixedCredentialsProvider
+import com.google.api.gax.rpc.FixedTransportChannelProvider
 import com.google.api.services.container.Container
 import com.google.cloud.container.v1.{ClusterManagerClient, ClusterManagerSettings}
 import com.google.common.util.concurrent.ThreadFactoryBuilder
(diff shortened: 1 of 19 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GKEService.scala
```diff
@@ -73,16 +73,15 @@ object GKEService {
       credential <- credentialResource(pathToCredential.toString)
       credentialsProvider = FixedCredentialsProvider.create(credential)
       executorProviderBuilder = ClusterManagerSettings.defaultExecutorProviderBuilder()
-      threadFactory = new ThreadFactoryBuilder()
```
mlilynolting, line 76: Does this change remove the thread naming on the threadFactory used by the backgroundExecutorProvider?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -123,13 +123,16 @@ object GooglePublisherInterpreter {
     val threadFactory = new ThreadFactoryBuilder().setNameFormat("goog-publisher-%d").setDaemon(true).build()
     val fixedExecutorProvider =
       FixedExecutorProvider.create(new ScheduledThreadPoolExecutor(numOfThreads, threadFactory))
+    val transportChannelProvider =
```
mlilynolting, line 127: Maybe just add a note that we have to use the TopicAdminSettings default provider because Publisher class doesn't have one?

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/package.scala
```diff
@@ -136,6 +138,19 @@ package object google2 {
 
   implicit val showProject: Show[Option[Project]] =
     Show.show[Option[Project]](project => s"project name: ${project.map(_.getName)}")
+
+  def getExecutorProvider[F[_]](builder: InstantiatingExecutorProvider.Builder,
```
mlilynolting, line 142: This is a great helper

mlilynolting (approved):
Looks good!

PR #1646, opened 2024-01-12
Publisher/Subscriber implementation for Azure Service Bus

giventocode:
In this PR:
- `utils2`:  New traits that define cloud-agnostic interfaces for messaging publishing and subscribing: `CloudPublisher` and `CloudSubscriber`
- `azure`: Azure Service Bus implementations of `CloudPublisher` and `CloudSubscriber`. 
- `google2`: Implemention of `CloudPublisher` and `CloudSubscriber` interfaces in the `GoogleSubscriber` and `GooglePublisher` classes.

Pending Items:
- [x] Add to the PR manual test that checks the connection and functionality of Service Bus.

**PR checklist**

In all cases:


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureSubscriber.scala
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+  def stringResource[F[_]: Async: StructuredLogger](
+    subscriberConfig: AzureServiceBusSubscriberConfig,
+    queue: Queue[F, AzureEvent[String]]
+  ): Resource[F, AzureSubscriber[F, String]] =
+    AzureSubscriberInterpreter.stringSubscriber(subscriberConfig, queue)
+}
+
+//using google protobuf Timestamp for consistency
```
Qi77Qi: think the comment is outdated?

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusReceiverClientWrapperInterp.scala
```diff
@@ -0,0 +1,70 @@
...
+      subscriberConfig.connectionString.getOrElse(throw new Exception("Connection string not provided"))
+    )
+    .processor()
+    .topicName(subscriberConfig.topicName)
+    .processMessage(handleMessage)
+    .processError(handleError)
+    .subscriptionName(subscriberConfig.subscriptionName)
+    .buildProcessorClient()
+
+  private def handleMessage(context: ServiceBusReceivedMessageContext): Unit =
```
Qi77Qi: is this one and `handleError` needed?

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisher.scala
```diff
@@ -0,0 +1,33 @@
...
+}
+
+object AzurePublisher {
+  def resource[F[_]: Async: StructuredLogger](
+    clientWrapper: AzureServiceBusSenderClientWrapper
+  ): Resource[F, AzurePublisher[F]] =
+    AzurePublisherInterpreter.publisher(clientWrapper)
+}
+
+final case class AzureServiceBusPublisherConfig(topicName: String, connectionString: Option[String])
```
Qi77Qi: any reason the connectionString is an `Option`?

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureSubscriberInterp.scala
```diff
@@ -0,0 +1,161 @@
...
+            (message: ServiceBusReceivedMessage) => handleMessage(callback, message),
+            (failure: Throwable) => handleReceiveError(callback, failure)
+          )
+      )
+    ).as(None)
+  }
+
+  private def handleReceiveError(callback: Either[Throwable, Unit] => Unit, failure: Throwable): Unit = {
+    val handleError = for {
+      _ <- logger.error(s"Error in message subscription: ${failure.getMessage}")
```
Qi77Qi: `logger.error(failure)(???)`...this way the stacktrace will be logged properly

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureSubscriberInterp.scala
```diff
@@ -0,0 +1,161 @@
+package org.broadinstitute.dsde.workbench.azure
+
+import cats.effect._
+import cats.effect.std.Dispatcher
+import cats.implicits._
+import com.azure.messaging.servicebus.ServiceBusReceivedMessage
+import com.google.protobuf.Timestamp
```
Qi77Qi: don't think the import is used anymore

cpate4:
@giventocode - Out of curiosity, is there a design doc this is associated with to better understand how this will fit in with the rest of the system / architecture? There has been internal discussions about introducing cloud-agnostic code for handling Pub/Sub messages, and I am curious if this fits in with that thought pattern. Thanks!

cc: callisto-team (@LizBaldo, @mlilynolting, @lucymcnatt, @jdcanas, @cindy-broadinstitute)

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisher.scala
```diff
@@ -0,0 +1,38 @@
...
+import cats.mtl.Ask
+import com.azure.messaging.servicebus.ServiceBusMessage
+import fs2.Pipe
+import io.circe.Encoder
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.typelevel.log4cats.StructuredLogger
+
+import java.time.Duration
+
+trait AzurePublisher[F[_]] {
```
dvoet: the is almost identical to `GooglePublisher`. Should we have a single trait that handles both then have interpreters for Azure and Google? Same goes for `AzureSubscriber` and `GoogleSubscriber`.

giventocode: A single trait would be ideal, but I could not find a practical way to implement it. The project is structured so that each library (google, google2, azure) is independent of each other. This means I would have to either create another library (overkill) or cross-reference google and azure libraries (ugly). Instead of doing that, I created the same signature to maintain the same semantics and facilitate the creation of a single trait with azure and google implementations in the service. This seemed like the least bad option, but I am open to any suggestions.

Qi77Qi: it can live in `util2`, both libs depend on it. don't feel strongly either way tho

Qi77Qi:
oh, I might've some outdated comments that I didn't submit previously...ignore if they're no longer applicable

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisher.scala
dvoet: This seems like the perfect spot to be cloud agnostic.

rtitle: An approach where the interface is in `util2` and the implementations are in `google2` and `azure` sounds good to me.

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisherInterp.scala
```diff
@@ -0,0 +1,82 @@
...
+    traceId.foreach(id => message.setCorrelationId(id.asString))
+
+    publishServiceBusMessage(message)
+  }
+  private def publishServiceBusMessage(message: ServiceBusMessage): F[Unit] =
+    withLogging(
+      Async[F]
+        .async[Unit] { cb =>
+          Async[F]
+            .delay(
```
rtitle: This maybe should be `Async[F].blocking(..)` instead of `Async[F].delay(..)`. Semantically they're the same, just `blocking` shift the execution on a new cached thread pool. 

(We've seen issues in the past with blocking calls happening on the main thread pool via `.delay`.)

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisher.scala
giventocode: When the connection string is not set, it will use MI.

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/messaging/CloudPublisher.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsde.workbench.util2.messaging
+
+import cats.mtl.Ask
+import fs2.Pipe
+import io.circe.Encoder
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+trait CloudPublisher[F[_]] {
+  def publish[MessageType: Encoder]: Pipe[F, MessageType, Unit]
+  def publishOne[MessageType: Encoder](message: MessageType, messageAttributes: Option[Map[String, String]] = None)(implicit ev: Ask[F, TraceId]): F[Unit]
```
giventocode: I would like to know if using Ask to wrap the TraceId as an implicit value is the best approach. I followed the existing trait (`GooglePublisher`) as a reference, but now I think it might be better to make it an optional parameter with a default value, similar to `messageAttributes`. I would appreciate any feedback.

Qi77Qi: I don't feel strongly passing explicitly or implicitly (might be easier to use in `Rawls` if it's explicit)...when the code is used in production, I think we should force caller to pass a `traceId` ideally. But I can see value providing default if you want to provide convenience to caller who doesn't care about being able to link log messages

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusReceiverClientWrapper.scala
```diff
@@ -0,0 +1,16 @@
...
+
+// A simple wrapper around the Azure ServiceBusReceiverAsyncClient to facilitate testing
+// given that the client is final and serializable, which prevents mockito from mocking it
+trait AzureServiceBusReceiverClientWrapper {
+  def startProcessor(): Unit
+  def stopProcessor(): Unit
+}
+
+object AzureServiceBusReceiverClientWrapper {
+  def createReceiverClientWrapper[F[_], MessageType](
```
rtitle: It doesn't seem like the `MessageType` type param is used here?

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusReceiverClientWrapperInterp.scala
```diff
@@ -0,0 +1,67 @@
...
+import scala.util.{Failure, Success, Try}
+
+class AzureServiceBusReceiverClientWrapperInterp(subscriberConfig: AzureServiceBusSubscriberConfig,
+                                                 messageHandler: AzureReceivedMessageHandler
+) extends AzureServiceBusReceiverClientWrapper {
+
+  private val processor: ServiceBusProcessorClient = createNewProcessor()
+
+  private def handleError(context: ServiceBusErrorContext): Unit =
+    println(s"Error when receiving message: ${context.getException}.")
```
rtitle, line 16: I think we should avoid the use of `println` - can we use `slf4j`?

I think you should be able to do something like the following:

1. Add the `StructuredLogger` implicit to `createReceiverClientWrapper`:
```
object AzureServiceBusReceiverClientWrapper {
  def createReceiverClientWrapper[F[_]: StructuredLogger](
    subscriberConfig: AzureServiceBusSubscriberConfig,
    messageHandler: AzureReceivedMessageHandler
  ): AzureServiceBusReceiverClientWrapper =
    new AzureServiceBusReceiverClientWrapperInterp(subscriberConfig, messageHandler)
}
```

2. Add `StructuedLogger` implicit to `AzureServiceBusReceiverClientWrapperInterp`:
```
class AzureServiceBusReceiverClientWrapperInterp[F[_]: StructuredLogger](…)
```

3. "Summon" the implicit at the call site:
```
  private def handleError(context: ServiceBusErrorContext): Unit =
    StructuredLogger[F].error(s"Error when receiving message: ${context.getException.getMessage}")
```

rtitle: I think we should avoid the use of `println` - can we use `slf4j`?

You could do something like:
```
  private val logger = LoggerFactory.getLogger(classOf[AzureServiceBusReceiverClientWrapperInterp])

… 

  private def handleError(context: ServiceBusErrorContext): Unit =
    logger.error(s"Error when receiving message", context.getException)
```

google2/CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
 
 This file documents changes to the `workbench-google2` library, including notes on how to upgrade to new versions.
 
+## 0.36
```
rtitle, line 5: Maybe add a `SBT Dependency:` heading under `0.36` to follow the existing pattern.

rtitle (approved):
Left some minor comments, but overall looks good to me. Thanks for refactoring the Google PubSub stuff to use a common interface

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusReceiverClientWrapperInterp.scala
giventocode: ahh, yes, thanks, these were temp statements from my manual testing. I'll remove them. Logging (with context) should be covered by the interpreters.

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/messaging/CloudSubscriber.scala
```diff
@@ -0,0 +1,18 @@
...
+  def messages: Stream[F, ReceivedMessage[MessageType]]
+  def start: F[Unit]
+  def stop: F[Unit]
+}
+
+final case class ReceivedMessage[A](msg: A,
+                                    traceId: Option[TraceId],
+                                    publishedTime: Option[Instant],
+                                    ackHandler: AckHandler
+)
```
cpate4: would it be possible to add comments / documentation for how these traits and case classes are intended to be used?

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/messaging/CloudPublisher.scala
```diff
@@ -0,0 +1,14 @@
...
+import io.circe.Encoder
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+trait CloudPublisher[F[_]] {
+  def publish[MessageType: Encoder]: Pipe[F, MessageType, Unit]
+  def publishOne[MessageType: Encoder](message: MessageType, messageAttributes: Option[Map[String, String]] = None)(
+    implicit ev: Ask[F, TraceId]
+  ): F[Unit]
+  def publishString: Pipe[F, String, Unit]
+}
```
cpate4, line 29: would it be possible to add comments / documentation for how these traits and case classes are intended to be used?

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/messaging/AckHandler.scala
```diff
@@ -0,0 +1,6 @@
+package org.broadinstitute.dsde.workbench.util2.messaging
+
+trait AckHandler {
+  def ack(): Unit
+  def nack(): Unit
+}
```
cpate4, line 17: would it be possible to add comments / documentation for how these traits and case classes are intended to be used?

project/Settings.scala
```diff
@@ -113,13 +113,13 @@ object Settings {
...
   ) ++ publishSettings
 
   val azureSettings = commonSettings ++ List(
     name := "workbench-azure",
     libraryDependencies ++= azureDependencies,
-    version := createVersion("0.6")
+    version := createVersion("0.7")
   ) ++ publishSettings
 
   val openTelemetrySettings = commonSettings ++ List(
```
cpate4, line 125: general thought questions ...
are the introduction of these interfaces and changes to base classes _possibly_ backwards incompatible changes (thinking through semver)

google2/CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
...
+
+SBT Dependency: "org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.36-TRAVIS-REPLACE-ME"
+
+### Changes
+
+- GoogleSubscriberInterpreter (**Breaking**):
```
cpate4, line 11: I think this confirms my suspicions in another comment re: this is a breaking change and I believe should be marked as such (e.g. 1.0-TRAVIS-REPLACE-ME) - congratulations!!! 🥳

project/Settings.scala
cpate4: ... see comments in `google2/CHANGELOG.md` - it might be that is the only place this backwards incompatible comment applies

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -63,18 +66,27 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: StructuredLogger]
...
     tracedLogging(asyncPublishMessage(byteString), s"com.google.cloud.pubsub.v1.Publisher.publish($byteString)")
   }
+  override def publishOne[MessageType: Encoder](message: MessageType,
+                                                messageAttributes: Option[Map[String, String]] = None
+  )(implicit ev: Ask[F, TraceId]): F[Unit] = {
+    val byteString = ByteString.copyFromUtf8(message.asJson.noSpaces)
+    tracedLogging(asyncPublishMessage(byteString, messageAttributes),
+                  s"com.google.cloud.pubsub.v1.Publisher.publish($byteString)"
+    )
+  }
```
cpate4, line 76: is this code tested elsewhere? is this something we should be validating? (where `#publishOne(...)` is our new way of publishing messages?)

cpate4: ... it looks like we created a new method with a separate code path and left the original codepath alone. Is it at all possible to push the coerce the two separate code paths? What I am thinking about / concerned with is this code having to support the same set of publish / subscribe. I believe we ideally need them to stay in sync, and the only change (ultimately) should be in the API that we are calling from our clients.

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisher.scala
```diff
@@ -39,10 +40,18 @@ trait GooglePublisher[F[_]] {
 }
 
 object GooglePublisher {
+  @deprecated("Use cloudPublisherResource instead", "0.36")
```
cpate4, line 43: The subscriber looks like it had a harder break - and as such, I am curious if there is any benefit in marking this as deprecated vs just removing it from the code

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GoogleSubscriberInterpreter.scala
```diff
@@ -303,4 +311,13 @@ final case class SubscriberConfig(
 final case class MaxRetries(value: Int) extends AnyVal
 final case class SubscriberDeadLetterPolicy(topicName: TopicName, maxRetries: MaxRetries)
 
+final private[google2] class AckGoogleHandler(consumer: AckReplyConsumer) extends AckHandler {
+  override def ack(): Unit = consumer.ack()
+  override def nack(): Unit = consumer.nack()
+}
+
+object AckGoogleHandler {
+  def createAckHandler(consumer: AckReplyConsumer): AckGoogleHandler = new AckGoogleHandler(consumer)
```
cpate4, line 320: an alternative might be to use `.apply(...)`, and you can use the `.createAckHandler(...)` call at the call-site, but totally up to you

LizBaldo:
LGTM, I am excited about having a common, cloud agnostic trait for pubsub :)
You might need to make tweaks once you start testing this within leonardo though. If you wanted to, you could leave this PR open, build and publish a snaphot version of the image and use that in your leo PR if you want to save yourself some dev testing cycles

LizBaldo (approved):

cpate4 (changes requested):
I definitely liked reading the Azure code (many more comments and descriptions in the code about what and why things were the way they were). 

Biggest questions/comments were:
- does the GCP stuff (at a minimum) need to be a semver breaking change and a major version bump, and
- would it be possible to add the same level of comments that we found in the Azure code paths to the interfaces / traits re: usage, etc?

Thanks!

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisherInterp.scala
```diff
@@ -0,0 +1,100 @@
...
+) extends CloudPublisher[F] {
+
+  override def publish[MessageType: Encoder]: Pipe[F, MessageType, Unit] = in =>
+    in.flatMap { message =>
+      Stream
+        .eval(publishMessage(message.asJson.noSpaces, None))
+    }
+
+  override def publishOne[MessageType: Encoder](message: MessageType,
+                                                messageAttributes: Option[Map[String, String]] = None
```
Qi77Qi: maybe do `messageAttributes: Map[String, String] = Map.empty`? Options add more permutation, but I think in this case, we don't need that extra complexity

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusSenderClientWrapper.scala
```diff
@@ -0,0 +1,16 @@
+package org.broadinstitute.dsde.workbench.azure
+
+import com.azure.messaging.servicebus.ServiceBusMessage
+import reactor.core.publisher.Mono
+
+// A simple wrapper around the Azure ServiceBusSenderAsyncClient to facilitate testing
+// given that the client is final and serializable, which prevents mockito from mocking it
+trait AzureServiceBusSenderClientWrapper {
```
Qi77Qi: shall we tag this one and the implementation class as `private`? I don't think client needs to be aware of this class

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
```diff
@@ -63,18 +66,27 @@ private[google2] class GooglePublisherInterpreter[F[_]: Async: StructuredLogger]
...
 
   private def publishMessage(message: String, traceId: Option[TraceId]): F[Unit] = {
     val byteString = ByteString.copyFromUtf8(message)
     withLogging(asyncPublishMessage(byteString), traceId, s"Publishing $message")
   }
 
-  private def asyncPublishMessage(byteString: ByteString): F[Unit] =
+  private def asyncPublishMessage(byteString: ByteString, attributes: Option[Map[String, String]] = None): F[Unit] =
```
cpate4: ... see @Qi77Qi 's comment above re: empty map vs none - 🙌

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/GooglePublisherInterpreter.scala
giventocode: I made minimal changes to the existing GCP implementation to avoid refactoring the tests and to have a quick way to reference the old implementation to facilitate troubleshooting issues as we refactor Leo. Cleaning this up can certainly be a follow-up item.

project/Settings.scala
giventocode: For Azure, this is a net new functionality that does not introduce any breaking changes.

cindy-broadinstitute (approved):
LGTM, it is great to have a cloud agnostic trait for pubsub.

google2/CHANGELOG.md
giventocode: This library hasn't had any major version upgrades, so I am not sure this change qualifies for that "honor" 😄. I understand and agree with the semver point, but since there is ongoing work in Leo to integrate these changes, and Leo is currently the only client of this library, a minor version update seems reasonable to me.

azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzurePublisherInterp.scala
giventocode: Yeah, I don't like the additional boiler plate required here as well. But this is the resource that made go this way:
https://stackoverflow.com/questions/65256690/best-way-to-model-optional-arguments
I am not a Scala expert, but Option seems to be the "preferred" way.

Qi77Qi: I think it's generally true for optional fields..but in this case, by just doing `messageAttributes: Map[String, String] = Map.empty`, caller can already optionally provide the argument...`Option` isn't adding any value, but rather complicating logic becuz now implementation needs to accomadate for `None`, empty map, which both are the same case

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/messaging/CloudSubscriber.scala
```diff
@@ -0,0 +1,37 @@
...
+  def stop: F[Unit]
+}
+
+/***
+ * A message received from a cloud messaging service.
+ * Contains the message itself, along with metadata about the message, and the ability to ack/nack the message via the ackHandler.
+ */
+final case class ReceivedMessage[A](msg: A,
+                                    traceId: Option[TraceId],
+                                    publishedTime: Option[Instant],
```
Qi77Qi: any reason why `publishedTime` is optional? in other words, when will it be valid for it to be None?

giventocode: Good catch. It should not be optional. Initially I thought this was a user-defined property, but later realized is system set and for SB is mapped to the EnqueueTime which should be in all messages. Will fix.

cpate4 (approved):
Thank you for doing this, @giventocode 
We did talk about the major version bump and NOT doing it.
@rtitle / @dvoet - do we have a documented process for how to do this major version bump so we can follow `semver`?

PR #1668, opened 2024-03-04
Don't fail tests on cleanup - simple version

blakery:
Ticket: [QA-2256](https://broadworkbench.atlassian.net/browse/QA-2256)
no longer throw exception when cleanup fails

**PR checklist**

In all cases:

[QA-2256]: https://broadworkbench.atlassian.net/browse/QA-2256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
+++ b/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
@@ -112,9 +112,9 @@ trait CleanUp extends TestSuiteMixin with ExceptionHandling with LazyLogging { s
 object CleanUp {
   implicit val errorReportSource: ErrorReportSource = ErrorReportSource("WorkbenchLibs.CleanUp")
 
-  def runCodeWithCleanup[T, C](testTrial: Try[T], cleanupTrial: Try[C]): T =
-    (testTrial, cleanupTrial) match {
-      case (Success(outcome), _) => outcome
-      case (Failure(t), _)       => throw t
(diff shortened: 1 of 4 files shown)


Ghost-in-a-Jar (approved):

serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/test/CleanUp.scala
```diff
@@ -114,18 +114,7 @@ object CleanUp {
 
   def runCodeWithCleanup[T, C](testTrial: Try[T], cleanupTrial: Try[C]): T =
     (testTrial, cleanupTrial) match {
```
rtitle: Nit: if we're not matching on `cleanupTrial`, then we don't need to include it in the match statement. Maybe it doesn't even need to be passed to this method?

blakery: We considered this.

In the end, we didn't want to change the signature, and figured that leaving it like this made it read fairly clearly that 'we don't care if clean-up fails'.

jyang-broad: Created a ticket to clean this up the signature, but it's too much scope to handle atm.

https://broadworkbench.atlassian.net/browse/QA-2255

jyang-broad (approved):
removes failure cases we don't care about (cleanup failures)

marctalbott (approved):

PR #1669, opened 2024-03-05
[IA-4578] Handle Google 403 "Compute Engine API not enabled" error

mlilynolting:
Treat it like "billing is disabled" error when getting Google Compute instance: return None, as though the instance didn't exist. Downstream, the Zombie Monitor cronjob will delete Leo runtime records corresponding to the instance, as desired.

Other uses of getInstance should continue to work unchanged. The Compute Engine API is enabled by Buffer when it prepares a Google project to be claimed by Rawls during GCP workspace creation. There is no window of time duringa  Terra workspace lifecycle when this API will be disabled, other than upon disablement of the Google billing account.

**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Workbench utility libraries, built for 2.13. You can find the full list of packa
 
 Contains utility functions for talking to Google APIs via com.google.cloud client library (more recent) via gRPC.
 
-Latest SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-google2" % "0.36-ad61f19"`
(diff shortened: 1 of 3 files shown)


LizBaldo (approved):
That makes sense to have the logic in here 👍

LizBaldo:
One thing though, is there a corresponding unit test to modify as well?

mlilynolting:
> One thing though, is there a corresponding unit test to modify as well?

It doesn't seem so? workbench-libs has "manual test" files which I admit I'm unsure of how to use. There aren't unit tests.

PR #1675, opened 2024-04-11
TOAZ-348 remove google oauth support, add billing b2c profile

dvoet:
Epic [TOAZ-347](https://broadworkbench.atlassian.net/browse/TOAZ-347)
[TOAZ-348](https://broadworkbench.atlassian.net/browse/TOAZ-348)
- remove all support for google oauth
- support for a new security scheme named `oidc_google_billing_scope` in open api yaml
- support for populating the google billing b2c profile in open api yaml
- support for passing b2c profile to token endpoint via query param in addition to form field because that's the only way I could figure out how to pass it in the swagger ui

**PR checklist**

In all cases:

[TOAZ-347]: https://broadworkbench.atlassian.net/browse/TOAZ-347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TOAZ-348]: https://broadworkbench.atlassian.net/browse/TOAZ-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oauth2/CHANGELOG.md
+++ b/oauth2/CHANGELOG.md
@@ -2,8 +2,13 @@
 
 This file documents changes to the `workbench-oauth2` library, including notes on how to upgrade to new versions.
 
+## 0.7
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-oauth2" % "0.7-TRAVIS-REPLACE-ME"`
+Changed:
+- removed support for google oauth2, only azure b2c authentication is supported now in Terra UIs and swagger-ui
+
 ## 0.6
-SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-oauth2" % "d314413"`
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-oauth2" % "0.6-d314413"`
 Changed:
 - Updated `OpenIDProviderMetadata` class to include an optional `end_session_endpoint`
 
--- a/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
+++ b/oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
@@ -8,7 +8,7 @@ import akka.http.scaladsl.model.HttpMethods.POST
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.server.Directives._
-import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.{Rejection, RejectionError, Route, ValidationRejection}
 import akka.stream.scaladsl.Flow
 import akka.util.ByteString
 import io.circe.Encoder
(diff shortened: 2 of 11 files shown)


Ghost-in-a-Jar (approved):

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -83,6 +87,24 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
     }
   }
 
+  /**
+   * Determine the query to use for the token request.
+   * If the policy parameter is present in both the query and the form, and they are different, throw an exception.
+   * If the policy parameter is present in the query or the form, use that value.
+   * Otherwise, use the query from the tokenUri.
+   */
+  private def determinePolicyQuery(policyInQuery: Option[String],
```
rtitle, line 96: Query params in the `token` endpoint are a little odd; normally it expects form params. The `oidc-client-ts` library (which Terra UI uses) has `extraTokenParams` which appear in the form ([code link](https://github.com/DataBiosphere/terra-ui/blob/dev/src/auth/oidc-broker.ts#L83)).

I see [this](https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/) in Swagger UI docs:
```
additionalQueryStringParams	
OAUTH_ADDITIONAL_PARAMS	
Additional query parameters added to authorizationUrl and tokenUrl. MUST be an object
```

If that is used do the params appear in the query (as the name implies) or the form?

rtitle (approved):
Looks good, glad to remove the google stuff!

I was a bit curious about the query vs form params in the `token` endpoint, but if it works not worth blocking the PR, I think.

oauth2/CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
 
 This file documents changes to the `workbench-oauth2` library, including notes on how to upgrade to new versions.
 
+## 0.7
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-oauth2" % "0.7-TRAVIS-REPLACE-ME"`
+Changed:
+- removed support for google oauth2, only azure b2c authentication is supported now in Terra UIs and swagger-ui
```
rtitle, line 8: This reminded me the version of swagger-ui is controlled by wb-libs. Seems we're on 4.11.1, the latest is [5.15.1](https://github.com/swagger-api/swagger-ui/releases). Maybe we could think about upgrading at some point..

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
dvoet: I agree having a query along with the form post is odd. The problem with the swagger ui oauth plugin is that it supports only 1 client. We have custom logic to get around that for client id (which we don't need anymore). But we would need the same kind of thing for `additionalQueryStringParams` because we have 1 client that has some and another that doesn't.

I am going to push forward as-is and look into making it better later, including an upgrade.

PR #1681, opened 2024-05-21
[WM-2631] Add dependency override for bouncy castle

salonishah11:
Jira ticket: https://broadworkbench.atlassian.net/browse/WM-2631

This PR adds dependency override for bouncy castle as the latest `client-java` in `io.kubernetes` package doesn't have non-vulnerable version of bouncy castle yet.

**PR checklist**

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -124,9 +124,9 @@ object Dependencies {
 
   // Note: this override can be removed when "io.kubernetes" % "client-java" publishes a new version containing
   // non-vulnerable bouncy castle version. See https://broadworkbench.atlassian.net/browse/WM-2631
-  val transitiveDependencyOverrides = Seq(
+  val bouncyCastleOverrides = Seq(
     //Override for bouncy castle to address CVE-2024-30172
-    "org.bouncycastle" % "bcpkix-jdk18on" % "1.78",
+    bouncyCastle, bouncyCastleProviderExt, bouncyCastleProvider
   )
 
   val commonDependencies = Seq(
@@ -215,7 +215,7 @@ object Dependencies {
     googleResourceManager,
     scalaCache,
     scalaTestMockito
-  ) ++ transitiveDependencyOverrides
+  )
 
   val azureDependencies = List(
     log4cats,
@@ -235,7 +235,7 @@ object Dependencies {
     byteBuddy
   ) ++ Seq(
     "net.minidev" % "json-smart" % jsonSmartV
-  ) ++ transitiveDependencyOverrides
+  ) ++ bouncyCastleOverrides
 
   val openTelemetryDependencies = List(
     catsEffect,
@@ -263,7 +263,7 @@ object Dependencies {
     circeGeneric,
     catsMtl,
     kubernetesClient
-  ) ++ transitiveDependencyOverrides
+  ) ++ bouncyCastleOverrides
 
   val serviceTestDependencies = commonDependencies ++ Seq(
     scalaLogging,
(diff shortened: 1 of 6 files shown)


project/Dependencies.scala
```diff
@@ -122,6 +122,13 @@ object Dependencies {
     "com.azure.resourcemanager" % "azure-resourcemanager-batch" % "1.0.0"
   val azureServiceBus = "com.azure" % "azure-messaging-servicebus" % "7.14.7"
 
+  // Note: this override can be removed when "io.kubernetes" % "client-java" publishes a new version containing
+  // non-vulnerable bouncy castle version. See https://broadworkbench.atlassian.net/browse/WM-2631
+  val transitiveDependencyOverrides = Seq(
+    //Override for bouncy castle to address CVE-2024-30172
+    "org.bouncycastle" % "bcpkix-jdk18on" % "1.78",
+  )
```
salonishah11, line 130: Note for reviewers - I have added `transitiveDependencyOverrides` to each dependency block below that was importing `kubernetesClient` defined on L75, but let me know if I should add or remove this from any block. Agora uses `workbench-oauth2` (which also imports `workbench-util2`) which is also being updated here.

salonishah11: I had initially updated `"io.kubernetes" % "client-java"` to the latest version (20.0.1) but it seems that there are some changes between 19.0.0 and 20.0.1 which caused the build to break. I think updating to 20.x would need updates to `KubernetesInterpreter` and I don't have the right context to make these changes. Hence I reverted the change for `"io.kubernetes" % "client-java"` and only added transitive dependency override.

project/Dependencies.scala
```diff
@@ -208,7 +215,7 @@ object Dependencies {
     googleResourceManager,
     scalaCache,
     scalaTestMockito
-  )
+  ) ++ transitiveDependencyOverrides
```
mlilynolting: google2 already has a bouncycastle dependency override so this is a duplicate

project/Dependencies.scala
```diff
@@ -122,6 +122,13 @@ object Dependencies {
     "com.azure.resourcemanager" % "azure-resourcemanager-batch" % "1.0.0"
   val azureServiceBus = "com.azure" % "azure-messaging-servicebus" % "7.14.7"
 
+  // Note: this override can be removed when "io.kubernetes" % "client-java" publishes a new version containing
+  // non-vulnerable bouncy castle version. See https://broadworkbench.atlassian.net/browse/WM-2631
+  val transitiveDependencyOverrides = Seq(
```
mlilynolting: I think many of the explicit dependencies here are already transitive overrides; could this be a more specific bouncyCastleOverrides Seq? google2 overrides three distinct ones:
```
  val bouncyCastle: ModuleID = "org.bouncycastle" % "bcpkix-jdk18on" % bouncyCastleVersion
  val bouncyCastleProviderExt: ModuleID = "org.bouncycastle" % "bcprov-ext-jdk18on" % bouncyCastleVersion
  val bouncyCastleProvider: ModuleID = "org.bouncycastle" % "bcprov-jdk18on" % bouncyCastleVersion
  ```
  All 3 could go into a single Seq and be concatenated to each lib

salonishah11: Yes I can do that. I didn't realize there were already pre-existing bouncy castle overrides. Thank you. I might just use them in the Seq here.

mlilynolting (approved):

mspector (approved):

PR #1683, opened 2024-05-29
remove dumps dir

dvoet:
these dumps seem to be coming from broadbot, Commit from GitHub Actions (Publish Jar), and they are just taking up space

**PR checklist**

In all cases:


cpate4 (approved):
This does solve the problem and removes the files, but also curious ... do we want / need to obliterate these from git history? 🫣 

I also went down the rabbit hole of trying to [figure out how to turn this off](https://www.graalvm.org/latest/reference-manual/native-image/overview/Options/), which then triggered a more general question of, 'would we happen to know why are we using this VM versus one of the other "standard ones"?'

Regardless ... #approved

PR #1691, opened 2024-07-01
PROD-972 ID-1323 Add group version and last synchronized version

Ghost-in-a-Jar:
https://broadworkbench.atlassian.net/browse/ID-1324
https://broadworkbench.atlassian.net/browse/PROD-972

We are adding a group version and last synchronized version so that sam can use them to reduce duplicate calls to the group directory api (list membership).

**PR checklist**

In all cases:

### DIFF
--- a/model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
+++ b/model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
@@ -29,7 +29,10 @@ final case class WorkbenchUser(id: WorkbenchUserId,
 )
 final case class WorkbenchUserId(value: String) extends WorkbenchSubject with ValueObject
 
-trait WorkbenchGroup { val id: WorkbenchGroupIdentity; val members: Set[WorkbenchSubject]; val email: WorkbenchEmail; val version: Integer; val lastSynchronizedVersion: Integer }
+trait WorkbenchGroup {
+  val id: WorkbenchGroupIdentity; val members: Set[WorkbenchSubject]; val email: WorkbenchEmail; val version: Integer;
+  val lastSynchronizedVersion: Integer
+}
 trait WorkbenchGroupIdentity extends WorkbenchSubject
 case class WorkbenchGroupName(value: String) extends WorkbenchGroupIdentity with ValueObject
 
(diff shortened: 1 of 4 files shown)


Shakespeared (approved):

dvoet (approved):
I am slightly concerned about unintended consequences of this. I don't recall why we thought it was a good idea to put `WorkbenchGroup` in workbench-libs. But I am not seeing any usage in either rawls or orch which are the only places I would expect.

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -29,7 +29,10 @@ final case class WorkbenchUser(id: WorkbenchUserId,
 )
 final case class WorkbenchUserId(value: String) extends WorkbenchSubject with ValueObject
 
-trait WorkbenchGroup { val id: WorkbenchGroupIdentity; val members: Set[WorkbenchSubject]; val email: WorkbenchEmail }
+trait WorkbenchGroup {
```
rtitle, line 32: Nit: the semicolons are a little weird, maybe replace with newlines?

model/src/main/scala/org/broadinstitute/dsde/workbench/model/WorkbenchIdentity.scala
```diff
@@ -29,7 +29,10 @@ final case class WorkbenchUser(id: WorkbenchUserId,
 )
 final case class WorkbenchUserId(value: String) extends WorkbenchSubject with ValueObject
 
-trait WorkbenchGroup { val id: WorkbenchGroupIdentity; val members: Set[WorkbenchSubject]; val email: WorkbenchEmail; val version: Integer; val lastSynchronizedVersion: Integer }
+trait WorkbenchGroup {
+  val id: WorkbenchGroupIdentity; val members: Set[WorkbenchSubject]; val email: WorkbenchEmail; val version: Integer;
+  val lastSynchronizedVersion: Integer
```
dvoet: maybe make `lastSynchronizedVersion` optional

rtitle (approved):
I also checked for usages of this trait and only see it in Sam.

Often when working in wb-libs I'll `sbt publish` my changes before merging, so I can test the wb-libs code together with Sam/Leo/whatever before merging.

samanehsan (approved):

PR #1705, opened 2024-10-02
AJ-2020: update swagger-ui, add Content-Security-Policy header

davidangb:
This PR is in service to https://broadworkbench.atlassian.net/browse/AJ-2020. That ticket mentions the need for Orch's swagger-ui page to have a `Content-Security-Policy` header. Orch relies on the swagger-ui from workbench-libs, so I'm making changes in workbench-libs.

* updated the version of swagger-ui to latest: 5.17.14
* re-evaluate andd tweaked our customizations to swagger-ui
* evaluated https://broadworkbench.atlassian.net/browse/TOAZ-205 and decided NOT to do that ticket; I think the current approach of checking in a single file containing our customizations is appropriate
* added a `Content-Security-Policy` header, with default value, when serving the swagger-ui html page
* added ability for callers to change the `Content-Security-Policy` headervalue

Tested by publishing to my local ivy, pulling the updates into a local Orchestration, and using the local Orch's swagger ui.

---

**PR checklist**

In all cases:

### DIFF
--- a/oauth2/src/main/resources/swagger/index.html
+++ b/oauth2/src/main/resources/swagger/index.html
@@ -58,7 +58,9 @@
 <script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
 <script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
 <!-- START Terra customization -->
-<!-- <script src="./swagger-initializer.js" charset="UTF-8"> </script> -->
+  <!-- swagger-ui's index.html calls ./swagger-initializer.js. Our customization bypasses that and uses inline
+        javascript instead. -->
+  <!-- <script src="./swagger-initializer.js" charset="UTF-8"> </script> -->
 <script type="text/javascript">
 
   // Adds support for pinning the auth bar when the user scrolls down far enough to hide the bar
--- a/oauth2/src/main/resources/swagger/index.html
+++ b/oauth2/src/main/resources/swagger/index.html
@@ -1,33 +1,15 @@
 <!-- Copied from the swagger-ui/index.html static file -->
 <!DOCTYPE html>
-<html lang="en" xmlns:th="http://www.thymeleaf.org">
+<html lang="en">
 <head>
   <meta charset="UTF-8">
   <title>Swagger UI</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" >
+  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" />
+  <link rel="stylesheet" type="text/css" href="index.css" />
   <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
+  <!-- START workbench-libs customization -->
   <style>
-    html
-    {
-      box-sizing: border-box;
-      overflow: -moz-scrollbars-vertical;
-      overflow-y: scroll;
-    }
-
-    *,
-    *:before,
-    *:after
-    {
-      box-sizing: inherit;
-    }
-
-    body
-    {
-      margin:0;
-      background: #fafafa;
-    }
-
     /* make the schema display full-width */
     .swagger-ui .model-example .model-box {
       display: block;
@@ -67,16 +49,17 @@
     .swagger-ui .hidden {
       display: none;
     }
-
   </style>
+  <!-- END workbench-libs customization -->
 </head>
 
 <body>
 <div id="swagger-ui"></div>
-
-<script src="./swagger-ui-bundle.js"> </script>
-<script src="./swagger-ui-standalone-preset.js"> </script>
-<script th:inline="javascript">
+<script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
+<script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
+<!-- START workbench-libs customization -->
+<!-- <script src="./swagger-initializer.js" charset="UTF-8"> </script> -->
+<script type="text/javascript">
 
   // Adds support for pinning the auth bar when the user scrolls down far enough to hide the bar
   var pinLoginPlugin = function(system) {
@@ -155,5 +138,6 @@
     window.ui = ui
   }
 </script>
+<!-- END workbench-libs customization -->
 </body>
 </html>
--- a/oauth2/src/main/resources/swagger/index.html
+++ b/oauth2/src/main/resources/swagger/index.html
@@ -8,7 +8,7 @@
   <link rel="stylesheet" type="text/css" href="index.css" />
   <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
-  <!-- START workbench-libs customization -->
+  <!-- START Terra customization -->
   <style>
     /* make the schema display full-width */
     .swagger-ui .model-example .model-box {
@@ -50,14 +50,14 @@
       display: none;
     }
   </style>
-  <!-- END workbench-libs customization -->
+  <!-- END Terra customization -->
 </head>
 
 <body>
 <div id="swagger-ui"></div>
 <script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
 <script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
-<!-- START workbench-libs customization -->
+<!-- START Terra customization -->
 <!-- <script src="./swagger-initializer.js" charset="UTF-8"> </script> -->
 <script type="text/javascript">
 
@@ -130,6 +130,6 @@
     window.ui = ui
   }
 </script>
-<!-- END workbench-libs customization -->
+<!-- END Terra customization -->
 </body>
 </html>
--- a/oauth2/src/main/resources/swagger/index.html
+++ b/oauth2/src/main/resources/swagger/index.html
@@ -86,15 +86,6 @@
     }
   }
 
-  // Removes the online validation since some code gen bugs cause us to have some specs that cause the validator to squawk
-  const clearValidator = function(system) {
-    return {
-      components: {
-        onlineValidatorBadge: function() { return null; },
-      }
-    };
-  }
-
   window.onload = function() {
     // Begin Swagger UI call region
     const ui = SwaggerUIBundle({
@@ -107,8 +98,7 @@
       ],
       plugins: [
         SwaggerUIBundle.plugins.DownloadUrl,
-        pinLoginPlugin,
-        clearValidator
+        pinLoginPlugin
       ],
       layout: 'StandaloneLayout',
       displayOperationId: true,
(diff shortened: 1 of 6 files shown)


oauth2/src/main/resources/swagger/index.html
```diff
@@ -1,33 +1,15 @@
 <!-- Copied from the swagger-ui/index.html static file -->
 <!DOCTYPE html>
-<html lang="en" xmlns:th="http://www.thymeleaf.org">
+<html lang="en">
 <head>
   <meta charset="UTF-8">
   <title>Swagger UI</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" >
+  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" />
+  <link rel="stylesheet" type="text/css" href="index.css" />
```
davidangb, line 8: These changes align our checked-in index.html with the index.html in the 5.17.14 webjar

oauth2/src/main/resources/swagger/index.html
```diff
@@ -1,33 +1,15 @@
...
-    {
-      box-sizing: inherit;
-    }
-
-    body
-    {
-      margin:0;
-      background: #fafafa;
-    }
-
```
davidangb, line 30: These classes are in the webjar's index.css and thus don't need to be repeated here

oauth2/src/main/resources/swagger/index.html
```diff
@@ -103,15 +86,6 @@
...
 
-  // Removes the online validation since some code gen bugs cause us to have some specs that cause the validator to squawk
-  const clearValidator = function(system) {
-    return {
-      components: {
-        onlineValidatorBadge: function() { return null; },
-      }
-    };
-  }
-
```
davidangb, line 114: modern swagger-ui supports a `validatorUrl` config value; I'm using that config value below instead of this custom override

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -134,8 +135,9 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
           }
         }
       } ~
-      (pathPrefixTest("swagger-ui") | pathPrefixTest("oauth2-redirect") | pathSuffixTest("js")
-        | pathSuffixTest("css") | pathPrefixTest("favicon")) {
+      // supporting files for swagger ui
+      (pathPrefixTest("swagger-ui") | pathPrefixTest("oauth2-redirect") | pathPrefixTest("favicon")
+        | pathSuffixTest("index.css") ) {
```
davidangb, line 140: * The usage of `pathSuffixTest` here for index.css was incorrect, and did not properly serve that file. Updated.
* The `pathSuffixTest("js")` was redundant; all the js we need to serve is already covered by `pathPrefixTest("swagger-ui")`

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -134,8 +141,9 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
           }
         }
       } ~
-      (pathPrefixTest("swagger-ui") | pathPrefixTest("oauth2-redirect") | pathSuffixTest("js")
-        | pathSuffixTest("css") | pathPrefixTest("favicon")) {
+      // supporting files for swagger ui
+      (pathPrefixTest("swagger-ui") | pathPrefixTest("oauth2-redirect") | pathPrefixTest("favicon")
+        | pathSuffixTest("index.css")) {
```
davidangb, line 161: * The usage of `pathSuffixTest` here for index.css was incorrect, and did not properly serve that file. Updated.
* The `pathSuffixTest("js")` was redundant; all the js we need to serve is already covered by `pathPrefixTest("swagger-ui")`

oauth2/src/main/resources/swagger/index.html
```diff
@@ -67,16 +49,17 @@
...
 <body>
 <div id="swagger-ui"></div>
-
-<script src="./swagger-ui-bundle.js"> </script>
-<script src="./swagger-ui-standalone-preset.js"> </script>
-<script th:inline="javascript">
+<script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
+<script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
+<!-- START Terra customization -->
+<!-- <script src="./swagger-initializer.js" charset="UTF-8"> </script> -->
```
LizBaldo: Is this something you want to delete or uncomment?

LizBaldo (approved):
I just have one minor comment about a possible cleanup, but otherwise LGTM 👍

oauth2/src/main/resources/swagger/index.html
davidangb: I've added an extra code comment here to explain - does that help?

LizBaldo: Ahh gotcha, yes thank you!

oauth2/src/main/resources/swagger/index.html
```diff
@@ -1,33 +1,15 @@
...
+<html lang="en">
 <head>
   <meta charset="UTF-8">
   <title>Swagger UI</title>
-  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" >
+  <link rel="stylesheet" type="text/css" href="./swagger-ui.css" />
+  <link rel="stylesheet" type="text/css" href="index.css" />
   <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
   <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
+  <!-- START Terra customization -->
```
rtitle, line 11: I guess Terra still has some customizations, so is it not worth doing https://broadworkbench.atlassian.net/browse/TOAZ-205 and reverting to the "stock" `index.html`?

Noting that I do enjoy some of the Terra customizations, like the floating Authorize button when you scroll. (I also haven't tried a modern version of swagger-ui.)

rtitle (approved):
Didn't test it, but this looks good to me

oauth2/src/main/resources/swagger/index.html
davidangb: Some of the customizations are absolutely necessary, like the call to `initOAuth` and the `oauth2RedirectUrl`. So, we can't use the stock index.html as-is; we need to be able to handle customizations one way or another. I poked at a few approaches:

1. Use the stock index.html, but tweak the akka-http routes such that index.css and swagger-initializer.js are served from workbench-libs instead of the webjar. This just trades committing the HTML file for committing a CSS file plus a JS file; not a win.
2. Use the stock index.html, but tweak the `processSwaggerIndex` function so it does a lot more on-the-fly string replacement to inject our customizations. This would work, but it felt harder for an engineer to understand because the source code for index.html is spread across the webjar plus some Scala string-replace code.
3. Keep the current approach of a checked-in copy of index.html, but add explicit comments to show where our customizations start and stop, hoping to make future swagger-ui updates easier. This is (obv) the approach I took in this PR; it felt simplest and most understandable.

rtitle: Ah makes sense - thanks for the explanation

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -111,12 +113,31 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
...
+        .withPath(Uri.Path.Empty)
+        .withQuery(Uri.Query.Empty)
+    ).getOrElse("")
+
+  // default value for the Content-Security-Policy header sent with the swagger-ui index.html file, which includes
+  // the cspOpenIdHost in the connect-src value
+  private val defaultCspHeader =
+    "default-src 'self'; script-src 'self' 'unsafe-inline'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; " +
+      s"connect-src 'self' $cspOpenIdHost; form-action 'none'; " +
+      "frame-ancestors 'none';"
```
davidangb: this adds the appropriate host - https://oauth-proxy.dsp-eng-tools.broadinstitute.org, https://terradevb2c.b2clogin.com, or https://terraprodb2c.b2clogin.com - based on the environment. That way we don't have to list all three hosts in prod.1

question to reviewers: how can we test this will work in prod, without actually releasing it to prod?

LizBaldo: Honestly given that this is for swagger only I would feel okay releasing to prod and revert if needed. I do not think we have a lot of non dev swagger users

LizBaldo (approved):

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
```diff
@@ -111,12 +113,31 @@ class OpenIDConnectAkkaHttpOps(private val config: OpenIDConnectConfiguration) {
       case _                  => Right(tokenUri.query())
     }
 
-  def swaggerRoutes(openApiYamlResource: String): Route = {
+  // Truncate the openId authority endpoint to get a source for use in the CSP header. If this fails, don't
+  // fail the entire page; but beware that swagger-ui login won't work
+  private def cspOpenIdHost: Uri =
```
rtitle: The types seem a little off here - it's a `Try(Uri).getOrElse(String)` (which I think derives to `Serializable`).

Maybe this?
```
  private def cspOpenIdHost: String =
    Try(
      Uri
        .parseAbsolute(config.openIdProvider.authorityEndpoint)
        .withPath(Uri.Path.Empty)
        .withQuery(Uri.Query.Empty)
        .toString()
    ).getOrElse("")
```

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
rtitle: Nice, it seems like this will work. For Leo anyway, I think the authority comes from helmfile here: https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/leonardo/live/prod.yaml#L89

I would probably say testing swagger on BEE and dev is probably sufficient - the prod authority should be wired through correctly already.

rtitle (commented):
Looks good overall, 1 comment about Scala types 🤓

rtitle (changes requested):
Looks good overall, 1 comment about Scala types 🤓

oauth2/src/main/scala/org/broadinstitute/dsde/workbench/oauth2/OpenIDConnectAkkaHttpOps.scala
davidangb: argg, you are totally right, good catch. Fixed in 6e944da1d76acfa88d4ebb49e2fccc06b8a78aba.

rtitle (approved):
:shipit:

rtitle (approved):

PR #1707, opened 2024-10-02
[TOAZ-370] Increase autoLock renew duration.

jsaun:
**PR checklist**

In all cases:


azure/src/main/scala/org/broadinstitute/dsde/workbench/azure/AzureServiceBusReceiverClientWrapperInterp.scala
```diff
@@ -37,6 +38,7 @@ private class AzureServiceBusReceiverClientWrapperInterp(subscriberConfig: Azure
       .receiveMode(ServiceBusReceiveMode.PEEK_LOCK)
       .prefetchCount(subscriberConfig.prefetchCount)
       .maxConcurrentCalls(subscriberConfig.maxConcurrentCalls)
+      .maxAutoLockRenewDuration(Duration.ofMinutes(25))
```
giventocode: does it make sense to add this to `AzureServiceBusSubscriberConfig`, and surface a config setting all the way to Leo?  In my experience, the SB timeout settings are tricky to get right, so it may be worth doing the extra work upfront to gain flexibility in the future. Along these lines, should we also expose the message lock duration?

giventocode (approved):

PR #1715, opened 2024-12-10
[DT-1091]Rollback Addition of SnapshotRequestSubmittedNotification 

rjohanek:
## Context 
The work to add this notification was done in workbench-libs in these two PRs:  [#1701](https://github.com/broadinstitute/workbench-libs/pull/1701), [#1702](https://github.com/broadinstitute/workbench-libs/pull/1702). However, the work was never completed in thurloe. Here is the unmerged thurloe PR: [#310]( https://github.com/broadinstitute/thurloe/pull/310). We paused on this work because if the data explorer is integrated with DUOS this is no longer necessary, so instead of merging this into thurloe, we decided to rollback this change. This mismatch is blocking development in workbench-libs and thurloe.

## Summary of Changes
Rolls back the work done in the two linked workbench-libs PRs. Makes a major version bump because it is a deletion of code that exists in the public API. I decided to make this a minor version bump, because it is a rollback and although it is deletion of code in the public API, since the version is <1.0, the API should not be considered stable at this point. Although, the checklist recommends deprecating code instead of deleting, I believe deleting is necessary in this case, to fully rollback the addition and resolve the incompatibility between thurloe and workbench-libs.

**PR checklist**

In all cases:

### DIFF
--- a/notifications/CHANGELOG.md
+++ b/notifications/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 This file documents changes to the `workbench-notifications` library, including notes on how to upgrade to new versions.
 
+## 0.9
+
+SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-notifications" % "0.8-TRAVIS-REPLACE-ME"
+
+- Removed notification for when a TDR snapshot access request is submitted
+
 ## 0.8
 
 SBT dependency: `"org.broadinstitute.dsde.workbench" %% "workbench-notifications" % "0.8-78b1597"
--- a/notifications/CHANGELOG.md
+++ b/notifications/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 This file documents changes to the `workbench-notifications` library, including notes on how to upgrade to new versions.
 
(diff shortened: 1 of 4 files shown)


notifications/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This file documents changes to the `workbench-notifications` library, including notes on how to upgrade to new versions.
 
+## 1.0
```
fboulnois: Since the version was <1.0, I think a bump to 0.9 *could* be justified. Per [semantic versioning](https://semver.org/#spec-item-4):

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

rjohanek: okay that makes sense to me! I can change it to 0.9

dvoet: I agree, especially since this is rollback

dvoet (approved):

fboulnois (approved):
looks good, thank you 👍

PR #1720, opened 2025-01-27
TSPS-402 update Teaspoons Successful Notification to include userDataTtl string

jsotobroad:
We need an extra param for our successful notification email so that we can inform the users how long they have before their data will be deleted and no longer downloadable.  This is the first step before making the change in thurloe and then updating Teaspoons

Jira ticket - https://broadworkbench.atlassian.net/browse/TSPS-402

**PR checklist**

In all cases:

### DIFF
--- a/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
+++ b/notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
@@ -213,11 +213,12 @@ object Notifications {
                                                timeCompleted: String,
                                                quotaConsumedByJob: String,
                                                quotaRemaining: String,
-                                               userDescription: String
+                                               userDescription: String,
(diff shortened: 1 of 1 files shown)


notifications/src/main/scala/org/broadinstitute/dsde/workbench/model/Notification.scala
```diff
@@ -213,11 +213,12 @@ object Notifications {
                                                timeCompleted: String,
                                                quotaConsumedByJob: String,
                                                quotaRemaining: String,
-                                               userDescription: String
+                                               userDescription: String,
+                                               userDataTtl: String
```
mmorgantaylor: can we call this something more descriptive? i'm not sure if this is supposed to be a duration or a timestamp

mmorgantaylor (approved):

PR #1724, opened 2025-02-10
CORE-278: disable polling for service account key creation

davidangb:
**PR checklist**

In all cases:

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,11 @@ Workbench utility libraries, built for 2.13. You can find the full list of packa
 
 Contains utility functions for talking to Google APIs and DAOs for Google PubSub, Google Directory, Google IAM, and Google BigQuery.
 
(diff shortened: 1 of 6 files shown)


marctalbott (approved):

dvoet (commented):
Why are we keeping the option to poll? I don't think it really enhances the perceived availability of a key.

davidangb:
> Why are we keeping the option to poll? I don't think it really enhances the perceived availability of a key.

@dvoet I originally left this as an option to give us the most flexibility, since we haven't yet fully solved the key-creation issues. But if we want to poll we can (in the short term) revert to the 0.34 library. And, I agree that polling isn't a reliable solution anyway. I've deleted the functionality altogether.

dvoet (approved):

marctalbott (approved):

PR #1726, opened 2025-03-19
[AN-468] Update to kubernetes Java Client to 23.0.0

LizBaldo:
JIRA Ticket: https://broadworkbench.atlassian.net/browse/AN-468

Decided on major breaking change since I had to address breaking changes on the kubernetes java client side of things

**PR checklist**

In all cases:

### DIFF
--- a/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
+++ b/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.workbench.google2
 
 import collection.JavaConverters._
 import io.kubernetes.client.openapi.models.{
+  RbacV1Subject,
   V1Container,
   V1ContainerPort,
   V1Namespace,
@@ -16,8 +17,7 @@ import io.kubernetes.client.openapi.models.{
   V1Service,
   V1ServiceAccount,
   V1ServicePort,
-  V1ServiceSpec,
-  V1Subject
+  V1ServiceSpec
 }
 import org.apache.commons.codec.binary.Base64
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName._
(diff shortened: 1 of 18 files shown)


google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
```diff
@@ -362,10 +362,10 @@ object JavaSerializableInstances {
           .rules(role.rules.map(_.getJavaSerialization).asJava)
     }
 
-  implicit val kubernetesSubjectSerializable: JavaSerializable[KubernetesSubject, V1Subject] =
-    new JavaSerializable[KubernetesSubject, V1Subject] {
-      def getJavaSerialization(subject: KubernetesSubject): V1Subject =
-        new V1Subject()
+  implicit val kubernetesSubjectSerializable: JavaSerializable[KubernetesSubject, RbacV1Subject] =
```
LizBaldo, line 365: Not 100% sure about V1Subject -> RbacV1Subject vs. FlowControlV1Subject to be honest, it is a bit tricky to track down

project/Settings.scala
```diff
@@ -130,13 +130,13 @@ object Settings {
   val google2Settings = commonSettings ++ List(
     name := "workbench-google2",
     libraryDependencies ++= google2Dependencies,
-    version := createVersion("0.37")
+    version := createVersion("0.4")
```
lucymcnatt: I don't think we can use just 0.4 here... there was already a 0.4 version a long time ago, but maybe 0.40 would work?  https://github.com/broadinstitute/workbench-libs/blob/39b4ef35769fcfe81b2a4eddea378ef8ef3d57c7/google2/CHANGELOG.md#01

google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala
lucymcnatt: yeah I couldn't find anything in recent changelogs or a little searching around, odd

project/Settings.scala
LizBaldo: 😭 Very good catch, thank you! I'll update this :)

LizBaldo:
Version should be fixed, and I also added the `TRAVIS_REPLACE_ME` for all of the other modules as well (otherwise they all end up with different versions, which can be annoying for services using the client, like Leo)

salonishah11 (approved):

lucymcnatt (approved):

aednichols (approved):

PR #1749, opened 2026-06-27
ensureGroupAllowsExternalMembers method

davidangb:
Add a new `ensureGroupAllowsExternalMembers()` method on GoogleDirectoryDAO.

This method retrieves a group's current settings, inspects the "allowsExternalMembers" setting, then updates the group to set "allowsExternalMembers = true" if not already set.

Sam could call this method when working with a group to ensure it has the appropriate value, such as just before syncing a group or in a repair loop over all groups.

---

**PR checklist**

In all cases:

### DIFF
--- a/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
+++ b/google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
@@ -27,6 +27,16 @@ trait GoogleDirectoryDAO {
   def isGroupMember(groupEmail: WorkbenchEmail, memberEmail: WorkbenchEmail): Future[Boolean]
   def listGroupMembers(groupEmail: WorkbenchEmail): Future[Option[Seq[String]]]
 
+  def ensureGroupAllowsExternalMembers(groupEmail: WorkbenchEmail): Future[GroupSettings]
+
+  /**
+   * Whether a group's settings permit external (outside the organization) members. Google represents the
+   * allowExternalMembers setting as a nullable string; this normalizes it to a Boolean. A null or otherwise
+   * unparseable value is treated as false.
+   */
+  def allowsExternalMembers(settings: GroupSettings): Boolean =
+    java.lang.Boolean.parseBoolean(settings.getAllowExternalMembers)
+
   def lockedDownGroupSettings =
     new GroupSettings()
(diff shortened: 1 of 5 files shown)


google/src/main/scala/org/broadinstitute/dsde/workbench/google/HttpGoogleDirectoryDAO.scala
```diff
@@ -56,6 +56,36 @@ class HttpGoogleDirectoryDAO(appName: String,
...
+        whenUsageLimited,
+        when404,
+        whenInvalidValueOnBucketCreation,
+        whenNonHttpIOException,
+        when400,
+        whenGroupDoesNotExist
+      ) { () =>
+        executeGoogleRequest(updater)
+      }
+    }
```
davidangb, line 74: I chose to make this new method, which calls `groups().patch()`, instead of reusing the `updateGroupSettings` just above, which calls `groups().update()`. Since we are only asking to change this one flag, using patch felt safer than a full update.

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
```diff
@@ -27,6 +27,16 @@ trait GoogleDirectoryDAO {
   def isGroupMember(groupEmail: WorkbenchEmail, memberEmail: WorkbenchEmail): Future[Boolean]
   def listGroupMembers(groupEmail: WorkbenchEmail): Future[Option[Seq[String]]]
 
+  def ensureGroupAllowsExternalMembers(groupEmail: WorkbenchEmail): Future[GroupSettings]
```
aednichols: "Ensure" is not always my favorite coding verb... without looking at the function implementation, one could think that it e.g. throws an exception when external groups are not enabled.

`enableExternalMembersIfNeeded`?

aednichols (approved):
Thank you for digging into this! One 200% optional semantic gripe 🙂

snf2ye (approved):

marctalbott (approved):

google/src/main/scala/org/broadinstitute/dsde/workbench/google/GoogleDirectoryDAO.scala
davidangb: renamed! Happy to let someone else name things :)

PR #1751, opened 2026-07-14
CTM-569: Exclude k8s client

davidangb:
CTM-569

The following modules no longer include `io.kubernetes:client-java` as dependencies:
* errorReporting
* google
* notifications
* oauth2
* opentelemetry
* serviceTest
* util2

by excluding `client-java`, we also exclude its often-problematic transitive dependencies such as `jackson-databind`, bouncycastle, and `jose4j`.

In some cases, this should make dependency-management and security updates easier in consuming services by eliminating troublesome libraries.

Tested by publishing `-SNAP` versions from my local environment, then including those versions in Orch, and seeing the benefit.

---

**PR checklist**

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -73,6 +73,11 @@ object Dependencies {
   val googleDataproc: ModuleID =    "com.google.cloud" % "google-cloud-dataproc" % "4.27.0"
   val googleContainer: ModuleID = "com.google.cloud" % "google-cloud-container" % "2.31.0"
   val kubernetesClient: ModuleID = "io.kubernetes" % "client-java" % "23.0.0"
+  // util2 only needs io.kubernetes.client.openapi.ApiException (used in withLogging). That class lives
+  // in the lighter client-java-api sub-artifact and is self-contained (references only JDK types), so we
+  // drop all of client-java-api's transitive deps (okhttp, gson, jackson, swagger, ...) — none are needed.
+  val kubernetesClientApi: ModuleID =
+    "io.kubernetes" % "client-java-api" % "23.0.0" excludeAll ExclusionRule(organization = "*", name = "*")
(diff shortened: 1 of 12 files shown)


project/Dependencies.scala
```diff
@@ -259,9 +259,8 @@ object Dependencies {
     circeCore,
     circeParser,
     circeGeneric,
-    catsMtl,
-    kubernetesClient
-  ) ++ bouncyCastleOverrides
+    catsMtl
+  )
```
davidangb, line 263: since util2 no longer includes bouncyCastle, there is no need for its overrides

util2/src/main/scala/org/broadinstitute/dsde/workbench/util2/package.scala
```diff
@@ -52,6 +52,32 @@ package object util2 {
...
+  // The response body logged on failure: for a Kubernetes ApiException that is getResponseBody (read
+  // reflectively), otherwise the exception message. Falls back to getMessage if the reflective call fails.
+  private def failureResponse(t: Throwable): Option[String] =
+    if (isKubernetesApiException(t))
+      scala.util
+        .Try(Some(t.getClass.getMethod("getResponseBody").invoke(t).asInstanceOf[String]))
+        .getOrElse(Some(t.getMessage))
+    else
+      Some(t.getMessage)
+
```
davidangb, line 80: this jumps through some hoops to handle k8s exceptions like it did before this PR, but those hoops feel worth it to eliminate a big dependency.

aednichols (approved):
Neat! This is a very thoughtful and forward-thinking enhancement.

I was _almost_ able to exclude `io.kubernetes` from Rawls in https://github.com/broadinstitute/rawls/pull/3567, but it crashed at runtime looking for that one exception class. [CI log link.](https://github.com/broadinstitute/rawls/actions/runs/29284612512/job/86934009391)

jgainerdewar (approved):

